PR #4, opened 2023-07-02
chore: Fully adopt the "main realm/context" terminology

andreubotella:
It used to be that the realm/context created with the isolate was sometimes called "global context" and sometimes "main context/realm". This goes back to the days before `JsRealm`, when the `JsRuntime::global_context` method returned a `v8::Global` for the only `v8::Context` that existed in the isolate. Nowadays that is no longer the case, and we should call it the main realm.

That change by itself would also rename some properties/variables from `is_global` to `is_main`, which is also used elsewhere to refer to the top-level `JsRuntime`/isolate in a process. Therefore, the names are changed to `is_main_runtime` and `is_main_realm` to disambiguate.


andreubotella:
Since most of this PR's diff is taken up by updating uses of changed APIs, I thought it would help review if I listed all of the actual changes:

`deno_core` API breaking changes:

- `JsRuntime::global_context` is renamed to `main_context`.
- `JsRuntime::global_realm` is renamed to `main_realm`.

Private API changes:

- `JsRuntime::is_main` is renamed to `is_main_runtime`.
- `JsRuntimeState::global_realm` is renamed to `main_realm`.
- `JsRealmInner::is_global` is renamed to `is_main_realm`.

Variable renames:

- The third argument of `JsRuntimeInspector::new`, which is currently `is_main`, is renamed to `is_main_runtime`. (Also a public documentation change.)
- The `global_context` and `global_realm` variables in `JsRuntime::new_inner` are renamed to `main_context` and `main_realm`, respectively.
- The `global_realm` variables in `JsRuntime::dynamic_import_module_evaluate` and `JsRuntime::mod_evalute` are renamed to `main_realm`.
- The `global_realm` variables in the `test_set_promise_reject_callback_realms` and `js_realm_async_ops` tests (both in `core/runtime/tests/jsrealm.rs`) are renamed to `main_realm`.

Other changes:

- The main realm is shown in the inspector as `main realm`, rather than `global context`. This change is user-facing, but probably not breaking or too important.
- The documentation of `JsRuntime::execute_script` and `JsRuntime::execute_script_static` is updated to replace one reference to "global context" with "main realm", in a context where the difference between context and realm doesn't matter.
- The string used in the `test_set_promise_reject_callback_realms` test to test the main realm is changed from `"global_realm"` to `"main_realm"`.

mmastrac (approved):

PR #7, opened 2023-07-02
fix: ignore already evaluated extension modules

nayeemrmn:
Pulled from https://github.com/denoland/deno/pull/19519. This is pretty much all it takes to replace `init_ops()` calls with `init_ops_and_esm()` without error, so snapshotting embedders can just use the latter whether they are creating a snapshot or consuming it.

(Using that _might_ be a footgun without the rest of the changes from that PR because it leaves the JS source paths in the extensions list and if any were mistakenly omitted from the snapshot, the program will attempt to load them at runtime in production. That's what I attempted to address with the `exclude_js_sources` feature. But it was too complex and introduced differences in the test build. Maybe we can find a different safety for that which doesn't rely on conditional compilation.)

### DIFF
--- a/core/runtime/jsruntime.rs
+++ b/core/runtime/jsruntime.rs
@@ -901,6 +901,16 @@ impl JsRuntime {
               panic!("{} not present in the module map", specifier)
             })
         };
+        {
+          let module_map_rc = self.module_map.clone();
+          let module_map = module_map_rc.borrow();
+          let handle = module_map.handles.get(mod_id).unwrap().clone();
+          let mut scope = realm.handle_scope(self.v8_isolate());
+          let handle = v8::Local::new(&mut scope, handle);
+          if handle.get_status() == v8::ModuleStatus::Evaluated {
+            continue;
+          }
+        }
         let receiver = self.mod_evaluate(mod_id);
         self.run_event_loop(false).await?;
         receiver
--- a/core/runtime/tests.rs
+++ b/core/runtime/tests.rs
@@ -2394,3 +2394,25 @@ fn ops_in_js_have_proper_names() {
   "#;
   runtime.execute_script_static("test", src).unwrap();
 }
+
+#[test]
+fn duplicate_extension_esm_specifier() {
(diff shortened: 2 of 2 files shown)


core/runtime/jsruntime.rs
```diff
@@ -901,6 +901,16 @@ impl JsRuntime {
               panic!("{} not present in the module map", specifier)
             })
         };
+        {
+          let module_map_rc = self.module_map.clone();
+          let module_map = module_map_rc.borrow();
+          let handle = module_map.handles.get(mod_id).unwrap().clone();
+          let mut scope = realm.handle_scope(self.v8_isolate());
+          let handle = v8::Local::new(&mut scope, handle);
+          if handle.get_status() == v8::ModuleStatus::Evaluated {
```
dsherret, line 910: Maybe I'm misunderstanding, but what if the new module is slightly different than the one found in the snapshot? Shouldn't we panic instead since this isn't supported? https://github.com/denoland/deno_core/pull/39

nayeemrmn: Maybe we could add a debug assert to check that the content is the same. Then again it's not likely to happen with the `extension!()` declarations. And IMO it's conceptually similar to encountering an import of an already-evaluated module with the loader in a new state with new content for that module; it would be standard to ignore it in that case.

If things surrounding module evaluation were more idempotent like with V8 / ESM semantics, I've seen that it would simplify and limit the need for conditional compilation in certain uses.

dsherret: Someone could be loading stuff dynamically though and not realize this, then be super confused when their updated code wasn't updated because a past snapshot had the old out of date code. I think we should hard fail if someone provides the same module twice and it's more efficient to enforce that.

nayeemrmn:
I realised it's not possible to share code between a crate's `build.rs` and its runtime code either way, so unifying `init_ops()` and `init_ops_and_esm()` doesn't create much deduping potential after all. Closing.

PR #9, opened 2023-07-03
Support for global template and global object

bartlomieju:


mmastrac (approved):
LGTM but I think we'll want to switch from `Box<Fn<...>>` to `fn(...)` to avoid a box at startup.

bartlomieju:
Let's improve boxing situation after we get everything working in `deno` itself.

PR #12, opened 2023-07-03
feat(core): Implement v8::XXX args and return

mmastrac:
`op2` support for `v8::Value` and related types. Functions may take arguments of the form `&v8::String` (or `&mut v8::String`), `v8::Local<v8::String>`, or an `Option<...>` of any of these. Under the hood, the op system will check the type of the argument and automatically throw a type error if it does not match.

Implements:

 - [x] v8::Local<v8::XXX> for arguments (including automatic casting so `v8::String` will work)
 - [x] v8::Local<v8::XXX> for return types
 - [x] Option<v8::Local<...>> for arguments and returns
 - [x] &v8::XXX and Option<&v8:XXX> for arguments (ie: pass a `&mut v8::String` without a handle)
 - [x] HandleScope

Additional changes:

 - [x] Fixes clippy attributes on ops
 - [x] Refactoring of signature parser to be slightly less tied to our final arg/return types
 - [x] Ensures that lifetimes are properly added to `call`

### DIFF
--- a/core/runtime/ops.rs
+++ b/core/runtime/ops.rs
@@ -332,6 +332,7 @@ mod tests {
       op_test_v8_option_string,
       op_test_v8_type_return,
       op_test_v8_type_return_option,
+      op_test_v8_type_handle_scope,
     ]
   );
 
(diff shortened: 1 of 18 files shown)


core/runtime/ops.rs
```diff
@@ -631,4 +637,100 @@ mod tests {
   // We don't actually test this one -- we just want it to compile
   #[op2(core, fast)]
   pub fn op_test_generics<T: Clone>() {}
+
+  /// Tests v8 types without a handle scope
+  #[allow(clippy::needless_lifetimes)]
```
bartlomieju, line 642: Can this be handled by the op itself so we don't need to spam these ignores everywhere in our code?

mmastrac: This is only used here in the tests because the lifetime is useless and I wanted some tests that included it.

bartlomieju (approved):
LGTM

PR #13, opened 2023-07-05
fix: Re-export RuntimeSnapshotOptions

edrevo:
It seems like JsRuntimeForSnapshot cannot be created from outside of deno-core because RuntimeSnapshotOptions isn't publicly exported.

This is a port of https://github.com/denoland/deno/pull/19661


bartlomieju (approved):
Thanks!

PR #14, opened 2023-07-05
fix(modules): no duplicate loads

bartlomieju:
This is an edge case that @satyarohith discovered when trying to optimize
loading of source code. We have some situations with redirects that a load
is scheduled `load A`, a module gets loaded in `load B` that happens before
`load A` and the source code is consumed at this point. Then `load A` happens
even though the module has already been loaded.

### DIFF
--- a/core/modules/mod.rs
+++ b/core/modules/mod.rs
@@ -523,10 +523,18 @@ impl RecursiveModuleLoad {
             let request = module_request.clone();
             let specifier =
               ModuleSpecifier::parse(&module_request.specifier).unwrap();
+            let asserted_module_type = module_request.asserted_module_type.clone();
             let referrer = referrer.clone();
             let loader = self.loader.clone();
             let is_dynamic_import = self.is_dynamic_import();
+            let module_map_rc = self.module_map_rc.clone();
             let fut = async move {
+              if module_map_rc
+                .borrow()
+                .get_id(&specifier, asserted_module_type).is_some() {
+                  return Ok(());
+                };
+
               let load_result = loader
                 .load(&specifier, Some(&referrer), is_dynamic_import)
                 .await;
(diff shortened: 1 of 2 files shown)


core/modules/mod.rs
```diff
@@ -523,10 +523,18 @@ impl RecursiveModuleLoad {
...
             let referrer = referrer.clone();
             let loader = self.loader.clone();
             let is_dynamic_import = self.is_dynamic_import();
+            let module_map_rc = self.module_map_rc.clone();
             let fut = async move {
+              if module_map_rc
+                .borrow()
+                .get_id(&specifier, asserted_module_type).is_some() {
+                  return Ok(());
+                };
```
bartlomieju, line 536: I think this is what the fix should look like, but the return type is broken here

bartlomieju: @mmastrac I think in this case we should return `Ok(None)` and change how `RecursiveModuleLoad` to ignore such return values - ie. The load can be scheduled, but it can decide to not return anything if the file has been already loaded. Does that make sense to you?

core/modules/tests.rs
```diff
@@ -1250,3 +1250,109 @@ fn import_meta_snapshot() {
...
+  let mut runtime = JsRuntime::new(RuntimeOptions {
+    module_loader: Some(loader),
+    ..Default::default()
+  });
+
+  let spec = resolve_url("file:///main.js").unwrap();
+  let a_id = runtime.load_main_module(&spec, None).await.unwrap();
+
+  runtime.run_event_loop(false).await.unwrap();
+  runtime.mod_evaluate(a_id).await.unwrap().unwrap();
```
nayeemrmn, line 1357: This hangs,
```suggestion
  #[allow(clippy::let_underscore_future)]
  let _ = runtime.mod_evaluate(a_id);
  runtime.run_event_loop(false).await.unwrap();
```

bartlomieju:
Closing in favor of #56.

PR #18, opened 2023-07-07
refactor: migrate some ops_builtin_v8 to op2 macro

bartlomieju:
These are all the ops I managed to migrate to `op2` macro.

### DIFF
--- a/core/ops_builtin.rs
+++ b/core/ops_builtin.rs
@@ -356,12 +356,12 @@ fn op_is_proxy(value: serde_v8::Value) -> bool {
   value.v8_value.is_proxy()
 }
 
-#[op(v8)]
-fn op_str_byte_length(
+#[op2(core)]
+pub fn op_str_byte_length(
   scope: &mut v8::HandleScope,
-  value: serde_v8::Value,
+  value: v8::Local<v8::Value>,
 ) -> u32 {
-  if let Ok(string) = v8::Local::<v8::String>::try_from(value.v8_value) {
+  if let Ok(string) = v8::Local::<v8::String>::try_from(value) {
     string.utf8_length(scope) as u32
   } else {
     0
--- a/core/ops_builtin.rs
+++ b/core/ops_builtin.rs
@@ -346,9 +346,10 @@ async fn op_shutdown(
   resource.shutdown().await
 }
 
-#[op]
-fn op_format_file_name(file_name: String) -> String {
-  format_file_name(&file_name)
+#[op2(core)]
+#[string]
+fn op_format_file_name(#[string] file_name: &str) -> String {
+  format_file_name(file_name)
 }
 
 #[op(fast)]
@@ -357,7 +358,7 @@ fn op_is_proxy(value: serde_v8::Value) -> bool {
 }
 
 #[op2(core)]
-pub fn op_str_byte_length(
+fn op_str_byte_length(
   scope: &mut v8::HandleScope,
   value: v8::Local<v8::Value>,
 ) -> u32 {
--- a/core/ops_builtin_v8.rs
+++ b/core/ops_builtin_v8.rs
@@ -44,19 +44,18 @@ pub fn op_unref_op(scope: &mut v8::HandleScope, promise_id: i32) {
   context_state.borrow_mut().unrefed_ops.insert(promise_id);
 }
 
-#[op(v8)]
-fn op_set_promise_reject_callback<'a>(
+#[op2(core)]
+pub fn op_set_promise_reject_callback<'a>(
   scope: &mut v8::HandleScope<'a>,
-  cb: serde_v8::Value,
-) -> Result<Option<serde_v8::Value<'a>>, Error> {
-  let cb = to_v8_fn(scope, cb)?;
+  cb: v8::Local<'a, v8::Function>,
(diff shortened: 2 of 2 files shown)


bartlomieju:
Currently biggest blocker to migrate more ops appears to be ability to return `bool` and `Result<>` from `#[op2(core)]`

core/ops_builtin.rs
```diff
@@ -346,22 +346,23 @@ async fn op_shutdown(
...
 fn op_is_proxy(value: serde_v8::Value) -> bool {
   value.v8_value.is_proxy()
 }
 
-#[op(v8)]
+#[op2(core)]
 fn op_str_byte_length(
   scope: &mut v8::HandleScope,
-  value: serde_v8::Value,
+  value: v8::Local<v8::Value>,
```
mmastrac, line 363: I think this can be `&v8::Value`, but I'm wondering if we should just allow this function to throw type errors and make it `fast`. Does it need to handle invalid v8 values?

core/ops_builtin_v8.rs
```diff
@@ -861,53 +854,49 @@ fn op_apply_source_map(
...
 
-#[op(v8)]
-fn op_store_pending_promise_rejection<'a>(
-  scope: &mut v8::HandleScope<'a>,
-  promise: serde_v8::Value<'a>,
-  reason: serde_v8::Value<'a>,
+#[op2(core)]
+pub fn op_store_pending_promise_rejection(
+  scope: &mut v8::HandleScope,
+  promise: v8::Local<v8::Promise>,
```
mmastrac, line 880: I added a place where we can make conversion to Globals automatic but didn't implement it -- I can circle back and implement this to reduce some more boilerplate here.

core/ops_builtin_v8.rs
```diff
@@ -144,23 +138,22 @@ fn op_eval_context<'a>(
...
+  scope.enqueue_microtask(cb);
 }
 
-#[op(v8)]
-fn op_create_host_object<'a>(
+#[op2(core)]
+pub fn op_create_host_object<'a>(
   scope: &mut v8::HandleScope<'a>,
-) -> serde_v8::Value<'a> {
+) -> v8::Local<'a, v8::Value> {
```
mmastrac, line 152: I think this could return `v8::Local<v8::Object>`

core/ops_builtin_v8.rs
```diff
@@ -31,43 +32,36 @@ fn to_v8_fn(
...
-#[op(v8)]
-fn op_set_promise_reject_callback<'a>(
+#[op2(core)]
+pub fn op_set_promise_reject_callback<'a>(
   scope: &mut v8::HandleScope<'a>,
-  cb: serde_v8::Value,
-) -> Result<Option<serde_v8::Value<'a>>, Error> {
-  let cb = to_v8_fn(scope, cb)?;
+  cb: v8::Local<'a, v8::Function>,
+) -> Option<v8::Local<'a, v8::Value>> {
```
mmastrac, line 51: Return `v8::Local<v8::Function>`?

core/ops_builtin.rs
mmastrac: Thinking about it more, it's probably better to keep it slow so we can use utf8_length, but I think `&v8::String` might be better.

bartlomieju: I expect it will break some API if we change it to `v8::String` (which I assume will make it throw `TypeError` if the passed object is not a string instead of returning `0`).

core/ops_builtin_v8.rs
bartlomieju: I'm not sure I'm a fan of this solution - creating `Global`s is expensive, I think it shouldn't be hidden away.

bartlomieju:
> Currently biggest blocker to migrate more ops appears to be ability to return `bool` and `Result<>` from `#[op2(core)]`

It appears the blocker is now gone, I will migrate more ops before giving it for a review.

core/ops_builtin.rs
mmastrac: I couldn't see any callers at first glance, but we could revisit this one later on to confirm.

core/ops_builtin_v8.rs
mmastrac: I think if that's the case, we should add a `#[global]` which draws the eye to the expense. I don't mind adding slightly more expensive boilerplate if we make it searchable and visible.

core/ops_builtin.rs
bartlomieju: I believe it's used in `ext/node` (notice that it's exposed as `core.byteLength()`)

core/ops_builtin_v8.rs
bartlomieju: +1 for attribute

bartlomieju:
@mmastrac it appears that there's still no way to have an op accept a `&mut HandleScope` and return `serde` structs. Is that by design? (I get borrow error)

mmastrac:
> @mmastrac it appears that there's still no way to have an op accept a `&mut HandleScope` and return `serde` structs or `Result<>`. Is that by design? (I get borrow error)

Sounds like a bug. I will try to create a repro and fix it

mmastrac (approved):
LGTM! Great work

PR #21, opened 2023-07-08
fix(ops): defer mut ref for ops2 scope

mmastrac:
Allows returning from functions that take a `HandleScope`


bartlomieju (approved):
LGTM! Thanks for a quick fix

PR #25, opened 2023-07-09
feat(ops): Make v8::* ops fast-capable

mmastrac:
Ops that are `v8::*`-style are fast-capable, so let's implement this. We start sharing a bit of code between fast and slow ops to support this.

### DIFF
--- a/core/runtime/ops.rs
+++ b/core/runtime/ops.rs
@@ -301,6 +301,34 @@ pub fn to_str<'a, const N: usize>(
   string.to_rust_cow_lossy(scope, buffer)
 }
 
+/// Converts from a raw [`v8::Value`] to the expected V8 data type.
+#[inline(always)]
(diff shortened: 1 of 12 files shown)


core/runtime/ops.rs
```diff
@@ -698,7 +726,7 @@ mod tests {
 
   /// Tests v8 types without a handle scope
   #[allow(clippy::needless_lifetimes)]
-  #[op2(core)]
+  #[op2(core, fast)]
```
bartlomieju, line 729: Did compiler now force you to mark it as fast :D ?

bartlomieju (approved):
LGTM!

core/runtime/ops.rs
mmastrac: It did : 😂

PR #28, opened 2023-07-10
feat(ops): Implement #[opstate] for op2

mmastrac:


core/runtime/ops.rs
```diff
@@ -872,4 +881,35 @@ mod tests {
...
+  pub fn op_state_mut(state: &mut OpState, value: u32) {
+    *state.borrow_mut() = value;
+  }
+
+  #[tokio::test]
+  pub async fn test_op_state() -> Result<(), Box<dyn std::error::Error>> {
+    run_test2(
+      10000,
+      "op_state_rc",
+      "if (__index__ == 0) { op_state_rc(__index__) } else { assert(op_state_rc(__index__) == __index__ - 1) }",
```
bartlomieju, line 909: Where is that `__index__` coming from?

core/runtime/ops.rs
```diff
@@ -872,4 +881,35 @@ mod tests {
     )?;
     Ok(())
   }
+
+  #[op2(core, fast)]
+  pub fn op_state_rc(state: Rc<RefCell<OpState>>, value: u32) -> u32 {
```
bartlomieju, line 886: Can we prevent this somehow? `Rc<RefCell<OpState>>` shouldn't be used in a sync op (`&mut OpState` should be used instead`)

ops/op2/test_cases/sync/v8_ref_option.out
```diff
@@ -15,7 +15,7 @@ impl deno_core::_ops::Op for op_v8_lifetime {
             use deno_core::v8::fast_api::Type;
             use deno_core::v8::fast_api::CType;
             deno_core::v8::fast_api::FastFunction::new(
-                &[Type::V8Value, Type::V8Value, Type::V8Value],
```
mmastrac, line 18: This was a bug..  added an assert to catch it.

core/runtime/ops.rs
mmastrac: Definitely -- in this case I wanted to make sure it was working. I think it may be best to add a post-processing step when I add async support that denies this and clean it up, and then migrate this test to async ops. I can make sure that I add a TODO in the roadmap to clean this up later.

core/runtime/ops.rs
mmastrac: This is provided by the `run_test2` method.

core/runtime/ops.rs
bartlomieju: Sounds good to me

bartlomieju (approved):
LGTM

PR #29, opened 2023-07-10
feat(ops): Implement #[state] for automatic OpState extraction

mmastrac:
This provides a useful shortcut for `OpState`-based parameters using a `#[state]` annotation. Parameters annotation w/`#[state]` can be either `&Something`, `&mut Something`, `Option<&Something>`, or `Option<&mut Something>`.

Note that due to the Rust aliasing rules, only one `&mut Something` is allowed. This might be fixable in the future if we modify how the internal state system works. In many cases, a single mutable borrow should be sufficient.


core/runtime/ops.rs
```diff
@@ -901,15 +903,42 @@ mod tests {
...
+    let old_value = *value;
+    *value = new_value;
+    old_value
+  }
+
+  #[op2(core, fast)]
+  pub fn op_state_multi_attr(
+    #[state] value32: &u32,
+    #[state] value16: &u16,
+    #[state] value8: Option<&u8>,
```
bartlomieju, line 917: This feels very magical... I'm not sure how I feel about it. Is there a way to drop this borrow somehow?

mmastrac: OpState is borrowed during the entire time we're in these functions (regardless of whether we're doing this with the helpers or directly against OpState) -- the `borrow` on `GothamState` is a fake borrow that is more like a "deref the item with this key".

Technically you can control the lifetime of the borrow using Rc<RefCell<OpState>> but usually we just assume that OpState is locked during the entire run of a sync function.

mmastrac: This just desugars to:

```
let a: &A = opstate.borrow().borrow::<A>()
let b: &B = opstate.borrow().borrow::<B>()
```

Where it can be extremely useful is if we tweak the state to allow for multiple mutable borrows that might not be as ergonomic as this, or if we decide to change the state interface.

PR #30, opened 2023-07-10
refactor: migrate even more ops to op2 macro

bartlomieju:


bartlomieju:
@mmastrac what do you suggest we do here?

Screenshot 2023-07-10 at 21 49 34

mmastrac (approved):
LGTM

PR #31, opened 2023-07-10
fix(serde_v8): Eliminate illegal & -> &mut transmute

mmastrac:
This code was not and will not be 100% absolutely safe, but Rust nightly will warn about these illegal transmutations and this code is _somewhat_ more correct.


mmastrac:
Merging this into my buffer changes

PR #33, opened 2023-07-11
feat(ops): Basic buffer support for op2 

mmastrac:
Adds:

 - `#[buffer] &[u8]`
 - `#[buffer] &mut [u8]`

TODO:

 - `#[buffer] *const u8` (all primitive types) -- not recommended but used in FFI
 - `#[buffer] *mut u8` (all primitive types)
 - `#[buffer] bytes::Bytes`
 - `#[buffer] JSBuffer`
 - `#[buffer] V8Slice`
 - `#[buffer] Vec<..>`
 - `#[buffer] Box<[..]>`
 
TODO: Alternative buffer modes:

 - `#[buffer(shared)]` (the default)
 - `#[buffer(copy)]` 
 - `#[buffer(detach)]`

### DIFF
--- a/core/runtime/ops.rs
+++ b/core/runtime/ops.rs
@@ -348,22 +348,22 @@ pub fn serde_v8_to_rust<'a, T: Deserialize<'a>>(
 pub fn to_nonresizable_v8_slice(
   scope: &mut v8::HandleScope,
   input: v8::Local<v8::Value>,
-) -> Result<serde_v8::V8Slice, ()> {
+) -> Result<serde_v8::V8Slice, &'static str> {
   let (buf, offset, length) =
     if let Ok(buf) = v8::Local::<v8::ArrayBufferView>::try_from(input) {
       let Some(buffer) = buf.buffer(scope) else {
-      return Err(());
-    };
+        return Err("buffer missing");
+      };
       (buffer, buf.byte_offset(), buf.byte_length())
     } else if let Ok(buf) = v8::Local::<v8::ArrayBuffer>::try_from(input) {
       (buf, 0, buf.byte_length())
     } else {
-      return Err(());
+      return Err("expected ArrayBuffer or ArrayBufferView");
     };
 
   let store = buf.get_backing_store();
   if store.is_resizable_by_user_javascript() {
-    return Err(());
+    return Err("expected non-resizable buffer");
   }
   let slice =
     unsafe { serde_v8::V8Slice::from_parts(store, offset..(offset + length)) };
--- a/core/runtime/ops.rs
+++ b/core/runtime/ops.rs
@@ -343,6 +343,29 @@ pub fn serde_v8_to_rust<'a, T: Deserialize<'a>>(
   from_v8(scope, input)
 }
 
+/// Retrieve a [`serde_v8::V8Slice`] from a value.
+pub fn to_v8_slice(
+  scope: &mut v8::HandleScope,
+  input: v8::Local<v8::Value>,
+) -> Result<serde_v8::V8Slice, ()> {
+  let (buf, offset, length) =
+    if let Ok(buf) = v8::Local::<v8::ArrayBufferView>::try_from(input) {
+      let Some(buffer) = buf.buffer(scope) else {
+      return Err(());
+    };
+      (buffer, buf.byte_offset(), buf.byte_length())
+    } else if let Ok(buf) = v8::Local::<v8::ArrayBuffer>::try_from(input) {
+      (buf, 0, buf.byte_length())
+    } else {
+      return Err(());
+    };
+
+  let store = buf.get_backing_store();
+  let slice =
+    unsafe { serde_v8::V8Slice::from_parts(store, offset..(length - offset)) };
+  Ok(slice)
+}
+
 #[cfg(test)]
 mod tests {
   use crate::error::generic_error;
@@ -394,6 +417,7 @@ mod tests {
       op_state_mut,
       op_state_mut_attr,
       op_state_multi_attr,
+      op_buffers,
     ],
     state = |state| {
       state.put(1234u32);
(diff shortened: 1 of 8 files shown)


ops/op2/signature.rs
```diff
@@ -249,16 +308,33 @@ pub struct ParsedSignature {
   pub generic_bounds: BTreeMap<String, String>,
 }
 
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+enum BufferMode {
+  /// Shared buffers that may possibly change on the JavaScript side upon re-entry into
+  /// V8. Rust code should not treat these as traditional buffers.
+  Shared,
```
aapoalas, line 315: This is a bit confusing since SharedArrayBuffer is a thing but this doesn't actually refer to that.

ops/op2/signature.rs
```diff
@@ -488,10 +577,12 @@ fn parse_generics(
 }
 
 fn parse_attributes(attributes: &[Attribute]) -> Result<Attributes, ArgError> {
-  let attrs = attributes
-    .iter()
-    .filter_map(parse_attribute)
-    .collect::<Vec<_>>();
+  let mut attrs = vec![];
```
aapoalas, line 580: **suggestion:** Maybe use `Vec::with_capacity(attrs.len())`?

ops/op2/signature.rs
mmastrac: Correct -- I don't have a better name, bikeshedding appreciated!

aapoalas: Hmm... Unsafe / Unsound would be technically correct, and would also serve as a reminder that this isn't really the best place to be at and that we should try to avoid adding any more of this sort of usage.

Although: The "one mutable XOR any number of immutable" borrow rule is being respected in synchronous ops, right? Since for the duration of the call JS is blocked we're guaranteed to have sole ownership of the buffer. ... Well, except if our call callbacks into JS and the callback mutates the buffer...

aapoalas:
**thought:** It's possible for V8 ArrayBuffers to be created from external pointers that do not follow the normal alignment rule of V8 ABs (code says always aligned to 4 bytes, experience seems to suggest this is 8 bytes for 64 bit environments). In these cases the TypedArrays will then happily allow for themselves to be created at odd offsets. This needs to be remembered when eg. a `Float64Array` is converted into a Rust slice.

ops/op2/signature.rs
mmastrac: It's the wild west setting -- there's no guarantees at all!

mmastrac: I changed this to `unsafe` for now -- we can revisit

ops/op2/signature.rs
mmastrac: Good point. As this code is leaning more towards readability than performance, I'm going to punt on this for now -- if we find that op parsing is slow overall, it may require a bit of perf work over the whole codebase.

core/runtime/ops.rs
```diff
@@ -343,6 +343,33 @@ pub fn serde_v8_to_rust<'a, T: Deserialize<'a>>(
   from_v8(scope, input)
 }
 
+/// Retrieve a [`serde_v8::V8Slice`] from a value.
+#[allow(clippy::result_unit_err)]
+pub fn to_nonresizable_v8_slice(
+  scope: &mut v8::HandleScope,
+  input: v8::Local<v8::Value>,
+) -> Result<serde_v8::V8Slice, ()> {
```
bartlomieju, line 351: Is there a specific reason why this function returns a unit value for the error?

core/runtime/ops.rs
```diff
@@ -343,6 +343,33 @@ pub fn serde_v8_to_rust<'a, T: Deserialize<'a>>(
...
+  let (buf, offset, length) =
+    if let Ok(buf) = v8::Local::<v8::ArrayBufferView>::try_from(input) {
+      let Some(buffer) = buf.buffer(scope) else {
+      return Err(());
+    };
```
bartlomieju, line 356: This formatting seems funky

ops/op2/dispatch_slow.rs
```diff
@@ -360,6 +365,47 @@ pub fn from_arg(
   Ok(res)
 }
 
+pub fn from_arg_buffer(
+  generator_state: &mut GeneratorState,
+  arg_ident: &Ident,
+  buffer: &Buffer,
+) -> Result<TokenStream, V8MappingError> {
+  let throw_exception =
+    throw_type_error(generator_state, "expected buffer".to_string())?;
```
bartlomieju, line 374: Isn't this eagerly evaluating the type error?

ops/op2/signature.rs
```diff
@@ -337,6 +444,9 @@ fn stringify_token(tokens: impl ToTokens) -> String {
     .map(|s| s.to_string())
     .collect::<Vec<_>>()
     .join("")
+    // Ick.
+    // TODO(mmastrac): Should we pretty-format this instead?
+    .replace(" , ", ", ")
```
bartlomieju, line 449: Probably not worth it - every time I got a compile error the spaces around didn't bother me - it was readable enough

ops/op2/dispatch_slow.rs
mmastrac: Nope -- this is at generation time. It eagerly evaluates the string, but I'm not worrying too much about allocations here (a few hundred ops should process in milliseconds even if we're lazy about it).

core/runtime/ops.rs
mmastrac: Huh.. looks like the rustfmt bug with else statements. Will fix.

bartlomieju (approved):
LGTM

core/runtime/ops.rs
mmastrac: This code is a bit awkward right now -- the error is specified in the generated code and this just returns a success or failure. It's somewhat complicated to deal with errors in the codegen parts and it matches the v8 casts above.

I would eventually like to have these return &' static str that becomes the actual JS error.

I'll see if there's something I can do here right now to switch away from the unit errors, but I may need to do a pass on all the op args to ensure that errors work as we expect.

mmastrac: Fixed in the latest -- we now use a static string here. I'll add a TODO on the roadmap to circle back on errors, but it should be good enough for now I think.

PR #34, opened 2023-07-12
feat(ops): Support #[buffer(copy)], add self-documenting test

mmastrac:
Adds slow and fast call support for:

 - `#[buffer(copy)] Vec<u8>`
 - `#[buffer(copy)] Box<[u8]>`
 - `#[buffer(copy)] bytes::Bytes`

We also start generating and testing a README with various parameter types.


core/runtime/ops.rs
```diff
@@ -1091,4 +1066,38 @@ mod tests {
...
+  ) {
+    assert_eq!(input1[0], input2[0]);
+    assert_eq!(input2[0], input3[0]);
+    input1[0] = 0xff;
+    assert_ne!(input1[0], input2[0]);
+    assert_eq!(input2[0], input3[0]);
+    input2[0] = 0xff;
+    assert_eq!(input1[0], input2[0]);
+    assert_ne!(input2[0], input3[0]);
+  }
```
bartlomieju, line 1087: Really thorough tests!

ops/op2/README.md
```diff
@@ -0,0 +1,564 @@
...
+```
+
+</td><td>
+✅
+</td><td>
+
+</td><td>
+Extracts an object from `OpState`.
+</td></tr>
+</table>
```
bartlomieju, line 564: This is just fantastic

serde_v8/magic/v8slice.rs
```diff
@@ -83,6 +83,16 @@ impl V8Slice {
...
+
+  /// Create a [`Vec<u8>`] copy of this slice data.
+  pub fn to_vec(&self) -> Vec<u8> {
+    self.as_slice().to_vec()
+  }
+
+  /// Create a [`Box<[u8]>`] copy of this slice data.
+  pub fn to_boxed_slice(&self) -> Box<[u8]> {
+    self.to_vec().into_boxed_slice()
+  }
```
bartlomieju, line 95: Are these only temporary?

bartlomieju (approved):
LGTM, great progress 💪

serde_v8/magic/v8slice.rs
mmastrac: Yeah, I think these will go away once we have a better understanding of what async buffers look like and potentially can create a new type. For now, these match the methods available on slices and Vec (both make copies) which should avoid confusion.

PR #35, opened 2023-07-12
feat(op2): Remove support for &mut v8, add support for f32/f64

mmastrac:
Cleanup PR to ensure that already-implemented ops are working as expected:

 - `&mut v8:...` is not a valid type at this time
 - `f32` and `f64` were missing support
 - Add extensive tests for transmutation
 - Fix doc bugs


bartlomieju (approved):
LGTM

PR #36, opened 2023-07-12
feat: `CreateSnapshotOptions` - `with_runtime_cb`

dsherret:
I'm not sure this is the best solution.

In dprint-plugin-prettier I need a way to create a snapshot and not use 7-bit ascii (it's too much of a hassle to make work because a lot of the code in prettier has special characters). It seems, that executing the script in the runtime like so works in my situation:

https://github.com/dprint/dprint-plugin-prettier/pull/47/files#diff-d0d98998092552a1d3259338c2c71e118a5b8343dd4703c0c7f552ada7f9cb42R91-R96

...but to re-use the `create_snapshot` function, I need a hook to make modifications to the runtime.


mmastrac:
@dsherret I wonder if we should allow for UTF-16-encoded scripts to bypass this issue. I suspect we can keep most of the performance gains.

@nayeemrmn, what do you think about this? Perhaps we could augment FastString to fall back to a new `StaticTwoByte` if the strings are not ASCII?

core/runtime/snapshot_util.rs
```diff
@@ -22,6 +22,7 @@ pub struct CreateSnapshotOptions {
   pub extensions: Vec<Extension>,
   pub compression_cb: Option<Box<CompressionCb>>,
   pub snapshot_module_load_cb: Option<ExtModuleLoaderCb>,
+  pub with_runtime_cb: Option<Box<dyn Fn(&mut JsRuntimeForSnapshot)>>,
```
mmastrac, line 25: I'm good with this callback. One small nit -- can we name this type?

dsherret:
> @dsherret I wonder if we should allow for UTF-16-encoded scripts to bypass this issue. I suspect we can keep most of the performance gains.

I wouldn't spend too much effort fixing this for me, but it might be nice.

mmastrac (approved):
LGTM

nayeemrmn:
> Perhaps we could augment FastString to fall back to a new `StaticTwoByte` if the strings are not ASCII?

But for snapshotters who are using `include_js_for_snapshotting` it won't be static but loaded at `build.rs` runtime.

Maybe we should just make the ASCII debug assert opt-in as a feature.

PR #39, opened 2023-07-13
fix: better message when someone provides a module that was already included in a snapshot

dsherret:
I accidentally provided an extension again after creating a snapshot and it cost me a bit of time. This might help someone in the future.


mmastrac (approved):
LGTM!

dsherret:
Merged this for now, but might be reverted due to #7 (though I think we shouldn't do that, but I don't know too much about deno_core)

PR #40, opened 2023-07-17
feat(ops): op2 async support (slow path only)

mmastrac:
This implements:

 - #[op2(async)] for both `fn -> impl Future` and `async fn`.
 - Fastcalls are not (yet) supported
 - Result<impl Future<...>> is not yet supported


core/runtime/ops.rs
```diff
@@ -143,6 +143,84 @@ pub fn queue_async_op<'s>(
   None
 }
 
+#[inline]
+pub fn map_async_op_infallible<R: 'static>(
+  ctx: &OpCtx,
+  promise_id: i32,
```
bartlomieju, line 149: Are these two wrappers temporary or will they still be used when `op` gets removed?

core/runtime/ops.rs
```diff
@@ -512,6 +613,64 @@ mod tests {
...
+  /// Run a test for a single op.
+  async fn run_async_test(
+    repeat: usize,
+    op: &str,
+    test: &str,
+  ) -> Result<(), AnyError> {
+    let mut runtime = JsRuntime::new(RuntimeOptions {
+      extensions: vec![testing::init_ops_and_esm()],
+      ..Default::default()
+    });
```
bartlomieju, line 625: Random thought: we do this a lot, maybe we should add `RuntimeOptions::with_extensions()` to remove some of these verbose lines?

core/runtime/ops.rs
```diff
@@ -1165,4 +1324,115 @@ mod tests {
...
+
+  #[tokio::test]
+  pub async fn test_op_async_sleep() -> Result<(), Box<dyn std::error::Error>> {
+    run_async_test(5, "op_async_sleep", "await op_async_sleep()").await?;
+    run_async_test(5, "op_async_sleep_impl", "await op_async_sleep_impl()")
+      .await?;
+    Ok(())
+  }
+
+  #[op2(core, async)]
```
bartlomieju, line 1383: Random question: does the order of attributes matter here? Ie. can we do `op2(async, core)` as well?

ops/compile_test_runner/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+
+[package]
+name = "deno_ops_compile_test_runner"
+version = "0.1.0"
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+readme = "README.md"
+repository.workspace = true
+description = "Proc macro for writing Deno Ops"
```
bartlomieju, line 10: Nitpick: looks like the description wasn't updated. Also can we somehow mark this crate so it doesn't get published by accident?

ops/op2/README.md
```diff
@@ -15,6 +15,26 @@ string. At this time there is no way to avoid this copy, though the `op` code
...
+and
+
+```rust
+fn op_xyz(...) -> impl Future<Output = X> {}
+```
+
+These are desugared to a function that looks like so:
+
+```rust
+fn op_xyz(promise_id: u32, ...) -> Option<X> {}
```
bartlomieju, line 35: I don't understand it - they get desugared into something that doesn't return Future at all?

ops/op2/dispatch_async.rs
```diff
@@ -0,0 +1,163 @@
...
+  let deno_core = &generator_state.deno_core;
+
+  if matches!(
+    signature.ret_val,
+    RetVal::ResultFuture(_) | RetVal::ResultFutureResult(_)
+  ) {
+    output.extend(quote! {
+      let Ok(#result) = #result else {
+        // Handle eager error -- this will leave only a Future<R> or Future<Result<R>>
+        // ...
```
bartlomieju, line 99: Is this a missing TODO here?

bartlomieju (approved):
A few questions, but not a blocker to land and iterate further

core/runtime/ops.rs
mmastrac: These will replace the other forms in the file (`map_async_opXXX` and `queue_fast_async_op`/`queue_async_op`), but I'll remove OpResult at that time.

core/runtime/ops.rs
mmastrac: I considered requiring them to be alphabetically sorted but didn't get to it in this PR. I think now that you've pointed it out, I'll make that a requirement (and print the expected order if it is not correct).

ops/op2/README.md
mmastrac: Correct...  it's because it returns X if the eager poll was successful, and feeds the future into pending ops otherwise. I'll clarify this text here because it's not clear.

PR #41, opened 2023-07-17
feat(realms): Add support for modules in realms.

andreubotella:
TODO: Detailed description.

This PR is a port of denoland/deno#15760.

Towards denoland/deno#13239.


bartlomieju:
Alright, I think we can land this PR now :) awesome work @andreubotella.

@nayeemrmn please take a look as well

core/runtime/jsruntime.rs
```diff
@@ -883,7 +863,8 @@ impl JsRuntime {
       JsRealm::new(realm)
     };
 
-    self.init_extension_js(&realm, None)?;
+    // TODO(andreubotella): Should preserve_snapshotted_modules be in CreateRealmOptions?
+    self.init_extension_js(&realm, None, None)?;
```
nayeemrmn, line 867: Is there a problem with using `preserve_snapshotted_modules` from the runtime options on each realm? Since realms apply the snapshot implicitly and this is just another snapshot option.

nayeemrmn (approved):
LGTM

core/runtime/jsruntime.rs
andreubotella: I don't really have much of a clear idea at this point of what `preserve_snapshotted_modules` does in detail or how it's supposed to be used, other than that `deno_runtime` uses it for `node:` modules. I don't know if it makes sense to keep the options used at runtime initialization.

andreubotella: While trying to understand how the snapshotted modules interact with the embedder-provided module loader, I noticed that currently they are not included at all in the realm's module map:

https://github.com/denoland/deno_core/blob/main/core/runtime/jsruntime.rs#L871

(that was PR #45, which was split off this one)

I will update this PR to include those snapshotted modules in the module map, and to support `preserve_snapshotted_modules`.

andreubotella: I think this should probably be a follow-up PR though.

andreubotella: I will revert the change that moves the code that handles `preserved_snapshotted_modules` inside `init_extension_js`, since it is now a lot clearer to me that it doesn't belong there. I will handle that in `new_inner` and `create_realm` in the follow-up PR.

bartlomieju (approved):
LGTM!

PR #43, opened 2023-07-18
feat(ops): Benchmark sync ops, support for Option<numeric>, faster strings

mmastrac:
Approximately 25% faster strings from `op` -> `op2`.

### DIFF
--- a/core/runtime/jsruntime.rs
+++ b/core/runtime/jsruntime.rs
@@ -350,6 +350,7 @@ impl JsRuntimeState {
 fn v8_init(
   v8_platform: Option<v8::SharedRef<v8::Platform>>,
   predictable: bool,
+  expose_natives: bool,
 ) {
   // Include 10MB ICU data file.
   #[repr(C, align(16))]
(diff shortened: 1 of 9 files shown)


core/runtime/jsruntime.rs
```diff
@@ -464,7 +474,11 @@ pub struct CreateRealmOptions {
 impl JsRuntime {
   /// Only constructor, configuration is done through `options`.
   pub fn new(mut options: RuntimeOptions) -> JsRuntime {
-    JsRuntime::init_v8(options.v8_platform.take(), cfg!(test));
+    JsRuntime::init_v8(
+      options.v8_platform.take(),
+      cfg!(test),
+      options.unsafe_expose_natives_and_gc,
```
bartlomieju, line 480: Could you limit it to `cfg!(test)` and unsafe_expose_natives_and_gc - let's not expose a footgun to users

mmastrac: #[cfg(test)] doesn't seem to work but I'll investigate options.

bartlomieju: How about:
```
JsRuntime::init_v8(
      options.v8_platform.take(),
      cfg!(test),
      cfg!(test) && options.unsafe_expose_natives_and_gc,
)
```

bartlomieju: Sounds good, it's a reasonable solution

core/runtime/ops.rs
```diff
@@ -19,6 +19,11 @@ use std::option::Option;
 use std::task::Context;
 use std::task::Poll;
 
+/// The default string buffer size on the stack that prevents mallocs in some
+/// string functions. Keep in mind that Windows only offers 1MB stacks by default,
+/// so this is a limited resource!
+pub const STRING_STACK_BUFFER_SIZE: usize = 1024;
```
bartlomieju, line 25: Awesome, really easy to tweak now!

bartlomieju (approved):
LGTM, fantastic results

PR #53, opened 2023-07-19
refactor: rewrite more ops to op2 macro

bartlomieju:


aapoalas (approved):

mmastrac (approved):
LGTM

aapoalas:
Hmm... Although that's what we've been doing for this whole time and in that sense it makes sense, I think we might have a bit of a problem with fast calls and void returns: The slow call will return null for a void return, but the fast call will return undefined.

aapoalas (approved):

bartlomieju:
@aapoalas we added a TODO. We will revisit this topic again once we make more progress on decoupling `serde_v8`.

PR #54, opened 2023-07-19
feat(ops): op2 support for Result<impl Future<...>> 

mmastrac:
Supports `Result<impl Future<...>>`-style async ops like so:

```rust
  #[op2(async, core)]
  pub fn op_async_result_impl(
    mode: u8,
  ) -> Result<impl Future<Output = Result<(), Error>>, Error> {
    if mode == 0 {
      return Err(generic_error("early exit"));
    }
    return Ok(async move {
      if mode == 1 {
        return Err(generic_error("early async exit"));
      }
      tokio::time::sleep(Duration::from_millis(500)).await;
      if mode == 2 {
        return Err(generic_error("late async exit"));
      }
      return Ok(());
    });
  }
```

### DIFF
--- a/core/runtime/ops.rs
+++ b/core/runtime/ops.rs
@@ -1430,7 +1430,7 @@ mod tests {
     if mode == 0 {
       return Err(generic_error("early exit"));
     }
-    return Ok(async move {
+    Ok(async move {
       if mode == 1 {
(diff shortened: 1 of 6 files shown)


bartlomieju (approved):
LGTM, provided the tests pass once lint error is fixed

PR #56, opened 2023-07-19
fix(modules): check duplicate loads for aliases

nayeemrmn:
Supersedes #14. Checking the module map in load futures no longer works without borrow issues since the modules-in-realms work.

Normally `RecursiveModuleLoad::visited` prevents duplicate loads from happening, but it doesn't work for redirected/aliased modules for when both the alias and actual specifiers are used.

This adds `RecursiveModuleLoad::visited_by_alias` which targets these cases and is checked just before the loader is invoked.


bartlomieju (approved):
LGTM, nice fix!

PR #57, opened 2023-07-19
refactor(ops): op2 cleanup -- using macros to simplify code 

mmastrac:
Experiment with using a macro to reduce a lot of quoting and borrowing boilerplate.


dsherret (approved):
LGTM

PR #58, opened 2023-07-19
feat(ops): op2 support for v8::Global<...> 

mmastrac:
This adds support for v8 globals, along with a benchmark. `v8::Global`s are constructed from a `*mut Isolate` that is stashed in `OpCtx`, making this faster than when using a scope.

```
test bench_op_v8_global                     ... bench:      37,070 ns/iter (+/- 7,077)
test bench_op_v8_global_scope               ... bench:      44,636 ns/iter (+/- 1,668)
test bench_op_v8_local                      ... bench:       3,289 ns/iter (+/- 75)
test bench_op_v8_local_nofast               ... bench:       6,209 ns/iter (+/- 196)
test bench_op_v8_local_scope                ... bench:      12,640 ns/iter (+/- 440)
```

In addition, string methods that can use the isolate to skip scope creation have been updated.


core/ops_builtin_v8.rs
```diff
@@ -38,9 +38,8 @@ pub fn op_unref_op(scope: &mut v8::HandleScope, promise_id: i32) {
 #[op2(core)]
 pub fn op_set_promise_reject_callback<'a>(
   scope: &mut v8::HandleScope<'a>,
-  cb: v8::Local<'a, v8::Function>,
+  #[global] cb: v8::Global<v8::Function>,
```
bartlomieju, line 41: Very nice :)

core/runtime/ops.rs
```diff
@@ -393,6 +393,17 @@ pub fn to_string_ptr(string: &v8::fast_api::FastApiOneByteString) -> String {
   }
 }
 
+/// Converts a [`v8::String`] to an owned string.
```
bartlomieju, line 396:
```suggestion
/// Converts a [`v8::Value`] to an owned string.
```

bartlomieju (approved):

PR #61, opened 2023-07-20
fix(ops): Terminate scope of args and temporaries in op2 to avoid borrow errors

mmastrac:
Instead of:

```rust
let arg0 = ...;
let arg1 = ...;
let result = Self::call(...);
```

We now do:

```rust
let result = {
  let arg0 = ...;
  let arg1 = ...;
  Self::call(...)
}
```

This ensures that the scope of all args and temporaries ends after the call to `Self::call`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -236,7 +236,7 @@ checksum = "3c65c2ffdafc1564565200967edc4851c7b55422d3913466688907efd05ea26f"
 dependencies = [
  "deno-proc-macro-rules-macros",
  "proc-macro2",
- "syn 2.0.22",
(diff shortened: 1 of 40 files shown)


bartlomieju (approved):
LGTM

mmastrac:
Instead of:

```rust
let arg0 = ...;
let arg1 = ...;
let result = Self::call(...);
```

We now do:

```rust
let result = {
  let arg0 = ...;
  let arg1 = ...;
  Self::call(...)
}
```

This ensures that the scope of all args and temporaries ends after the call to `Self::call`.

PR #68, opened 2023-07-21
fix(modules): read symlinks when resolving files

### LINKED ISSUE #69: Doesn’t load symlinks properly

roj1512:
This pull request attempts to address #69 by looking into the resolved file paths to see if they are symlinks, and reassigning the resolved file path to where the symlink is heading to.

Closes #69.


roj1512:
If we were to go this way, testing this would require actual files. In that case, could you also tell me where should I create those files, please?

Also, how should I build the CLI with my changes of deno_core?

nayeemrmn:
This fix belongs in the CLI, core mostly has no concept of the file system

roj1512:
Thanks for your help, @nayeemrmn!

PR #73, opened 2023-07-22
feat(ops): op2 support for u64/i64 return values

mmastrac:
We can support u64/i64 in the return value position, though due to an unknown bug in fastcalls (unsure if it is in rusty_v8 or deno_core), we cannot make them fast yet. In anticipation that this will be fixed at some point, we start using `FastCall::new_with_bigint`.

Using a 64-bit value requires `#[bigint]` to remind implementers that this is more than just a simple number and may have some performance implications.


bartlomieju (approved):
LGTM :)

PR #78, opened 2023-07-23
refactor: don't use builder in extension!()

### LINKED ISSUE #49: Use const `OpDecl` arrays in extensions
With the work by @littledivy and others, we finally have a fully-`const` `OpDecl` that is `Copy`. This should allow us to generate `const` arrays of ops in extensions, and reduce the number of `Vec` allocations at startup.

### LINKED ISSUE #50: Build a const struct from the `extension!` macro
We should have enough infrastructure in place that the `extension!` macro can generate a const struct that can be fully computed at compile-time. If not, let's make the rest of this struct const-compatible and then deprecate and remove the `ExtensionBuilder`.

nayeemrmn:
To address #49 and #50 while still allowing opt-in dynamic extensions and `customizer`s, this PR changes fields like:
```rust
pub struct Extension {
  js_files: Vec<ExtensionFileSource>,
  esm_files: Vec<ExtensionFileSource>,
  ops: Option<Vec<OpDecl>>,
}
```
to:
```rust
pub struct Extension {
  js_files: Cow<'static, [ExtensionFileSource]>,
  esm_files: Cow<'static, [ExtensionFileSource]>,
  ops: Cow<'static, [OpDecl]>,
}
```
The `extension!()` macro no longer uses `ExtensionBuilder` but a `mut Extension`. This is initialized at runtime with borrows of generated static arrays of ops, sources etc.. For bog-standard extensions there will be no further allocations for these arrays. But customizers or otherwise dynamic usage can opt into mutations by calling `.to_mut()` on them.

Closes #49.
Re #50, I don't think it's possible that `Extension`s as passed into `RuntimeOptions` can be fully `const`. At minimum there needs to be some `init_*()` function which accepts custom parameters as specified in `extension!( options = {} )` at runtime and collapses that into an `opstate_fn` closure. Extracting that out wouldn't be any better than this solution.


nayeemrmn:
I think the only notable API breakage is that customizers like:
```rust
  customizer = |ext: &mut deno_core::ExtensionBuilder| {
    ext.esm(vec![ExtensionFileSource {
      specifier: "ext:cli/runtime/js/99_main.js",
      code: ExtensionFileSourceCode::LoadedFromFsDuringSnapshot(
        std::path::PathBuf::from(deno_runtime::js::PATH_FOR_99_MAIN_JS),
      ),
    }]);
  }
```
must be rewritten to:
```rust
  customizer = |ext: &mut deno_core::Extension| {
    ext.esm_files.to_mut().push(ExtensionFileSource {
      specifier: "ext:cli/runtime/js/99_main.js",
      code: ExtensionFileSourceCode::LoadedFromFsDuringSnapshot(
        std::path::PathBuf::from(deno_runtime::js::PATH_FOR_99_MAIN_JS),
      ),
    });
  }
```
Also a consequence of not using the builder in the macro is that all of `Extension`'s fields have to be made public. We could make some methods to hide stuff but IMO not a big issue either way.

nayeemrmn:
I consolidated `include_js_files!()` and `include_js_files_with_specifiers!()` into one, as well as the `esm_with_specifiers = [("node:assert", "assert.ts")]` syntax becoming `esm = ["assert.ts" with_specifier "node:assert"]`. That's another notable API change. (Essentially I had to because concatenating the two static slices from each at compile time is weirdly hard.)

CLI repo works with the following patch:
```diff
diff --git a/cli/build.rs b/cli/build.rs
index a71692a31..2a72d4c2d 100644
--- a/cli/build.rs
+++ b/cli/build.rs
@@ -317,13 +317,13 @@ deno_core::extension!(
     "40_testing.js",
     "99_main.js"
   ],
-  customizer = |ext: &mut deno_core::ExtensionBuilder| {
-    ext.esm(vec![ExtensionFileSource {
+  customizer = |ext: &mut deno_core::Extension| {
+    ext.esm_files.to_mut().push(ExtensionFileSource {
       specifier: "ext:cli/runtime/js/99_main.js",
       code: ExtensionFileSourceCode::LoadedFromFsDuringSnapshot(
-        std::path::PathBuf::from(deno_runtime::js::PATH_FOR_99_MAIN_JS),
+        deno_runtime::js::PATH_FOR_99_MAIN_JS,
       ),
-    }]);
+    });
   }
 );

diff --git a/ext/fs/lib.rs b/ext/fs/lib.rs
index 7ba6cd7ca..d27712927 100644
--- a/ext/fs/lib.rs
+++ b/ext/fs/lib.rs
@@ -19,7 +19,6 @@ use crate::ops::*;
 use deno_core::error::AnyError;
 use deno_core::OpState;
 use std::cell::RefCell;
-use std::convert::From;
 use std::path::Path;
 use std::rc::Rc;

diff --git a/ext/kv/lib.rs b/ext/kv/lib.rs
index db7c5c22e..a781f4579 100644
--- a/ext/kv/lib.rs
+++ b/ext/kv/lib.rs
@@ -8,7 +8,6 @@ use std::borrow::Cow;
 use std::cell::RefCell;
 use std::num::NonZeroU32;
 use std::rc::Rc;
-use std::vec;

 use codec::decode_key;
 use codec::encode_key;
diff --git a/ext/node/lib.rs b/ext/node/lib.rs
index 89ee87cd3..8f910ac63 100644
--- a/ext/node/lib.rs
+++ b/ext/node/lib.rs
@@ -446,58 +446,55 @@ deno_core::extension!(deno_node,
     "path/separator.ts",
     "readline/promises.ts",
     "repl.ts",
-    "wasi.ts"
-  ],
-  esm_with_specifiers = [
-    dir "polyfills",
-    ("node:assert", "assert.ts"),
-    ("node:assert/strict", "assert/strict.ts"),
-    ("node:async_hooks", "async_hooks.ts"),
-    ("node:buffer", "buffer.ts"),
-    ("node:child_process", "child_process.ts"),
-    ("node:cluster", "cluster.ts"),
-    ("node:console", "console.ts"),
-    ("node:constants", "constants.ts"),
-    ("node:crypto", "crypto.ts"),
-    ("node:dgram", "dgram.ts"),
-    ("node:diagnostics_channel", "diagnostics_channel.ts"),
-    ("node:dns", "dns.ts"),
-    ("node:dns/promises", "dns/promises.ts"),
-    ("node:domain", "domain.ts"),
-    ("node:events", "events.ts"),
-    ("node:fs", "fs.ts"),
-    ("node:fs/promises", "fs/promises.ts"),
-    ("node:http", "http.ts"),
-    ("node:http2", "http2.ts"),
-    ("node:https", "https.ts"),
-    ("node:module", "01_require.js"),
-    ("node:net", "net.ts"),
-    ("node:os", "os.ts"),
-    ("node:path", "path.ts"),
-    ("node:path/posix", "path/posix.ts"),
-    ("node:path/win32", "path/win32.ts"),
-    ("node:perf_hooks", "perf_hooks.ts"),
-    ("node:process", "process.ts"),
-    ("node:punycode", "punycode.ts"),
-    ("node:querystring", "querystring.ts"),
-    ("node:readline", "readline.ts"),
-    ("node:stream", "stream.ts"),
-    ("node:stream/consumers", "stream/consumers.mjs"),
-    ("node:stream/promises", "stream/promises.mjs"),
-    ("node:stream/web", "stream/web.ts"),
-    ("node:string_decoder", "string_decoder.ts"),
-    ("node:sys", "sys.ts"),
-    ("node:timers", "timers.ts"),
-    ("node:timers/promises", "timers/promises.ts"),
-    ("node:tls", "tls.ts"),
-    ("node:tty", "tty.ts"),
-    ("node:url", "url.ts"),
-    ("node:util", "util.ts"),
-    ("node:util/types", "util/types.ts"),
-    ("node:v8", "v8.ts"),
-    ("node:vm", "vm.ts"),
-    ("node:worker_threads", "worker_threads.ts"),
-    ("node:zlib", "zlib.ts"),
+    "wasi.ts",
+    "assert.ts" with_specifier "node:assert",
+    "assert/strict.ts" with_specifier "node:assert/strict",
+    "async_hooks.ts" with_specifier "node:async_hooks",
+    "buffer.ts" with_specifier "node:buffer",
+    "child_process.ts" with_specifier "node:child_process",
+    "cluster.ts" with_specifier "node:cluster",
+    "console.ts" with_specifier "node:console",
+    "constants.ts" with_specifier "node:constants",
+    "crypto.ts" with_specifier "node:crypto",
+    "dgram.ts" with_specifier "node:dgram",
+    "diagnostics_channel.ts" with_specifier "node:diagnostics_channel",
+    "dns.ts" with_specifier "node:dns",
+    "dns/promises.ts" with_specifier "node:dns/promises",
+    "domain.ts" with_specifier "node:domain",
+    "events.ts" with_specifier "node:events",
+    "fs.ts" with_specifier "node:fs",
+    "fs/promises.ts" with_specifier "node:fs/promises",
+    "http.ts" with_specifier "node:http",
+    "http2.ts" with_specifier "node:http2",
+    "https.ts" with_specifier "node:https",
+    "01_require.js" with_specifier "node:module",
+    "net.ts" with_specifier "node:net",
+    "os.ts" with_specifier "node:os",
+    "path.ts" with_specifier "node:path",
+    "path/posix.ts" with_specifier "node:path/posix",
+    "path/win32.ts" with_specifier "node:path/win32",
+    "perf_hooks.ts" with_specifier "node:perf_hooks",
+    "process.ts" with_specifier "node:process",
+    "punycode.ts" with_specifier "node:punycode",
+    "querystring.ts" with_specifier "node:querystring",
+    "readline.ts" with_specifier "node:readline",
+    "stream.ts" with_specifier "node:stream",
+    "stream/consumers.mjs" with_specifier "node:stream/consumers",
+    "stream/promises.mjs" with_specifier "node:stream/promises",
+    "stream/web.ts" with_specifier "node:stream/web",
+    "string_decoder.ts" with_specifier "node:string_decoder",
+    "sys.ts" with_specifier "node:sys",
+    "timers.ts" with_specifier "node:timers",
+    "timers/promises.ts" with_specifier "node:timers/promises",
+    "tls.ts" with_specifier "node:tls",
+    "tty.ts" with_specifier "node:tty",
+    "url.ts" with_specifier "node:url",
+    "util.ts" with_specifier "node:util",
+    "util/types.ts" with_specifier "node:util/types",
+    "v8.ts" with_specifier "node:v8",
+    "vm.ts" with_specifier "node:vm",
+    "worker_threads.ts" with_specifier "node:worker_threads",
+    "zlib.ts" with_specifier "node:zlib",
   ],
   options = {
     maybe_npm_resolver: Option<NpmResolverRc>,
@@ -516,7 +513,7 @@ deno_core::extension!(deno_node,
   },
   global_template_middleware = global_template_middleware,
   global_object_middleware = global_object_middleware,
-  customizer = |ext: &mut deno_core::ExtensionBuilder| {
+  customizer = |ext: &mut deno_core::Extension| {
     let mut external_references = Vec::with_capacity(7);

     global::GETTER_MAP_FN.with(|getter| {
@@ -554,7 +551,7 @@ deno_core::extension!(deno_node,
         named_getter: *descriptor,
       });
     });
-    ext.external_references(external_references);
+    ext.external_references.to_mut().extend(external_references);
   },
 );

diff --git a/runtime/build.rs b/runtime/build.rs
index 6fde45377..38218a439 100644
--- a/runtime/build.rs
+++ b/runtime/build.rs
@@ -275,14 +275,14 @@ mod startup_snapshot {
   deno_core::extension!(
     runtime_main,
     deps = [runtime],
-    customizer = |ext: &mut deno_core::ExtensionBuilder| {
-      ext.esm(vec![ExtensionFileSource {
+    customizer = |ext: &mut deno_core::Extension| {
+      ext.esm_files.to_mut().push(ExtensionFileSource {
         specifier: "ext:runtime_main/js/99_main.js",
         code: deno_core::ExtensionFileSourceCode::IncludedInBinary(
           include_str!("js/99_main.js"),
         ),
-      }]);
-      ext.esm_entry_point("ext:runtime_main/js/99_main.js");
+      });
+      ext.esm_entry_point = Some("ext:runtime_main/js/99_main.js");
     }
   );

@@ -290,8 +290,8 @@ mod startup_snapshot {
   deno_core::extension!(
     runtime_main,
     deps = [runtime],
-    customizer = |ext: &mut deno_core::ExtensionBuilder| {
-      ext.esm_entry_point("ext:runtime/90_deno_ns.js");
+    customizer = |ext: &mut deno_core::Extension| {
+      ext.esm_entry_point = Some("ext:runtime/90_deno_ns.js");
     }
   );
```

bartlomieju (commented):
This looks good to me and the changes to CLI are reasonable. I'll wait on Matt's review before approving.

core/extensions.rs
```diff
@@ -257,120 +273,107 @@ macro_rules! extension {
...
+          esm_files: std::borrow::Cow::Borrowed(&$crate::or!($($crate::include_js_files!( $name $( dir $dir_esm , )? $( $esm $( with_specifier $esm_specifier )? , )* ))?, [])),
+          esm_entry_point: $crate::or!($(Some($esm_entry_point))?, None),
+          ops: std::borrow::Cow::Borrowed(&[$($(
+            $( #[ $m ] )*
+            $( $op )::+ $( :: < $($op_param),* > )? :: DECL
+          ),+)?]),
+          external_references: std::borrow::Cow::Borrowed(&[ $( $external_reference ),* ]),
+          // Computed at runtime:
+          opstate_fn: None,
+          middleware_fn: None,
```
mmastrac, line 293: Looking at this more closely, these should probably also be converted to `fn()`. What do you think?

core/extensions.rs
```diff
@@ -389,44 +392,43 @@ macro_rules! extension {
...
-  global_template_middleware: Option<Box<GlobalTemplateMiddlewareFn>>,
-  global_object_middleware: Option<Box<GlobalObjectMiddlewareFn>>,
-  external_references: Option<Vec<v8::ExternalReference<'static>>>,
-  initialized: bool,
-  enabled: bool,
-  deps: Option<&'static [&'static str]>,
+  pub name: &'static str,
+  pub deps: &'static [&'static str],
+  pub js_files: Cow<'static, [ExtensionFileSource]>,
+  pub esm_files: Cow<'static, [ExtensionFileSource]>,
```
mmastrac, line 421: Do we modify these anywhere?

core/extensions.rs
```diff
@@ -195,6 +195,23 @@ macro_rules! ops {
   }
 }
 
+#[macro_export]
+macro_rules! or {
```
mmastrac, line 199: Love it. Can you add a one-line doc to this macro?

core/extensions.rs
```diff
@@ -195,6 +195,23 @@ macro_rules! ops {
...
+  ($e:expr, $fallback:expr) => {
+    $e
+  };
+  (, $fallback:expr) => {
+    $fallback
+  };
+}
+
+#[macro_export]
+macro_rules! concat_slices {
```
mmastrac, line 209: Is this used? I wonder if it might be more appropriate to call it "concat_array_literals" instead, however.

core/extensions.rs
```diff
@@ -389,44 +392,43 @@ macro_rules! extension {
...
+  pub deps: &'static [&'static str],
+  pub js_files: Cow<'static, [ExtensionFileSource]>,
+  pub esm_files: Cow<'static, [ExtensionFileSource]>,
+  pub esm_entry_point: Option<&'static str>,
+  pub ops: Cow<'static, [OpDecl]>,
+  pub external_references: Cow<'static, [v8::ExternalReference<'static>]>,
+  pub opstate_fn: Option<Box<OpStateFn>>,
+  pub middleware_fn: Option<Box<OpMiddlewareFn>>,
+  pub enabled: bool,
+  pub js_enabled: bool,
```
mmastrac, line 428: `js_enabled` might need a better name and/or some docs. Perhaps `will_contribute_js_to_runtime`? It isn't visible to anything but user code.

mmastrac (commented):
Looking good, but a few small changes needed.

mmastrac:
> Re https://github.com/denoland/deno_core/issues/50, I don't think it's possible that Extensions as passed into RuntimeOptions can be fully const. At minimum there needs to be some init_*() function which accepts custom parameters as specified in extension!( options = {} ) at runtime and collapses that into an opstate_fn closure. Extracting that out wouldn't be any better than this solution.

We could probably make the opstate_fn a `*const ()` and then cast as needed. As long as we ensure that deno_core is the only caller of this function, it will be safe enough.

core/extensions.rs
nayeemrmn: No it's not used, I forgot to remove it

core/extensions.rs
nayeemrmn: Might not need this after all, maybe I can just do `ext.esm_files = Cow::Borrowed(&[])` etc. in place of `js_enabled = false`.

nayeemrmn:
> We could probably make the opstate_fn a `*const ()` and then cast as needed. As long as we ensure that deno_core is the only caller of this function, it will be safe enough.

Do you mean assigning the closure at runtime to a static slot? Keep in mind the same extension may be `init_*`ed for multiple runtimes with different options i.e. web workers. If I'm not understanding correctly please give an example of what the user's `init_*()` or option passing should look like.

mmastrac:
> Do you mean assigning the closure at runtime to a static slot? Keep in mind the same extension may be init_*ed for multiple runtimes with different options i.e. web workers. If I'm not understanding correctly please give an example of what the user's init_*() or option passing should look like.

I thought through it a bit and there's a few complications here so I'm OK to punt on it for now, but we have a couple of options that require a bit more work than would be scoped for this PR:

1. We can make opstate_fn a `const` function pointer that returns a boxed closure (`fn(param_a, param_b, param_b) -> Box<Fn<&mut OpState>>`). The opstate_fn function pointer itself could be `const` and cast to a `*const ()` (which gets us one step closer to fully const extensions).

2. We can initialize OpState earlier as part of the ExtensionSet work so that we don't have to close over the configuration parameters. The `OpState` function would be a `const` function pointer of the form `fn(&mut OpState, param_a, param_b, param_b)`.  I prefer this one because it means we can avoid quite a few intermediate allocations at startup.

Let's wait on this one and see if there's an ergonomic opportunity to provide the `GothamState` with the `ExtensionSet`.

nayeemrmn:
I'll deprecate `ExtensionBuilder` in this PR also. Since `Extension`s fields are public it provides no value now.

nayeemrmn:
<details>
<summary>
Updated CLI patch with deprecated builder:
</summary>

```diff
diff --git a/bench_util/README.md b/bench_util/README.md
index 0ec15f0d0..e50f518e2 100644
--- a/bench_util/README.md
+++ b/bench_util/README.md
@@ -7,26 +7,18 @@ use deno_bench_util::bench_or_profile;
 use deno_bench_util::bencher::{benchmark_group, Bencher};
 use deno_bench_util::bench_js_sync};

-use deno_core::op_sync;
-use deno_core::serialize_op_result;
 use deno_core::Extension;
-use deno_core::JsRuntime;
-use deno_core::Op;
-use deno_core::OpState;
+
+#[op]
+fn op_nop() -> usize {
+  9
+}

 fn setup() -> Vec<Extension> {
-  let custom_ext = Extension::builder()
-    .ops(vec![
-      ("op_nop", |state, _| {
-        Op::Sync(serialize_op_result(Ok(9), state))
-      }),
-    ])
-    .build();
-
-  vec![
-    // deno_{ext}::init(...),
-    custom_ext,
-  ]
+  vec![Extension {
+    name: "my_ext"
+    ops: std::borrow::Cow::Borrowed(&[op_nop::DECL])
+  }]
 }

 fn bench_op_nop(b: &mut Bencher) {
diff --git a/bench_util/benches/utf8.rs b/bench_util/benches/utf8.rs
index 9bc7cdaee..242fb2bfc 100644
--- a/bench_util/benches/utf8.rs
+++ b/bench_util/benches/utf8.rs
@@ -10,21 +10,23 @@ use deno_core::ExtensionFileSource;
 use deno_core::ExtensionFileSourceCode;

 fn setup() -> Vec<Extension> {
-  vec![Extension::builder("bench_setup")
-    .js(vec![ExtensionFileSource {
+  vec![Extension {
+    name: "bench_setup",
+    js_files: std::borrow::Cow::Borrowed(&[ExtensionFileSource {
       specifier: "ext:bench_setup/setup.js",
       code: ExtensionFileSourceCode::IncludedInBinary(
         r#"
-      const hello = "hello world\n";
-      const hello1k = hello.repeat(1e3);
-      const hello1m = hello.repeat(1e6);
-      const helloEncoded = Deno.core.encode(hello);
-      const hello1kEncoded = Deno.core.encode(hello1k);
-      const hello1mEncoded = Deno.core.encode(hello1m);
+        const hello = "hello world\n";
+        const hello1k = hello.repeat(1e3);
+        const hello1m = hello.repeat(1e6);
+        const helloEncoded = Deno.core.encode(hello);
+        const hello1kEncoded = Deno.core.encode(hello1k);
+        const hello1mEncoded = Deno.core.encode(hello1m);
       "#,
       ),
-    }])
-    .build()]
+    }]),
+    ..Default::default()
+  }]
 }

 fn bench_utf8_encode_12_b(b: &mut Bencher) {
diff --git a/cli/build.rs b/cli/build.rs
index 8d379f604..9860412a1 100644
--- a/cli/build.rs
+++ b/cli/build.rs
@@ -317,13 +317,13 @@ deno_core::extension!(
     "40_testing.js",
     "99_main.js"
   ],
-  customizer = |ext: &mut deno_core::ExtensionBuilder| {
-    ext.esm(vec![ExtensionFileSource {
+  customizer = |ext: &mut deno_core::Extension| {
+    ext.esm_files.to_mut().push(ExtensionFileSource {
       specifier: "ext:cli/runtime/js/99_main.js",
       code: ExtensionFileSourceCode::LoadedFromFsDuringSnapshot(
-        std::path::PathBuf::from(deno_runtime::js::PATH_FOR_99_MAIN_JS),
+        deno_runtime::js::PATH_FOR_99_MAIN_JS,
       ),
-    }]);
+    });
   }
 );

diff --git a/ext/fs/lib.rs b/ext/fs/lib.rs
index 7ba6cd7ca..d27712927 100644
--- a/ext/fs/lib.rs
+++ b/ext/fs/lib.rs
@@ -19,7 +19,6 @@ use crate::ops::*;
 use deno_core::error::AnyError;
 use deno_core::OpState;
 use std::cell::RefCell;
-use std::convert::From;
 use std::path::Path;
 use std::rc::Rc;

diff --git a/ext/kv/lib.rs b/ext/kv/lib.rs
index db7c5c22e..a781f4579 100644
--- a/ext/kv/lib.rs
+++ b/ext/kv/lib.rs
@@ -8,7 +8,6 @@ use std::borrow::Cow;
 use std::cell::RefCell;
 use std::num::NonZeroU32;
 use std::rc::Rc;
-use std::vec;

 use codec::decode_key;
 use codec::encode_key;
diff --git a/ext/node/lib.rs b/ext/node/lib.rs
index 89ee87cd3..8f910ac63 100644
--- a/ext/node/lib.rs
+++ b/ext/node/lib.rs
@@ -446,58 +446,55 @@ deno_core::extension!(deno_node,
     "path/separator.ts",
     "readline/promises.ts",
     "repl.ts",
-    "wasi.ts"
-  ],
-  esm_with_specifiers = [
-    dir "polyfills",
-    ("node:assert", "assert.ts"),
-    ("node:assert/strict", "assert/strict.ts"),
-    ("node:async_hooks", "async_hooks.ts"),
-    ("node:buffer", "buffer.ts"),
-    ("node:child_process", "child_process.ts"),
-    ("node:cluster", "cluster.ts"),
-    ("node:console", "console.ts"),
-    ("node:constants", "constants.ts"),
-    ("node:crypto", "crypto.ts"),
-    ("node:dgram", "dgram.ts"),
-    ("node:diagnostics_channel", "diagnostics_channel.ts"),
-    ("node:dns", "dns.ts"),
-    ("node:dns/promises", "dns/promises.ts"),
-    ("node:domain", "domain.ts"),
-    ("node:events", "events.ts"),
-    ("node:fs", "fs.ts"),
-    ("node:fs/promises", "fs/promises.ts"),
-    ("node:http", "http.ts"),
-    ("node:http2", "http2.ts"),
-    ("node:https", "https.ts"),
-    ("node:module", "01_require.js"),
-    ("node:net", "net.ts"),
-    ("node:os", "os.ts"),
-    ("node:path", "path.ts"),
-    ("node:path/posix", "path/posix.ts"),
-    ("node:path/win32", "path/win32.ts"),
-    ("node:perf_hooks", "perf_hooks.ts"),
-    ("node:process", "process.ts"),
-    ("node:punycode", "punycode.ts"),
-    ("node:querystring", "querystring.ts"),
-    ("node:readline", "readline.ts"),
-    ("node:stream", "stream.ts"),
-    ("node:stream/consumers", "stream/consumers.mjs"),
-    ("node:stream/promises", "stream/promises.mjs"),
-    ("node:stream/web", "stream/web.ts"),
-    ("node:string_decoder", "string_decoder.ts"),
-    ("node:sys", "sys.ts"),
-    ("node:timers", "timers.ts"),
-    ("node:timers/promises", "timers/promises.ts"),
-    ("node:tls", "tls.ts"),
-    ("node:tty", "tty.ts"),
-    ("node:url", "url.ts"),
-    ("node:util", "util.ts"),
-    ("node:util/types", "util/types.ts"),
-    ("node:v8", "v8.ts"),
-    ("node:vm", "vm.ts"),
-    ("node:worker_threads", "worker_threads.ts"),
-    ("node:zlib", "zlib.ts"),
+    "wasi.ts",
+    "assert.ts" with_specifier "node:assert",
+    "assert/strict.ts" with_specifier "node:assert/strict",
+    "async_hooks.ts" with_specifier "node:async_hooks",
+    "buffer.ts" with_specifier "node:buffer",
+    "child_process.ts" with_specifier "node:child_process",
+    "cluster.ts" with_specifier "node:cluster",
+    "console.ts" with_specifier "node:console",
+    "constants.ts" with_specifier "node:constants",
+    "crypto.ts" with_specifier "node:crypto",
+    "dgram.ts" with_specifier "node:dgram",
+    "diagnostics_channel.ts" with_specifier "node:diagnostics_channel",
+    "dns.ts" with_specifier "node:dns",
+    "dns/promises.ts" with_specifier "node:dns/promises",
+    "domain.ts" with_specifier "node:domain",
+    "events.ts" with_specifier "node:events",
+    "fs.ts" with_specifier "node:fs",
+    "fs/promises.ts" with_specifier "node:fs/promises",
+    "http.ts" with_specifier "node:http",
+    "http2.ts" with_specifier "node:http2",
+    "https.ts" with_specifier "node:https",
+    "01_require.js" with_specifier "node:module",
+    "net.ts" with_specifier "node:net",
+    "os.ts" with_specifier "node:os",
+    "path.ts" with_specifier "node:path",
+    "path/posix.ts" with_specifier "node:path/posix",
+    "path/win32.ts" with_specifier "node:path/win32",
+    "perf_hooks.ts" with_specifier "node:perf_hooks",
+    "process.ts" with_specifier "node:process",
+    "punycode.ts" with_specifier "node:punycode",
+    "querystring.ts" with_specifier "node:querystring",
+    "readline.ts" with_specifier "node:readline",
+    "stream.ts" with_specifier "node:stream",
+    "stream/consumers.mjs" with_specifier "node:stream/consumers",
+    "stream/promises.mjs" with_specifier "node:stream/promises",
+    "stream/web.ts" with_specifier "node:stream/web",
+    "string_decoder.ts" with_specifier "node:string_decoder",
+    "sys.ts" with_specifier "node:sys",
+    "timers.ts" with_specifier "node:timers",
+    "timers/promises.ts" with_specifier "node:timers/promises",
+    "tls.ts" with_specifier "node:tls",
+    "tty.ts" with_specifier "node:tty",
+    "url.ts" with_specifier "node:url",
+    "util.ts" with_specifier "node:util",
+    "util/types.ts" with_specifier "node:util/types",
+    "v8.ts" with_specifier "node:v8",
+    "vm.ts" with_specifier "node:vm",
+    "worker_threads.ts" with_specifier "node:worker_threads",
+    "zlib.ts" with_specifier "node:zlib",
   ],
   options = {
     maybe_npm_resolver: Option<NpmResolverRc>,
@@ -516,7 +513,7 @@ deno_core::extension!(deno_node,
   },
   global_template_middleware = global_template_middleware,
   global_object_middleware = global_object_middleware,
-  customizer = |ext: &mut deno_core::ExtensionBuilder| {
+  customizer = |ext: &mut deno_core::Extension| {
     let mut external_references = Vec::with_capacity(7);

     global::GETTER_MAP_FN.with(|getter| {
@@ -554,7 +551,7 @@ deno_core::extension!(deno_node,
         named_getter: *descriptor,
       });
     });
-    ext.external_references(external_references);
+    ext.external_references.to_mut().extend(external_references);
   },
 );

diff --git a/ext/url/benches/url_ops.rs b/ext/url/benches/url_ops.rs
index 835dfea2e..b65a0e345 100644
--- a/ext/url/benches/url_ops.rs
+++ b/ext/url/benches/url_ops.rs
@@ -13,17 +13,20 @@ fn setup() -> Vec<Extension> {
   vec![
     deno_webidl::deno_webidl::init_ops_and_esm(),
     deno_url::deno_url::init_ops_and_esm(),
-    Extension::builder("bench_setup")
-      .esm(vec![ExtensionFileSource {
+    Extension {
+      name: "bench_setup",
+      esm_files: std::borrow::Cow::Borrowed(&[ExtensionFileSource {
         specifier: "ext:bench_setup/setup",
         code: ExtensionFileSourceCode::IncludedInBinary(
-          r#"import { URL } from "ext:deno_url/00_url.js";
-        globalThis.URL = URL;
+          r#"
+          import { URL } from "ext:deno_url/00_url.js";
+          globalThis.URL = URL;
         "#,
         ),
-      }])
-      .esm_entry_point("ext:bench_setup/setup")
-      .build(),
+      }]),
+      esm_entry_point: Some("ext:bench_setup/setup"),
+      ..Default::default()
+    },
   ]
 }

diff --git a/ext/web/benches/encoding.rs b/ext/web/benches/encoding.rs
index 12fff5d75..fc4ab37fc 100644
--- a/ext/web/benches/encoding.rs
+++ b/ext/web/benches/encoding.rs
@@ -30,22 +30,24 @@ fn setup() -> Vec<Extension> {
       Default::default(),
       None,
     ),
-    Extension::builder("bench_setup")
-      .esm(vec![ExtensionFileSource {
+    Extension {
+      name: "bench_setup",
+      esm_files: std::borrow::Cow::Borrowed(&[ExtensionFileSource {
         specifier: "ext:bench_setup/setup",
         code: ExtensionFileSourceCode::IncludedInBinary(
           r#"
-        import { TextDecoder } from "ext:deno_web/08_text_encoding.js";
-        globalThis.TextDecoder = TextDecoder;
-        globalThis.hello12k = Deno.core.encode("hello world\n".repeat(1e3));
+          import { TextDecoder } from "ext:deno_web/08_text_encoding.js";
+          globalThis.TextDecoder = TextDecoder;
+          globalThis.hello12k = Deno.core.encode("hello world\n".repeat(1e3));
         "#,
         ),
-      }])
-      .state(|state| {
+      }]),
+      esm_entry_point: Some("ext:bench_setup/setup"),
+      op_state_fn: Some(Box::new(|state| {
         state.put(Permissions {});
-      })
-      .esm_entry_point("ext:bench_setup/setup")
-      .build(),
+      })),
+      ..Default::default()
+    },
   ]
 }

diff --git a/ext/web/benches/timers_ops.rs b/ext/web/benches/timers_ops.rs
index 9d74abd17..589fc8534 100644
--- a/ext/web/benches/timers_ops.rs
+++ b/ext/web/benches/timers_ops.rs
@@ -24,23 +24,28 @@ fn setup() -> Vec<Extension> {
     deno_webidl::deno_webidl::init_ops_and_esm(),
     deno_url::deno_url::init_ops_and_esm(),
     deno_console::deno_console::init_ops_and_esm(),
-    deno_web::deno_web::init_ops_and_esm::<Permissions>(Default::default(), None),
-    Extension::builder("bench_setup")
-    .esm(vec![
-      ExtensionFileSource {
+    deno_web::deno_web::init_ops_and_esm::<Permissions>(
+      Default::default(),
+      None,
+    ),
+    Extension {
+      name: "bench_setup",
+      esm_files: std::borrow::Cow::Borrowed(&[ExtensionFileSource {
         specifier: "ext:bench_setup/setup",
-        code: ExtensionFileSourceCode::IncludedInBinary(r#"
-      import { setTimeout, handleTimerMacrotask } from "ext:deno_web/02_timers.js";
-      globalThis.setTimeout = setTimeout;
-      Deno.core.setMacrotaskCallback(handleTimerMacrotask);
-      "#)
-      },
-    ])
-    .state(|state| {
-      state.put(Permissions{});
-    })
-    .esm_entry_point("ext:bench_setup/setup")
-    .build()
+        code: ExtensionFileSourceCode::IncludedInBinary(
+          r#"
+            import { setTimeout, handleTimerMacrotask } from "ext:deno_web/02_timers.js";
+            globalThis.setTimeout = setTimeout;
+            Deno.core.setMacrotaskCallback(handleTimerMacrotask);
+          "#,
+        ),
+      }]),
+      esm_entry_point: Some("ext:bench_setup/setup"),
+      op_state_fn: Some(Box::new(|state| {
+        state.put(Permissions {});
+      })),
+      ..Default::default()
+    },
   ]
 }

diff --git a/ext/webidl/benches/dict.rs b/ext/webidl/benches/dict.rs
index b3d95c8a3..e7f4d0300 100644
--- a/ext/webidl/benches/dict.rs
+++ b/ext/webidl/benches/dict.rs
@@ -12,15 +12,17 @@ use deno_core::ExtensionFileSourceCode;
 fn setup() -> Vec<Extension> {
   vec![
     deno_webidl::deno_webidl::init_ops_and_esm(),
-    Extension::builder("deno_webidl_bench")
-      .esm(vec![ExtensionFileSource {
+    Extension {
+      name: "deno_webidl_bench",
+      esm_files: std::borrow::Cow::Borrowed(&[ExtensionFileSource {
         specifier: "ext:deno_webidl_bench/setup.js",
         code: ExtensionFileSourceCode::IncludedInBinary(include_str!(
           "dict.js"
         )),
-      }])
-      .esm_entry_point("ext:deno_webidl_bench/setup.js")
-      .build(),
+      }]),
+      esm_entry_point: Some("ext:deno_webidl_bench/setup.js"),
+      ..Default::default()
+    },
   ]
 }

diff --git a/runtime/build.rs b/runtime/build.rs
index 6fde45377..38218a439 100644
--- a/runtime/build.rs
+++ b/runtime/build.rs
@@ -275,14 +275,14 @@ mod startup_snapshot {
   deno_core::extension!(
     runtime_main,
     deps = [runtime],
-    customizer = |ext: &mut deno_core::ExtensionBuilder| {
-      ext.esm(vec![ExtensionFileSource {
+    customizer = |ext: &mut deno_core::Extension| {
+      ext.esm_files.to_mut().push(ExtensionFileSource {
         specifier: "ext:runtime_main/js/99_main.js",
         code: deno_core::ExtensionFileSourceCode::IncludedInBinary(
           include_str!("js/99_main.js"),
         ),
-      }]);
-      ext.esm_entry_point("ext:runtime_main/js/99_main.js");
+      });
+      ext.esm_entry_point = Some("ext:runtime_main/js/99_main.js");
     }
   );

@@ -290,8 +290,8 @@ mod startup_snapshot {
   deno_core::extension!(
     runtime_main,
     deps = [runtime],
-    customizer = |ext: &mut deno_core::ExtensionBuilder| {
-      ext.esm_entry_point("ext:runtime/90_deno_ns.js");
+    customizer = |ext: &mut deno_core::Extension| {
+      ext.esm_entry_point = Some("ext:runtime/90_deno_ns.js");
     }
   );
```
</details>

nayeemrmn:
Edited a corruption in the above diff 🙏

mmastrac (approved):
LGTM!

mmastrac:
@nayeemrmn Merged this one, thanks -- let's open a tracking PR in deno with your patch (pointing the `deno_core` crate in Cargo.toml at the `main` branch of deno_core) while we wait for the next release of deno_core.

PR #80, opened 2023-07-23
perf(primordials): Check the need for `SafeArrayIterator` each time

### LINKED ISSUE #6: I'll try to use the fast path instead of `SafeIterator` from primordials
> I'm reasonably certain it was intentional, and I believe engines are capable of optimizing it when `Array.prototype[Symbol.iterator]` is unmodified.

https://github.com/tc39/ecma262/issues/827#issuecomment-282540081

petamoriken:
Closes #6


bartlomieju:
@petamoriken is there a test we could add here?

petamoriken:
@bartlomieju I have no idea what tests to add for primordials.

bartlomieju:
Does this PR introduce a change in behavior?

petamoriken:
If `Array.prototype[Symbol.iterator]` and `%ArrayIteratorPrototype%.next` are not polluted, `new SafeIterator(arr)` just returns `arr` without wrapping anything.

I believe this will improve performance, but don't know how to check performance.

bartlomieju:
> If `Array.prototype[Symbol.iterator]` and `%ArrayIteratorPrototype%.next` are not polluted, `new SafeArrayIterator(arr)` just returns `arr` without wrapping anything.
> 
> I believe this will improve performance, but don't know how to check performance.

You could create a benchmark that uses one object with these props and one without an then see the difference between execution times for say 100_000 items in the iterator.

You can see exmaple benchmarks in `core/benches/ops/sync.rs`

mmastrac:
@petamoriken Would you still like to land this?

petamoriken:
@mmastrac I am still inclined to work on it myself, but I don't mind having someone else do the work.

PR #81, opened 2023-07-23
feat(realms): Support ESM extensions in realms

andreubotella:
This PR adds support for ESM snapshots and `preserve_snapshotted_modules` to realms, as well as fixing a few bugs where `JsRuntime::init_extension_js` was using `JsRuntime` methods instead of `JsRealm` methods in ESM extension-related code.

It also adds realm-specific variants of a few ESM extensions tests.


core/runtime/tests/jsrealm.rs
```diff
@@ -803,3 +803,392 @@ async fn test_realm_concurrent_dynamic_imports() {
...
+
+  let mut runtime = JsRuntime::new(RuntimeOptions {
+    module_loader: None,
+    startup_snapshot: Some(Snapshot::JustCreated(startup_data)),
+    ..Default::default()
+  });
+
+  struct Loader;
+
+  impl ModuleLoader for Loader {
```
mmastrac, line 838: Let's wait for #84 to land so we don't add more ModuleLoaders.

core/runtime/tests/jsrealm.rs
```diff
@@ -803,3 +803,392 @@ async fn test_realm_concurrent_dynamic_imports() {
...
+        .as_str(),
+      "foo"
+    );
+  }
+}
+
+/// Test that `RuntimeOptions::preserve_snapshotted_modules` also works without
+/// a snapshot.
+#[test]
+fn non_snapshot_preserve_snapshotted_modules() {
```
mmastrac, line 1098: For tests that are identical other than a small change, let's try to avoid copy+paste. Given that this particular test is already kind of combinatorial, maybe we can have a base test that takes two parameters: snapshot: bool and realm: bool.

bartlomieju (approved):
LGTM, thank you Andreu.

mmastrac:
LGTM -- we can resolve the conflict either by re-applying or using `#[allow(deprecated)]` for now. We should allow for direct source to be allowed in the `extension!` macro to avoid having to use the builder for this case.

andreubotella:
> LGTM -- we can resolve the conflict either by re-applying or using `#[allow(deprecated)]` for now. We should allow for direct source to be allowed in the `extension!` macro to avoid having to use the builder for this case.

I was already fixing the conflicts and changing any other new tests to use the `Extension` struct rather than the builder.

mmastrac:
Thanks! I filed #90 to allow us to use the `extension!` macro more easily for these tests.

mmastrac (approved):

PR #82, opened 2023-07-24
fix(realms): Call the right realm's format exception callback for promise rejections

andreubotella:
The `JsRuntime::check_promise_rejection` method takes care of failing the event loop run if it detects any rejected promise in any realm. It does so by delegating to `JsRealmInner::check_promise_rejection` for every realm; however, the scope it passes to that method is `runtime.handle_scope()` – that is, the main realm's scope. The `JsRealmInner` method in turn uses this scope, which results in the main realm's format exception callback being called, rather than the actual realm's.

This patch fixes this by changing `JsRealmInner::check_promise_rejection` to take a `&mut v8::Isolate` rather than a `&mut v8::HandleScope`, and to call `Self::handle_scope` to create a scope in the realm, as the rest of `JsRealm` and `JsRealmInner` methods do. It also adds realm-specific tests for the format exception callback.

### DIFF
--- a/core/runtime/jsrealm.rs
+++ b/core/runtime/jsrealm.rs
@@ -208,12 +208,13 @@ impl JsRealmInner {
 
   pub(crate) fn check_promise_rejections(
     &self,
-    scope: &mut v8::HandleScope,
+    isolate: &mut v8::Isolate,
   ) -> Result<(), Error> {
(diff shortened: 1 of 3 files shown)


bartlomieju (approved):
LGTM, thanks!

PR #83, opened 2023-07-24
fix(core): Re-run event loop after resolving ops

mmastrac:
Repro and fix for https://github.com/denoland/deno/issues/19903 and https://github.com/denoland/deno/issues/19455.

No perf impact seen in deno before and after.


bartlomieju:
We explicitly removed this wake up some time ago for performance reasons. Doesn't this change cause any performance regression?

bartlomieju (approved):
LGTM, but needs rebase

PR #84, opened 2023-07-25
refactor(core): Introduce a StaticModuleLoader

mmastrac:
Adds the following test utilities to remove a lot of boilerplate from the deno_core tests:

 - `StaticModuleLoader` to provide a set of static files that don't require the filesystem
 - `CountingModuleLoader` to test resolve/prepare/load counts
 - `LoggingModuleLoader` to log a list of load attempts

### DIFF
--- a/core/modules/loaders.rs
+++ b/core/modules/loaders.rs
@@ -10,8 +10,10 @@ use crate::modules::ModuleType;
 use crate::modules::ResolutionKind;
 use crate::resolve_import;
 use crate::Extension;
+
 use anyhow::anyhow;
 use anyhow::Error;
+use futures::future::ready;
 use futures::future::FutureExt;
 use std::cell::RefCell;
 use std::collections::HashMap;
@@ -26,10 +28,10 @@ pub trait ModuleLoader {
   /// algorithm described here:
   /// <https://html.spec.whatwg.org/multipage/webappapis.html#resolve-a-module-specifier>
   ///
-  /// `is_main` can be used to resolve from current working directory or
+  /// [`ResolutionKind::MainModule`] can be used to resolve from current working directory or
   /// apply import map for child imports.
   ///
-  /// `is_dyn_import` can be used to check permissions or deny
+  /// [`ResolutionKind::DynamicImport`] can be used to check permissions or deny
   /// dynamic imports altogether.
   fn resolve(
     &self,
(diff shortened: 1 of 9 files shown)


core/modules/loaders.rs
```diff
@@ -26,17 +29,19 @@ pub trait ModuleLoader {
...
   fn resolve(
     &self,
     specifier: &str,
     referrer: &str,
-    kind: ResolutionKind,
-  ) -> Result<ModuleSpecifier, Error>;
+    _kind: ResolutionKind,
+  ) -> Result<ModuleSpecifier, Error> {
+    Ok(resolve_import(specifier, referrer)?)
+  }
```
bartlomieju, line 44: I'm not a fan of this change - it's up to implementer to decide on the resolve implementation. Adding a default implementation might be a footgun.

core/modules/loaders.rs
```diff
@@ -251,3 +227,110 @@ impl ModuleLoader for FsModuleLoader {
...
+
+pub struct CountingModuleLoader<L: ModuleLoader> {
+  loader: L,
+  load_count: Cell<usize>,
+  prepare_count: Cell<usize>,
+  resolve_count: Cell<usize>,
+}
+
+#[derive(Copy, Clone, Default, Debug, Eq, PartialEq)]
+pub struct ModuleLoadEventCounts {
```
bartlomieju, line 285: Add `#[test]`?

core/modules/loaders.rs
mmastrac: I can revert this part -- many of the implementations outside of deno are just calling `resolve_import`, however.

I suspect it might be better for us to split resolution and loading at some point in the future/

core/modules/loaders.rs
```diff
@@ -78,9 +80,7 @@ impl ModuleLoader for NoopModuleLoader {
     referrer: &str,
     _kind: ResolutionKind,
   ) -> Result<ModuleSpecifier, Error> {
-    Err(generic_error(
```
mmastrac, line 81: Note that I'm intentionally changing `NoopModuleLoader` to allow for resolution but not loading -- this allows for JsRuntimes with no loader to correctly resolve pre-loaded modules which I think makes more sense (see `es_snapshot_without_runtime_module_loader`).

core/modules/loaders.rs
bartlomieju: Maybe, let's keep it reverted for now please

core/modules/loaders.rs
bartlomieju: Seems fine 👍

bartlomieju (approved):
LGTM, nice clean up!

PR #85, opened 2023-07-25
fix(core): Avoid cloning Vec<JsRealmInner> when we only have one realm

mmastrac:
Cloning the `known_realms` Vec in the inner loop adds a pretty significant overhead. We will need to get rid of this at some point but for now we can optimize the one-realm case.


bartlomieju (approved):
LGTM!

PR #91, opened 2023-07-26
fix(extensions): use expr instead of literal

nayeemrmn:
You can see this done for things like https://doc.rust-lang.org/std/macro.concat.html.

Lets users pass computed expressions as long as they are computed at compile time.

Leads to better error messages if a non-literal is passed:
```rust
extension!(foo, esm_entry_point = "".to_string(),);
```
Before:
```
error: no rules expected the token `.`
   --> core/extensions.rs:783:37
    |
241 | macro_rules! extension {
    | ---------------------- when calling this macro
...
783 | extension!(foo, esm_entry_point = "".to_string(),);
    |                                     ^ no rules expected this token in macro call
    |
    = note: while trying to match sequence start
```
After:
```
error[E0308]: mismatched types
   --> core/extensions.rs:783:35
    |
288 |             const V: Option<&'static str> = $crate::or!($(Some($esm_entry_point))?, None);
    |                                                           ---- arguments to this enum variant are incorrect
...
783 | extension!(foo, esm_entry_point = "".to_string(),);
    |                                   ^^^^^^^^^^^^^^ expected `&str`, found `String`
```

### DIFF
--- a/core/extensions.rs
+++ b/core/extensions.rs
@@ -246,9 +246,9 @@ macro_rules! extension {
     $(, bounds = [ $( $bound:path : $bound_type:ident ),+ ] )?
     $(, ops_fn = $ops_symbol:ident $( < $ops_param:ident > )? )?
     $(, ops = [ $( $(#[$m:meta])* $( $op:ident )::+ $( < $( $op_param:ident ),* > )?  ),+ $(,)? ] )?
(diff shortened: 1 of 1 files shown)


mmastrac (approved):
LGTM!

PR #94, opened 2023-07-27
fix(extensions): use literal for js source in macro

nayeemrmn:
Mistake I made in #91. `$js` has to stay a literal to avoid parsing ambiguities, just like `$esm:literal`. cc @mmastrac


mmastrac (approved):
Looks good

PR #99, opened 2023-07-27
fix(realms): Make realm creation not hang with async ops running

andreubotella:
The way `JsRuntime::init_extension_js` initialized ESM extensions without a snapshot was by running the event loop and blocking the thread ntil there were no more tasks. This works fine for the main realm, because it is created before any user code can start any async ops. However, other realms may be created while async ops are running, which would make realm creation hang until those ops are resolved.

Since extensions are not expected to use top-level await, there would be no need to run the entire event loop for modules to resolve; calling `JsRealm::evaluate_pending_module` would be enough. If the module is not resolved after calling that functions, then it or one of its dependencies has TLA, in which case we panic.

Since a panic like "TLA is not allowed" would be hard to debug, especially on codebases with many modules, we use `ModuleMap::find_stalled_top_level_await` to find at least one of the TLA locations, and we display it in the panic message.

### DIFF
--- a/core/runtime/jsruntime.rs
+++ b/core/runtime/jsruntime.rs
@@ -967,11 +967,36 @@ impl JsRuntime {
               panic!("{} not present in the module map", specifier)
             })
         };
-        let receiver = realm.mod_evaluate(self.v8_isolate(), mod_id);
-        self.run_event_loop(false).await?;
-        receiver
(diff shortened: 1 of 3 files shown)


bartlomieju (approved):
Lovely solution, LGTM!

PR #100, opened 2023-07-27
fix: Remove unnecessary clones in `v8::Local` constructors

andreubotella:
The second argument to `v8::Local` can be a reference, so cloning isn't needed.


bartlomieju (approved):
LGTM, very nice!

PR #102, opened 2023-07-29
feat(core): Add resource_handle (prereq for fast streams work)

mmastrac:
This simplifies much of the downstream code by removing the need for `cfg(unix)`/`cfg(windows)` in the intermediate stream/resource interfaces, and allows for fast stream optimizations on raw sockets.

Carved off from the larger streams PR.


bartlomieju (approved):
LGTM

PR #110, opened 2023-08-01
feat(ops): Add async(lazy) to avoid eager polling

mmastrac:
`#[op2(async(lazy))]` prevents eager polling and will eventually allow for fast async ops. 

This is distinctly different than `deferred` which polls the op immediately and then defers that result.


ops/op2/config.rs
```diff
@@ -0,0 +1,194 @@
...
+
+use crate::op2::Op2Error;
+
+#[derive(Debug, Default, Eq, PartialEq)]
+pub(crate) struct MacroConfig {
+  /// Use `(core)` for ops that live in `deno_core`.
+  pub core: bool,
+  /// Generate a fastcall method (must be fastcall compatible).
+  pub fast: bool,
+  /// Do not generate a fastcall method (must be fastcall compatible).
```
bartlomieju, line 14:
```
(must be fastcall compatible).
```

seems wrong here

mmastrac: That's actually correct -- `nofast` is only for potentially fast methods.

bartlomieju (approved):
LGTM

PR #111, opened 2023-08-01
feat(core): add a write_error method to Resource

mmastrac:
This method will be used to simplify the handling of resource streams that are backed by `tokio::sync::mpsc::channel`. Rather than having to juggle a shutdown/drop handler, the JS code can send an error directly through to the resource, which will get written to the pipe.


bartlomieju (approved):
LGTM

lucacasonato:
I'm confused by this - what resources should implement it. There should not be a stream type specific method on the resource, because to JS, all writable rids should behave identically.

PR #112, opened 2023-08-01
fix: run js event loop tick before polling dynamic imports

bartlomieju:
This fixes regressions reported in https://github.com/denoland/deno/issues/19903.

We mistakenly were not waking up the event loop again if ops made progress
and there are pending dynamic imports waiting for TLA to resolve.


mmastrac (approved):
LGTM

PR #117, opened 2023-08-03
feat(extensions): computed extension code

### LINKED ISSUE #116: `ExtensionFileSourceCode` variant for computed strings
As part of the approach I'm favouring for https://github.com/denoland/deno/issues/19399, we need another variant for computed strings. Since it would need to be cloned on `.load()` it should be stored as `Arc<str>`.
```rust
pub enum ExtensionFileSourceCode {
  IncludedInBinary(&'static str),
  LoadedFromFsDuringSnapshot(&'static str), // <- Path
  Computed(Arc<str>),
}
```
And then I think we can remove `snapshot_module_load_cb` in favour of just computing the TS transpilations, which we would still only do during snapshot creation (or a no-snapshot mode for dev in the future).

---

Using a `module_load_cb` (expanding that feature to work outside of snapshot creation) has the advantage of holding one less copy of modules throughout the process runtime including the one allocated by V8. But it also means it needs to be recomputed for other realms, so that allocation is meaningful. For CLI we don't care very much since again it's only done on snapshot creation.

Another solution is `ExtensionFileSourceCode::IncludedInBinary(computed_string.leak())` which will be stable later this month. That can take advantage of static ascii opts. But leaking may not always be an option for embedders, there's no harm in having this option.

nayeemrmn:
Closes #116.


core/extensions.rs
```diff
@@ -6,11 +6,12 @@ use anyhow::Error;
...
 use std::cell::RefCell;
 use std::rc::Rc;
+use std::sync::Arc;
 use std::task::Context;
 use v8::fast_api::FastFunction;
 use v8::ExternalReference;
 
-#[derive(Copy, Clone, Debug)]
+#[derive(Clone, Debug)]
 pub enum ExtensionFileSourceCode {
```
nayeemrmn, line 15: I made this `Copy` in #78 just because I could after removing `PathBuf`. It's not important.

nayeemrmn:
CLI patch showing this replacing `snapshot_module_load_cb`:
<details>
<summary></summary>

```diff
diff --git a/runtime/build.rs b/runtime/build.rs
index 828bc3c53..5a2a6eb89 100644
--- a/runtime/build.rs
+++ b/runtime/build.rs
@@ -17,39 +17,33 @@ mod startup_snapshot {
   use deno_core::snapshot_util::*;
   use deno_core::Extension;
   use deno_core::ExtensionFileSource;
-  use deno_core::ModuleCode;
+  use deno_core::ExtensionFileSourceCode;
   use deno_http::DefaultHttpPropertyExtractor;
   use std::path::Path;
 
-  fn transpile_ts_for_snapshotting(
-    file_source: &ExtensionFileSource,
-  ) -> Result<ModuleCode, AnyError> {
+  fn maybe_transpile_source(
+    source: &mut ExtensionFileSource,
+  ) -> Result<(), AnyError> {
     // Always transpile `node:` built-in modules, since they might be TypeScript.
-    let media_type = if file_source.specifier.starts_with("node:") {
+    let media_type = if source.specifier.starts_with("node:") {
       MediaType::TypeScript
     } else {
-      MediaType::from_path(Path::new(&file_source.specifier))
+      MediaType::from_path(Path::new(&source.specifier))
     };
 
-    let should_transpile = match media_type {
-      MediaType::JavaScript => false,
-      MediaType::Mjs => false,
-      MediaType::TypeScript => true,
-      _ => {
-        panic!(
-          "Unsupported media type for snapshotting {media_type:?} for file {}",
-          file_source.specifier
-        )
-      }
-    };
-    let code = file_source.load()?;
-
-    if !should_transpile {
-      return Ok(code);
+    match media_type {
+      MediaType::TypeScript => {}
+      MediaType::JavaScript => return Ok(()),
+      MediaType::Mjs => return Ok(()),
+      _ => panic!(
+        "Unsupported media type for snapshotting {media_type:?} for file {}",
+        source.specifier
+      ),
     }
+    let code = source.load()?;
 
     let parsed = deno_ast::parse_module(ParseParams {
-      specifier: file_source.specifier.to_string(),
+      specifier: source.specifier.to_string(),
       text_info: SourceTextInfo::from_string(code.as_str().to_owned()),
       media_type,
       capture_tokens: false,
@@ -62,7 +56,9 @@ mod startup_snapshot {
       ..Default::default()
     })?;
 
-    Ok(transpiled_source.text.into())
+    source.code =
+      ExtensionFileSourceCode::Computed(transpiled_source.text.into());
+    Ok(())
   }
 
   #[derive(Clone)]
@@ -304,7 +300,7 @@ mod startup_snapshot {
     // NOTE(bartlomieju): ordering is important here, keep it in sync with
     // `runtime/worker.rs`, `runtime/web_worker.rs` and `cli/build.rs`!
     let fs = std::sync::Arc::new(deno_fs::RealFs);
-    let extensions: Vec<Extension> = vec![
+    let mut extensions: Vec<Extension> = vec![
       deno_webidl::deno_webidl::init_ops_and_esm(),
       deno_console::deno_console::init_ops_and_esm(),
       deno_url::deno_url::init_ops_and_esm(),
@@ -348,13 +344,22 @@ mod startup_snapshot {
       runtime_main::init_ops_and_esm(),
     ];
 
+    for extension in &mut extensions {
+      for source in extension.esm_files.to_mut() {
+        maybe_transpile_source(source).unwrap();
+      }
+      for source in extension.js_files.to_mut() {
+        maybe_transpile_source(source).unwrap();
+      }
+    }
+
     let output = create_snapshot(CreateSnapshotOptions {
       cargo_manifest_dir: env!("CARGO_MANIFEST_DIR"),
       snapshot_path,
       startup_snapshot: None,
       extensions,
       compression_cb: None,
-      snapshot_module_load_cb: Some(Box::new(transpile_ts_for_snapshotting)),
+      snapshot_module_load_cb: None,
       with_runtime_cb: None,
     });
     for path in output.files_loaded_during_snapshot {
```
</details>

Should I remove or deprecate it?

bartlomieju (commented):
@nayeemrmn overall looks good and I like the cleanup. I'm a bit worried that using `Arc<str>` will cause us to keep some source code in memory after handing it off to V8 resulting in bigger binary sizes. Could you verify that it's not the case in Deno before we land it?

nayeemrmn:
Looks good - 100,100 kb -> 100,080 kb

bartlomieju (approved):
LGTM!

PR #121, opened 2023-08-05
ci: fix "deno fmt --check"

bartlomieju:


core/ops_builtin.rs
```diff
@@ -287,6 +287,10 @@ async fn op_read_all(
   Ok(vec)
 }
 
+//
+//
+//
+
```
mmastrac, line 293: What is this for?

mmastrac (approved):
LGTM

core/ops_builtin.rs
bartlomieju: I was debugging something else and it's a leftover.

bartlomieju:
mac 13m -> 10m
linux 13m -> 11m
Windows 13m -> 12m

PR #123, opened 2023-08-06
ci: remove build step

bartlomieju:


bartlomieju:
@mmastrac I'm not sure why we needed both `cargo build` and `cargo test`. Seems like the latter is enough.

mmastrac (approved):
LGTM

PR #136, opened 2023-08-06
fix: use call sites when checking for empty frames

nayeemrmn:
Will fix ... https://github.com/denoland/deno/issues/20034.

As we know, `msg.get_stack_frames()` does not get populated properly for async op errors. That's why we use the 'call sites' passed to `prepareStackTrace()` as a source of truth instead. I had forgotten to account for that in one odd place.

### DIFF
--- a/core/error.rs
+++ b/core/error.rs
@@ -624,6 +624,30 @@ pub(crate) fn is_aggregate_error(
   false
 }
 
+/// Check if the error has a proper stack trace. The stack trace checked is the
+/// one passed to `prepareStackTrace()`, not `msg.get_stack_trace()`.
+pub(crate) fn has_call_site(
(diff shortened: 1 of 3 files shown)


bartlomieju:
@nayeemrmn could you check if this change will help with https://github.com/denoland/deno/issues/18946?

nayeemrmn:
> could you check if this change will help with [denoland/deno#18946](https://github.com/denoland/deno/issues/18946)?

It doesn't, this only has to do with cases where the original error stack is replaced with the location of a dynamic import it passed through because it was wrongly detected as empty.

bartlomieju:
Any chance we could get a regression test for this?

bartlomieju (approved):
LGTM, thanks for a quick fix!

PR #137, opened 2023-08-08
fix: don't swallow rejections in poll_value()

nayeemrmn:
Unblocks https://github.com/denoland/deno/pull/20082.

Fixes three tangled issues:
- `JsRuntime::poll_value(promise)` swallows non-awaited rejections returned from `poll_event_loop()` when `promise` is already settled.
- `JsRuntime::poll_value(promise)` runs the event loop once even if `promise` is already settled.
- `JsRuntime::call_and_await()` needs to explicitly call `check_promise_rejections()` at the end.

Some test expectations might need to add `(in promise)` to error strings. These are correct. There's at least one in CLI repo:
```diff
diff --git a/cli/tests/testdata/node/test.out b/cli/tests/testdata/node/test.out
index 8b7f0780f..3c54a15e8 100644
--- a/cli/tests/testdata/node/test.out
+++ b/cli/tests/testdata/node/test.out
@@ -147,7 +147,7 @@ error: Error: rejected from reject fail
     at [WILDCARD]
 
 ./node/test.js (uncaught error)
-error: Error: rejected from unhandled rejection fail
+error: (in promise) Error: rejected from unhandled rejection fail
   Promise.reject(new Error("rejected from unhandled rejection fail"));
                  ^
     at [WILDCARD]
```

### DIFF
--- a/core/runtime/jsruntime.rs
+++ b/core/runtime/jsruntime.rs
@@ -1220,7 +1220,9 @@ impl JsRuntime {
       }
       v8::Global::new(scope, promise.unwrap())
     };
-    self.resolve_value(promise).await
+    let result = self.resolve_value(promise).await;
+    self.check_promise_rejections()?;
+    result
   }
 
   /// Returns the namespace object of a module.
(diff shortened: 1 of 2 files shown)


bartlomieju:
@nayeemrmn what do you mean by
> After fixing these, JsRuntime::call_and_await() doesn't account for non-awaited rejections.

How that should be handled by the embedder?

nayeemrmn:
> @nayeemrmn what do you mean by
> 
> > After fixing these, JsRuntime::call_and_await() doesn't account for non-awaited rejections.
> 
> How that should be handled by the embedder?

I wrote it confusingly. It was a third issue introduced by fixing the first two but it's also fixed.

bartlomieju (approved):
LGTM!

PR #138, opened 2023-08-09
refactor: move get_error_class_fn to OpCtx

### LINKED ISSUE #51: Migrate `get_error_class_fn` to `OpCtx`
The code paths that retrieve `get_error_class_fn` are currently `borrow`ing the `OpState` from a `RefCell` first. Let's move these to `OpCtx` directly (one copy of this function pointer per `OpCtx` so that we can simply read it as we need it.

This may impact both op and op2 code. If major changes are required to `op` code, let's leave the old copy of `get_error_class_fn` on OpState and get rid of it when we fully deprecate op1

nayeemrmn:
Closes #51. `cargo test error` works in main repo.

### DIFF
--- a/core/ops.rs
+++ b/core/ops.rs
@@ -129,6 +129,7 @@ pub struct OpCtx {
   /// A stashed Isolate that ops can use in callbacks
   pub isolate: *mut Isolate,
   pub state: Rc<RefCell<OpState>>,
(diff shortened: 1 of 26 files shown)


bartlomieju (approved):
LGTM, so much cleaner

PR #142, opened 2023-08-09
feat(core): Add split_* methods to BufView and V8Slice

mmastrac:
With this PR, we no longer allow `BufMutView` or `BufView` to be backed by `Vec` (except in the case of a `bytes::Bytes`-backed `BufView` which could handle it internally). This allows us to add the `split_off` and `split_to` methods to `V8Slice` and our `Buf*View` buffers, which means that we can take subsets of existing V8 or Rust buffers.

Note that our mutable buffers are not cheaply cloneable, but we do allow for cheap splitting which allows for a buffer to be segmented into a cloneable read-only segment and a non-cloneable read-write segment.

This is required for us to handle channels that receive buffers from JavaScript, as internal buffering limits may require us to split those V8-backed arraybuffers to fit within other internal buffers (ie: calling read(1024) with a source that is providing 2048-byte buffers).

`op_read_all` needed to be refactored.

Performance impact: none expected

Future work unlocked: we can now take advantage of `MaybeUninit` down the line, as well as making our BufViews cheaply cloneable.

Patch needed for `deno`:

```
--- a/ext/io/fs.rs
+++ b/ext/io/fs.rs
@@ -170,14 +170,10 @@ impl FsStat {
 pub trait File {
   fn read_sync(self: Rc<Self>, buf: &mut [u8]) -> FsResult<usize>;
   async fn read(self: Rc<Self>, limit: usize) -> FsResult<BufView> {
-    let vec = vec![0; limit];
-    let buf = BufMutView::from(vec);
-    let (nread, buf) = self.read_byob(buf).await?;
-    let mut vec = buf.unwrap_vec();
-    if vec.len() != nread {
-      vec.truncate(nread);
-    }
-    Ok(BufView::from(vec))
+    let buf = BufMutView::new(limit);
+    let (nread, mut buf) = self.read_byob(buf).await?;
+    buf.truncate(nread);
+    Ok(buf.into_view())
   }
```

### DIFF
--- a/core/buffer_strategy.rs
+++ b/core/buffer_strategy.rs
@@ -1,3 +1,5 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+
 /// Our standard buffer size if we don't know what else to do.
 const STANDARD_BUFFER_SIZE: usize = 64 * 1024;
 /// Our buffer size if the resource is tiny or likely empty.
--- a/core/buffer_strategy.rs
+++ b/core/buffer_strategy.rs
@@ -1,0 +1,251 @@
+/// Our standard buffer size if we don't know what else to do.
+const STANDARD_BUFFER_SIZE: usize = 64 * 1024;
+/// Our buffer size if the resource is tiny or likely empty.
+const TINY_BUFFER_SIZE: usize = 1024;
+const MAX_GROW_LEN: usize = 16 * 1024 * 1024;
+
+/// Our size classes.
+enum SizeClass {
+  // Zero min, no max (ie: no hint given)
+  Unknown,
+  // Likely empty (min == max == 0)
+  LikelyEmpty,
+  // Likely known (min == max)
+  LikelyKnown(usize),
+  // Bounded, but not known
+  Between(usize, usize),
+  // Bounded maximum, but not known
+  BoundedMax(usize),
+  // Bounded minimum, but not known
+  BoundedMin(usize),
+}
+
+impl SizeClass {
+  fn from_size_hint(min: usize, maybe_max: Option<usize>) -> Self {
+    match (min, maybe_max) {
+      (0, None) => Self::Unknown,
+      (min, None) => Self::BoundedMin(min),
+
+      (0, Some(0)) => Self::LikelyEmpty,
+      (0, Some(max)) => Self::BoundedMax(max),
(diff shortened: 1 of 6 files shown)


core/buffer_strategy.rs
```diff
@@ -0,0 +1,251 @@
+/// Our standard buffer size if we don't know what else to do.
```
bartlomieju, line 1: Missing copyright - we should probably add copyright checker to this repo.

core/buffer_strategy.rs
```diff
@@ -0,0 +1,251 @@
...
+const STANDARD_BUFFER_SIZE: usize = 64 * 1024;
+/// Our buffer size if the resource is tiny or likely empty.
+const TINY_BUFFER_SIZE: usize = 1024;
+const MAX_GROW_LEN: usize = 16 * 1024 * 1024;
```
bartlomieju, line 5: Was this selected arbitrarly?

core/io.rs
```diff
@@ -64,6 +64,53 @@ impl BufView {
...
+      inner: other,
+      cursor: 0,
+    }
+  }
+
+  /// Split the underlying buffer. The other piece will have a cursor of zero while this buffer
+  /// will maintain the current cursor position.
+  pub fn split_to(&mut self, at: usize) -> Self {
+    assert!(at <= self.len());
+    let at = at + self.cursor;
```
bartlomieju, line 100: Shouldn't we assert that this is also <= self.len?

bartlomieju (approved):
Couple nits. Does this PR brings us any closer to premises of https://github.com/denoland/deno/pull/18452?

core/io.rs
mmastrac: In both the bytes and v8 slice cases, there's a matching assert in the underlying V8Slice/Bytes, so we'll catch those there.

core/buffer_strategy.rs
bartlomieju: That's concerning to hear. I'm looking forward to seeing what you propose

mmastrac:
@bartlomieju 

> Couple nits. Does this PR brings us any closer to premises of https://github.com/denoland/deno/pull/18452?

Not in a massive meaningful way but we can definitely address it after. It doesn't make things less safe, nor does it make it harder if we decide to go the safe route.

PR #143, opened 2023-08-10
fix(serde_v8): rawbytes was missing Drop and Into implementations

mmastrac:


bartlomieju (approved):
Can this be tested somehow?

PR #155, opened 2023-08-23
fix(serde_v8): v8slice split methods were incorrect after multiple splits

mmastrac:
This was causing the first part of large chunks in a ReadableStream to get repeated. Instead of responding with chunks ABCDEF of a large HTTP response, we'd respond with AAAAAA.

```
12:45 $ curl -D- localhost:8000 | grep "<html"
<html lang="en">
<html lang="en">
<html lang="en">
<html lang="en">
<html lang="en">
<html lang="en">
<html lang="en">
<html lang="en">
<html lang="en">
<html lang="en">
```

Fixes https://github.com/denoland/fresh/issues/1699

### DIFF
--- a/serde_v8/magic/v8slice.rs
+++ b/serde_v8/magic/v8slice.rs
@@ -40,6 +40,27 @@ impl Debug for V8Slice {
 unsafe impl Send for V8Slice {}
 
 impl V8Slice {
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM

PR #158, opened 2023-08-24
feat(core): add is_terminal to ResourceHandle

mmastrac:
This will allow us to remove some `cfg` structures from `deno`.


bartlomieju (approved):
LGTM

PR #159, opened 2023-08-25
upgrade v8 to disable pointer compression, fix memleak

ry:


core/runtime/jsruntime.rs
```diff
@@ -326,7 +326,7 @@ fn v8_init(
   #[repr(C, align(16))]
   struct IcuData([u8; 10541264]);
   static ICU_DATA: IcuData = IcuData(*include_bytes!("icudtl.dat"));
-  v8::icu::set_common_data_72(&ICU_DATA.0).unwrap();
+  v8::icu::set_common_data_73(&ICU_DATA.0).unwrap();
```
mmastrac, line 329: I had to downgrade the ICU data as part of https://github.com/denoland/deno_core/pull/153 -- do we need to upgrade it again?

ry: Yes, this is back to v8 11.7.439.6 which depends on ICU 73

mmastrac: Sorry, I meant the `icudtl.dat` file itself -- the version pre-downgrade was a different size (10631872). I'm not sure if we need to include that file in this update.

If we need to update this file, it should be available in this commit here: https://github.com/mmastrac/deno_core/blob/fe80221e774fdfde8640571a4b80446fa542d6bb/core/runtime/icudtl.dat

bartlomieju: Yes, we should include the other data file.

ry: fixed and added a test

mmastrac (approved):
LGTM! Great test.

PR #162, opened 2023-08-28
feat(ops): support #[serde] return from async ops

mmastrac:
This was unimplemented for async function return values


bartlomieju (approved):
LGTM, thanks!

PR #163, opened 2023-08-28
fix(ops): support where T: Trait + 'static in generics

### LINKED ISSUE #140: `op2` should allow `where` clause with `'static`
`op2` supports `where T: Trait` , but it does not support `where T: 'static`. This should be considered valid as there may be cases where the op requires a `'static` trait but the trait does not need to be `'static`.

mmastrac:
Fixes #140


bartlomieju (approved):
LGTM

PR #164, opened 2023-08-28
feat(ops): support most return types from async functions

### LINKED ISSUE #139: `op2` async numeric return values
The async return values for `op2` are sparse. We need to ensure that all numeric return values are supported.

As part of this work, `valid_retvals.md` needs to be updated with a new column for `async` support.

Example op that we cannot port:

```
#[op2(async)]
#[bigint]
pub async fn op_cache_storage_open<CA>(
  state: Rc<RefCell<OpState>>,
  #[string] cache_name: String,
) -> Result<i64, AnyError>
```

mmastrac:
Fixes #139 among other things


bartlomieju (approved):
LGTM

PR #167, opened 2023-08-30
fix: repository URL in Cargo.toml

bajtos:
Change the repository from `denoland/deno` to `denoland/deno_core` to fix the
information shown by https://crates.io/crates/deno_core

BTW is there a reason why Deno crates don't publish their API docs to https://docs.rs/?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ members = [
 authors = ["the Deno authors"]
 edition = "2021"
 license = "MIT"
-repository = "https://github.com/denoland/deno"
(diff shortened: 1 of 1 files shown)


bajtos:
> BTW is there a reason why Deno crates don't publish their API docs to https://docs.rs/?

Nevermind. I see the docs are available at https://docs.rs/deno_core/0.205.0/deno_core/, it's just Crates.io that does not link there 🤷🏻

bartlomieju (approved):
Thanks!

PR #168, opened 2023-08-30
fix(ops): Handle #[smi] types

mmastrac:
`#[smi]`-tagged arguments were going through the wrong path because I neglected to restore the custom `Numeric(__SMI__)` handling in  #165.

To fix this, we use a `RustToV8Marker` to override the default behaviour and ensure that SMIs are always serialized as v8 i32s. Fixes a Deno crash in the HTTP test.

### DIFF
--- a/core/runtime/ops.rs
+++ b/core/runtime/ops.rs
@@ -835,6 +835,8 @@ mod tests {
     Ok(())
   }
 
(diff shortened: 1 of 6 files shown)


bartlomieju (approved):
LGTM

PR #171, opened 2023-08-31
feat: support import attributes

bartlomieju:
https://tc39.es/proposal-import-attributes/

Pretty much the same as import assertions, which are getting deprecated.


dsherret (approved):
LGTM

PR #181, opened 2023-09-08
fix(core): Fix miri test problems to remove UB in CancelHandle

mmastrac:
This now passes cleanly:

```
cargo +nightly miri test -- async_cancel
cargo +nightly miri test -- async_cell
```

### DIFF
--- a/core/async_cancel.rs
+++ b/core/async_cancel.rs
@@ -399,13 +399,9 @@ mod internal {
         Some((head, rc)) => {
           // Register this `Cancelable` node with a `CancelHandle` head node.
           assert_ne!(self, head);
-          // TODO(piscisaureus): safety comment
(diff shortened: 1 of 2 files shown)


core/async_cancel.rs
```diff
@@ -779,6 +775,7 @@ mod tests {
     assert!(!cancel_handles.iter().any(|c| !c.is_canceled()));
   }
 
+  #[cfg(not(miri))]
```
bartlomieju, line 778: Looks like bad formatting here

bartlomieju (approved):

core/async_cancel.rs
mmastrac: Sorry, I am not sure what you meant here. What are you seeing?

PR #183, opened 2023-09-08
chore(core): deprecate ResourceTable::close

mmastrac:
Follow-up to #179 -- we don't want to use this method.

### DIFF
--- a/core/resources.rs
+++ b/core/resources.rs
@@ -458,7 +458,7 @@ impl ResourceTable {
   /// counted, therefore pending ops are not automatically cancelled. A resource
   /// may implement the `close()` method to perform clean-ups such as canceling
   /// ops.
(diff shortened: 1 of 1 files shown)


core/resources.rs
```diff
@@ -458,6 +458,7 @@ impl ResourceTable {
   /// counted, therefore pending ops are not automatically cancelled. A resource
   /// may implement the `close()` method to perform clean-ups such as canceling
   /// ops.
+  #[deprecated="This method may deadlock. Use take() and close() instead."]
```
mmastrac, line 461:
```suggestion
  #[deprecated = "This method may deadlock. Use take() and close() instead."]
```

bartlomieju (approved):

PR #200, opened 2023-09-14
feat: add explicit JsRuntime::init_platform method

lucacasonato:
Add note about thread inheritance requierments of V8 platform.

There is a follow up to be done to ensure our tests follow the new thread
inheritance requirements.


bartlomieju (approved):
LGTM

PR #202, opened 2023-09-14
feat(ops): add #[number] for 64-bit value to avoid bigints

mmastrac:
Assuming the value fits in Number.MIN_SAFE_INTEGER..Number.MAX_SAFE_INTEGER, you can safely return large values as floats.


bartlomieju (approved):
Nice

PR #214, opened 2023-09-19
feat(op2): support JsRuntimeState as op parameter

lucacasonato:
Add support for accessing `JsRuntimeState` in ops. This is supported both in
fast and slow paths of ops.


ops/op2/signature.rs
```diff
@@ -658,6 +662,8 @@ pub enum SignatureError {
   InvalidWherePredicate(String),
   #[error("State may be either a single OpState parameter, one mutable #[state], or multiple immultiple #[state]s")]
   InvalidOpStateCombination,
+  #[error("JsRuntimeState may only be used in one parameter")]
+  InvalidMultipleJsRuntimeState,
```
mmastrac, line 666: Love it. This should probably happen for the other single cardinality params too.

mmastrac (approved):
LGTM

PR #216, opened 2023-09-19
chore: enable source map handling for TS example

lucacasonato:
Also always enables `Error.prepareStackTrace` so that users of
`deno_core` don't need to set it up themselves.


mmastrac (approved):
LGTM

PR #224, opened 2023-09-20
feat: add core.currentUserCallSite()

lucacasonato:
This function returns the first call site in the current stack that is
"user" code, i.e. has a file name starting with neither `ext:` or
`node:`.

### DIFF
--- a/core/ops_builtin_v8.rs
+++ b/core/ops_builtin_v8.rs
@@ -803,6 +803,15 @@ pub struct Location {
   column_number: u32,
 }
 
+fn write_line_and_col_to_ret_buf(
+  ret_buf: &mut [u8],
+  line_number: u32,
+  column_number: u32,
+) {
+  ret_buf[0..4].copy_from_slice(&line_number.to_le_bytes());
+  ret_buf[4..8].copy_from_slice(&column_number.to_le_bytes());
+}
+
 // Returns:
 // 0: no source mapping performed, use original location
 // 1: mapped line and column, but not file name. new line and column are in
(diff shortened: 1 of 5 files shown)


core/ops_builtin_v8.rs
```diff
@@ -874,6 +879,73 @@ pub fn op_apply_source_map_filename(
...
+    let file_name = frame
+      .get_script_name(scope)
+      .unwrap()
+      .to_rust_string_lossy(scope);
+    // TODO: this condition should be configurable. It's a CLI assumption.
+    if (file_name.starts_with("ext:") || file_name.starts_with("node:"))
+      && i != frame_count - 1
+    {
+      continue;
+    }
```
bartlomieju, line 905: Seems fine for now, but yeah, in the future this should be configurable

bartlomieju (approved):

core/ops_builtin_v8.rs
```diff
@@ -874,6 +879,73 @@ pub fn op_apply_source_map_filename(
...
 }
 
+#[op2(core)]
+#[string]
+pub fn op_current_user_call_site(
+  scope: &mut v8::HandleScope,
+  js_runtime_state: &mut JsRuntimeState,
+  #[buffer] ret_buf: &mut [u8],
+) -> String {
+  let stack_trace = v8::StackTrace::current_stack_trace(scope, 10).unwrap();
```
nayeemrmn, line 889: This might be missing async stack frames, we do the whole `prepareStackTrace()` thing in `JsError::from_exception()` because of that.

bartlomieju: Oh! I couldn't remember the reason why we did that, but yeah, it's the async stack frames

lucacasonato: That's actually fine for my uses. It's ok to say that it will capture the first _sync_ user call site above the call. The use case I have is `Deno.test`'s location finder, which in both `Deno.test` and `t.step()` will have the user's call site synchronously in a stack frame above the current one. This works for `t.step()`, because we capture the stack trace before suspending at the first `await` in the function.

bartlomieju: Could you verify it that it still works if you have a top-level await before a `Deno.test` declaration?

PR #226, opened 2023-09-21
feat: add JsRuntime::op_names

igorzi:
This is needed in order to implement `Deno.metrics()` equivalent in Rust.

### DIFF
--- a/core/runtime/jsruntime.rs
+++ b/core/runtime/jsruntime.rs
@@ -1180,6 +1180,18 @@ impl JsRuntime {
     state.op_state.clone()
   }
 
+  /// Returns the runtime's op names, ordered by OpId.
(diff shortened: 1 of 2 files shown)


core/runtime/jsruntime.rs
```diff
@@ -1180,6 +1180,18 @@ impl JsRuntime {
     state.op_state.clone()
   }
 
+  /// Returns the runtime's op names, ordered by OpId.
+  pub fn op_names(&self) -> Vec<String> {
```
bartlomieju, line 1184: Since `OpDecl.name` is a `&'static str` maybe we can return `Vec<&'static str>` to avoid excessive copies?

igorzi: makes sense. done.

bartlomieju (approved):

PR #236, opened 2023-09-23
fix(ops): roundtrip external null pointers as JS null

### LINKED ISSUE #235: op2 external null pointer should be JS null
From discussion with @aapoalas:
Constructing v8:Externals with null pointers is undefined behaviour and may cause problems once pointer compression is turned on.
Instead they should roundtrip as JS null values, matching what happens in the Fast API path.

```rs
#[op2(core, fast)]
fn op_external_make_null() -> *const std::ffi::c_void {
  0 as _
}
```

lrowe:
This matches the behaviour of the Fast API path.

Constructing v8:Externals with null pointers is undefined behaviour and may cause problems once pointer compression is turned on.

Closes #235

### DIFF
--- a/core/runtime/ops_rust_to_v8.rs
+++ b/core/runtime/ops_rust_to_v8.rs
@@ -278,7 +278,13 @@ to_v8_retval!((f32, f64): |value, rv| rv.set_double(value as _));
 // Heavier primitives with no retval shortcuts
 //
 
-to_v8!((*const c_void, *mut c_void): |value, scope| v8::External::new(scope, value as _));
+to_v8!((*const c_void, *mut c_void): |value, scope| {
+  if value.is_null() {
+    v8::Local::<v8::Value>::from(v8::null(scope))
+  } else {
+    v8::Local::<v8::Value>::from(v8::External::new(scope, value as _))
+  }
+});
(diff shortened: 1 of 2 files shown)


core/runtime/ops_rust_to_v8.rs
```diff
@@ -278,7 +278,13 @@ to_v8_retval!((f32, f64): |value, rv| rv.set_double(value as _));
 // Heavier primitives with no retval shortcuts
 //
 
-to_v8!((*const c_void, *mut c_void): |value, scope| v8::External::new(scope, value as _));
+to_v8!((*const c_void, *mut c_void): |value, scope| {
```
mmastrac, line 281: Perfect.. let's implement `to_v8_retval` as well since we have the extra-fast path of null.

lrowe: I'm not sure how to do this since there is no `ReturnValue.set_external` and `to_v8_retval` does not pass scope to its callback, which is necessary for constructing an `External`.

https://v8.github.io/api/head/classv8_1_1ReturnValue.html

aapoalas (approved):

core/runtime/ops_rust_to_v8.rs
mmastrac: My bad, we're good with this!

mmastrac (approved):

PR #240, opened 2023-09-25
feat(ops): support *mut u8 and *const u8

### LINKED ISSUE #229: op2 buffer pointers support
Needed for ext/ffi.

```rust
#[op2(fast)]
pub fn op_ffi_ptr_of(
  #[buffer] buf: *const u8
) -> Result<*const c_void, AnyError> {
  return Ok(buf as *const c_void)
}
```

mmastrac:
Fixes #229. Note that we cannot pass the address of zero-length buffers because of a v8 bug.


bartlomieju (approved):
LGTM

PR #244, opened 2023-09-27
feat(ops): zero-cost-if-unused, pluggable metrics for op2

mmastrac:
Introduces a basic metrics API for op2 that is zero cost when unused, pluggable-per-op and a flexible system that we can add more to in the future.

As we remove op1 from the system, we will likely be able to simplify the async portion of this metrics system. Right now we're juggling the old and new op systems.

The metrics function gets a reference to the op's per-realm OpCtx, and a metrics event (`Enter`, `Leave`, `Exception`). Async function also have `LeaveAsync`, `ExceptionAsync` if they return a value/error asynchronously.

```rust
  let mut runtime = JsRuntime::new(RuntimeOptions {
    extensions: vec![test_ext::init_ops()],
    op_metrics_fn: Some(Box::new(move |op| {
      if !op.name.starts_with("op_async") && !op.name.starts_with("op_sync") {
        return None;
      }
      Some(Rc::new(move |op, metrics| {
        println!("{} {:?}", op.name, metrics);
      }))
    })),
    ..Default::default()
  });
```

### DIFF
--- a/core/extensions.rs
+++ b/core/extensions.rs
@@ -100,7 +100,7 @@ pub struct OpDecl {
   pub is_v8: bool,
   pub arg_count: u8,
   pub(crate) slow_fn: OpFnRef,
-  pub(crate) slow_fn_metrics: Option<OpFnRef>,
+  pub(crate) slow_fn_metrics: OpFnRef,
   pub(crate) fast_fn: Option<FastFunction>,
   pub(crate) fast_fn_metrics: Option<FastFunction>,
 }
@@ -125,9 +125,9 @@ impl OpDecl {
       is_v8,
       arg_count,
       slow_fn,
-      slow_fn_metrics: None,
+      slow_fn_metrics: slow_fn,
       fast_fn,
-      fast_fn_metrics: None,
+      fast_fn_metrics: fast_fn,
     }
   }
 
@@ -150,7 +150,7 @@ impl OpDecl {
       is_v8: false,
       arg_count,
       slow_fn,
-      slow_fn_metrics: Some(slow_fn_metrics),
+      slow_fn_metrics,
       fast_fn,
       fast_fn_metrics,
     }
--- a/core/extensions.rs
+++ b/core/extensions.rs
@@ -175,9 +175,9 @@ impl OpDecl {
   pub const fn with_implementation_from(self, from: &Self) -> Self {
     Self {
       slow_fn: from.slow_fn,
-      slow_fn_metrics: from.slow_fn_metrics,
+      slow_fn_with_metrics: from.slow_fn_with_metrics,
       fast_fn: from.fast_fn,
-      fast_fn_metrics: from.fast_fn_metrics,
+      fast_fn_with_metrics: from.fast_fn_with_metrics,
       ..self
     }
   }
--- a/core/extensions.rs
+++ b/core/extensions.rs
@@ -99,8 +99,10 @@ pub struct OpDecl {
   pub is_unstable: bool,
   pub is_v8: bool,
   pub arg_count: u8,
-  pub(crate) v8_fn_ptr: OpFnRef,
+  pub(crate) slow_fn: OpFnRef,
+  pub(crate) slow_fn_metrics: Option<OpFnRef>,
   pub(crate) fast_fn: Option<FastFunction>,
+  pub(crate) fast_fn_metrics: Option<FastFunction>,
 }
 
 impl OpDecl {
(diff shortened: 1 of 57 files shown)


core/extensions.rs
```diff
@@ -99,8 +99,10 @@ pub struct OpDecl {
   pub is_unstable: bool,
   pub is_v8: bool,
   pub arg_count: u8,
-  pub(crate) v8_fn_ptr: OpFnRef,
+  pub(crate) slow_fn: OpFnRef,
+  pub(crate) slow_fn_metrics: OpFnRef,
   pub(crate) fast_fn: Option<FastFunction>,
+  pub(crate) fast_fn_metrics: Option<FastFunction>,
```
bartlomieju, line 105: Could you describe the difference between these fields?

core/ops.rs
```diff
@@ -147,10 +168,17 @@ impl OpCtx {
...
+    metrics_fn: Option<OpMetricsFn>,
   ) -> Self {
     let mut fast_fn_c_info = None;
 
-    if let Some(fast_fn) = &decl.fast_fn {
+    let fast_fn = if metrics_fn.is_some() {
+      &decl.fast_fn
+    } else {
+      &decl.fast_fn_metrics
+    };
```
bartlomieju, line 179: A comment here would be useful, I don't understand what's going on :D

core/extensions.rs
bartlomieju: Maybe they should be named `*_with_metrics`?

core/runtime/jsruntime.rs
```diff
@@ -377,6 +380,10 @@ pub struct RuntimeOptions {
   /// executed tries to load modules.
   pub module_loader: Option<Rc<dyn ModuleLoader>>,
 
+  /// Provide a function that may optionally provide a metrics collector
+  /// for a given op.
+  pub op_metrics_fn: Option<OpMetricsFactoryFn>,
```
bartlomieju, line 385: Could you include example either on `OpMetricsFactoryFn` or directly in the docstring for this field?

core/runtime/jsruntime.rs
```diff
@@ -1965,7 +1976,7 @@ impl JsRuntime {
           break;
         };
         // TODO(mmastrac): If this task is really errored, things could be pretty bad
-        let (promise_id, op_id, resp) = item.unwrap();
+        let PendingOp(promise_id, op_id, resp, metrics_event) = item.unwrap();
```
bartlomieju, line 1979: Is it really free if each now now returns `metrics_event`?

core/ops.rs
```diff
@@ -120,6 +123,22 @@ pub fn to_op_result<R: Serialize + 'static>(
...
+  Enter,
+  /// Left an op synchronously.
+  Leave,
+  /// Left an op asynchronously.
+  LeaveAsync,
+  /// Left an op synchronously with an exception.
+  Exception,
+  /// Left an op asynchronously with an exception.
+  ExceptionAsync,
+}
```
bartlomieju, line 138: Bikeshedding time: `Completed(Async)` and `Error(Async)`?

core/runtime/jsruntime.rs
mmastrac: I wanted to hold off on that while we iterate on the API a bit. This may change or evolve and I wanted to plug it into deno itself before committing to the API fully.

core/runtime/jsruntime.rs
mmastrac: It's effectively free -- we're just setting an extra boolean in a tuple struct here. Once we have a chance to revisit this async dispatch, it will likely turn out to be cheaper.

core/ops.rs
mmastrac: `Started`/`Completed`?

bartlomieju: Previously we used "dispatched" and "completed". That works too

core/runtime/jsruntime.rs
mmastrac: There's a test in this PR that demonstrates the current version of the API, but I'm sure this will evolve:

https://github.com/denoland/deno_core/pull/244/files#diff-cdffb2ed85028df867198ed8c4cfb966082c88913fb9722eb65bec42aa3de9e3R633

core/ops.rs
mmastrac: I went with Dispatched/Completed.

bartlomieju (approved):
LGTM, really good first pass; curious to see where we'll take it from here.

PR #246, opened 2023-09-29
fix(ops): correctly handle multiple array buffers in op_deserialize

marcosc90:
This PR fixes `op_deserialize` when multiple array buffers are transferred.

fixes  https://github.com/denoland/deno/issues/20727

### DIFF
--- a/core/runtime/tests/mod.rs
+++ b/core/runtime/tests/mod.rs
@@ -3,6 +3,7 @@ use crate::error::AnyError;
 use crate::JsRuntime;
 use crate::OpState;
 use crate::RuntimeOptions;
+use crate::CrossIsolateStore;
 use deno_ops::op;
 use serde_v8::JsBuffer;
 use std::cell::RefCell;
@@ -82,6 +83,7 @@ fn setup(mode: Mode) -> (JsRuntime, Arc<AtomicUsize>) {
     get_error_class_fn: Some(&|error| {
       crate::error::get_custom_error_class(error).unwrap()
     }),
+    shared_array_buffer_store: Some(CrossIsolateStore::default()),
     ..Default::default()
   });
 
--- a/core/runtime/tests/mod.rs
+++ b/core/runtime/tests/mod.rs
@@ -1,9 +1,9 @@
 use crate as deno_core;
 use crate::error::AnyError;
+use crate::CrossIsolateStore;
 use crate::JsRuntime;
 use crate::OpState;
 use crate::RuntimeOptions;
-use crate::CrossIsolateStore;
 use deno_ops::op;
 use serde_v8::JsBuffer;
 use std::cell::RefCell;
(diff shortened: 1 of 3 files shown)


marcosc90:
For some reason `op_serialize` work differently in `deno_core` (at least on tests), so can't make the test work in `deno_core` (it passes)

```js

function testIssue20727() {
  // https://github.com/denoland/deno/issues/20727
  Deno.core = Deno.core ?? [Deno.internal].core;
 
  const ab = new ArrayBuffer(10);
  const transferList = [ab]
  Deno.core.ops.op_serialize(
    { ab },
    { transferredArrayBuffers: transferList },
  );
  // transferList is [0] in deno, and [{}] in deno_core
  const data = {
    array1: new Uint32Array([]),
    array2: new Float32Array([]),
  };

  const transferredArrayBuffers = [
    data.array1.buffer,
    data.array2.buffer,
  ];
  const serializedMultipleTransferredBuffers = Deno.core.ops.op_serialize(
    { id: 2, data },
    { transferredArrayBuffers },
  );

  // should not throw
  const x = Deno.core.ops.op_deserialize(
    serializedMultipleTransferredBuffers,
    { transferredArrayBuffers },
  );
}

testIssue20727();
```

`transferList` after `op_serialize` is `[0]` when running from deno, and `[{}]` when running from `deno_core` test file

The above test fails on deno, and works when using this branch.

marcosc90:
https://github.com/denoland/deno_core/blob/52a50c1ffecc491e471114fca65144923cea0b90/core/ops_builtin_v8.rs#L402

`state.shared_array_buffer_store` is empty when called from `deno_core` tests which is why I can't test it. Any guidance is appreciated, if not I can place the test in deno.

bartlomieju:
@marcosc90 I will take a look at it tomorrow and try to give you some pointers 👍

lucacasonato (approved):
LGTM

lucacasonato:
@marcosc90 `deno_core` tests are probably not passing the `shared_array_buffer_store` in `JsRuntimeOptions`

bartlomieju:
@marcosc90 Luca is right here - you'll need to pass `JsRuntimeOptions.shared_array_buffer_store` and the tests should pass at that point. Could you add the test?

core/runtime/tests/mod.rs
```diff
@@ -82,6 +83,7 @@ fn setup(mode: Mode) -> (JsRuntime, Arc<AtomicUsize>) {
     get_error_class_fn: Some(&|error| {
       crate::error::get_custom_error_class(error).unwrap()
     }),
+    shared_array_buffer_store: Some(CrossIsolateStore::default()),
```
marcosc90, line 86: @bartlomieju @lucacasonato haven't found any sample where `shared_array_buffer_store` is set, but this did the trick.

The test fails on `main` (when `shared_array_buffer_store` is set)

bartlomieju: That's correct!

bartlomieju (approved):
LGTM, nice fix, thank you Marcos!

PR #247, opened 2023-10-02
feat: FeatureChecker

bartlomieju:
A simple struct that will allow us to easily access a mapping of enabled
features that can be specified on startup of `JsRuntime` - the Deno CLI
will be using feature flags to update this struct.


core/feature_checker.rs
```diff
@@ -0,0 +1,43 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+
+use std::collections::BTreeMap;
+
+#[derive(Default, Debug)]
+pub struct FeatureChecker {
+  features: BTreeMap<&'static str, bool>,
```
dsherret, line 7: Maybe `enabled_features: HashSet<&'static str>`

core/feature_checker.rs
```diff
@@ -0,0 +1,42 @@
...
+
+  // TODO(bartlomieju): remove this.
+  /// Check if `--unstable` flag has been passed. If not then exit the process
+  /// with exit code 70.
+  pub fn check_legacy_unstable(&self, api_name: &str) {
+    if !self.legacy_unstable {
+      eprintln!(
+        "Unstable API '{api_name}'. The --unstable flag must be provided."
+      );
+      std::process::exit(70);
```
dsherret, line 39: Maybe we should return an error instead of printing and exiting the process.

bartlomieju: The current unstable checker in the CLI exits the process with this code and I'd prefer to keep it that way. I left TODOs that this should be removed, hopefully we can migrate away from a single `--unstable` flag for catch all for Deno 2.

core/feature_checker.rs
bartlomieju: Probably a good idea, it's not like there's gonna be entry set to `false` anyway

core/feature_checker.rs
dsherret: Why do we do that? Wouldn't it be better to error? Perhaps we should rename this to `ensure_legacy_unstable_or_exit`.

bartlomieju: Because we didn't want people to feature check if `--unstable` flag is enabled by calling a function and catching an exception.

bartlomieju: I renamed the method, I like your suggestion better

dsherret (approved):
LGTM

mmastrac (approved):
LGTM!

PR #248, opened 2023-10-02
feat(ops): support #[anybuffer]

mmastrac:
Support `#[anybuffer]` in fastcalls for slices/pointers. This will accept any ArrayBuffer or ArrayBufferView.

This PR is fairly large because I needed to clean up the buffer argument handling a bit.


ops/op2/dispatch_fast.rs
```diff
@@ -1,3 +1,5 @@
+use crate::op2::V8SignatureMappingError;
```
bartlomieju, line 1: Move under the copyright

bartlomieju (approved):
LGTM, verified that it fixes the problem in Deno

PR #252, opened 2023-10-03
feat(core,ops): add `*mut v8::Isolate` as a dangerous but fast arg

mmastrac:
`op2` ops can request a `*mut v8::Isolate` which allows them to do all sorts of dangerous things, but requires an `unsafe` block to make it into something that can be used. This is a "I know what I'm doing" op2 argument so we leave it as a pointer.

Generally you'll want to use this with `nofast` to avoid the fastcall "no-js" scope.

With this patch HTTP 131k -> 133k, async ops deferred 743k ->  761k.

```
test baseline                               ... bench:         506 ns/iter (+/- 20)
test bench_op_v8_global                     ... bench:      23,064 ns/iter (+/- 2,293)
test bench_op_v8_global_scope               ... bench:      30,598 ns/iter (+/- 2,850)
test bench_op_v8_isolate_nofast             ... bench:       5,903 ns/iter (+/- 270)
test bench_op_v8_local                      ... bench:       3,232 ns/iter (+/- 580)
test bench_op_v8_local_nofast               ... bench:       6,184 ns/iter (+/- 431)
test bench_op_v8_local_scope                ... bench:      12,652 ns/iter (+/- 292)
test bench_op_v8_scope                      ... bench:      11,727 ns/iter (+/- 547)
```


bartlomieju (approved):
LGTM, looks a bit scary, but seeing changes to `op_set_has_tick_scheduled` et al, it seems worth it 👍

PR #254, opened 2023-10-03
feat(ops): allow JsBuffer arguments to be fast

mmastrac:
Only supported in sync ops for now. Reduces the need for scope creation for all JsBuffer usage (except in detaching mode).


bartlomieju (approved):
LGTM, this is very nice 👍 is there any visible perf difference?

mmastrac:
I can definitely run a bench, but it should be 4-5x faster by my estimates. We save both the slowcall cost and scope cost which are both pretty heavy.

mmastrac:
Added benchmark in https://github.com/denoland/deno_core/pull/270 -- it wasn't quite as big as I hoped, so there's probably more room for improvement (~30% faster). Need to follow up on this one, maybe in rusty_v8.

PR #256, opened 2023-10-04
fix: op2 #[anybuffer] byte offset

bartlomieju:
Trying to fix WPT for WS which construct `Uint8Array` with byte offset of 2 and length 8;
and expect a Uint8Array of length 6 to be returned.


mmastrac (approved):
LGTM!

PR #266, opened 2023-10-12
fix: move Deno builtin init to bindings.js

littledivy:


bartlomieju:
Closing for now; we need to go back to `rusty_v8` and figure this thing with snapshots.

PR #269, opened 2023-10-15
feat(ops): fast async(lazy) and async(deferred)

mmastrac:
This doesn't move the needle much on async ops that are submitted and immediately awaited upon, but it does offer slighly better latency when taking a promise from an async op and then waiting on it after performing other work.

One of the big refactorings in this PR is ensuring that async and sync fast ops share the same code path. The same refactoring can be applied to slow ops, but that would be slightly too much to tackle in this PR.

```
running 14 tests
test baseline                             ... bench:       1,215 ns/iter (+/- 55)
test baseline_promise                     ... bench:      49,198 ns/iter (+/- 1,829)
test baseline_sync                        ... bench:      59,163 ns/iter (+/- 2,600)
test bench_op_async_void                  ... bench:     144,511 ns/iter (+/- 4,112)
test bench_op_async_void_deferred         ... bench:   2,069,403 ns/iter (+/- 51,417)
test bench_op_async_void_deferred_nofast  ... bench:   2,074,620 ns/iter (+/- 94,785)
test bench_op_async_void_deferred_return  ... bench:   2,100,433 ns/iter (+/- 72,988)
test bench_op_async_void_lazy             ... bench:   1,972,197 ns/iter (+/- 73,343)
test bench_op_async_void_lazy_nofast      ... bench:   1,946,581 ns/iter (+/- 57,475)
test bench_op_async_yield                 ... bench:   2,035,129 ns/iter (+/- 71,406)
test bench_op_async_yield_deferred        ... bench:   2,064,232 ns/iter (+/- 523,829)
test bench_op_async_yield_deferred_nofast ... bench:   2,040,228 ns/iter (+/- 81,427)
test bench_op_async_yield_lazy            ... bench:   3,083,485 ns/iter (+/- 99,416)
test bench_op_async_yield_lazy_nofast     ... bench:   3,015,587 ns/iter (+/- 68,670)
```

### DIFF
--- a/core/benches/ops/async.rs
+++ b/core/benches/ops/async.rs
@@ -1,0 +1,210 @@
+use bencher::*;
+use deno_core::error::generic_error;
+use deno_core::*;
+use std::ffi::c_void;
+
+deno_core::extension!(
+  testing,
+  ops = [
+    op_void,
+    op_make_external,
+    op_async_void,
+    op_async_void_lazy,
+    op_async_void_lazy_nofast,
+    op_async_void_deferred,
+    op_async_void_deferred_nofast,
+    op_async_yield,
+    op_async_yield_lazy,
+    op_async_yield_lazy_nofast,
+    op_async_yield_deferred,
+    op_async_yield_deferred_nofast,
+  ],
+);
+
+#[op2(fast)]
+pub fn op_void() {
+}
+
+#[op2(fast)]
+pub fn op_make_external() -> *const c_void {
+  std::ptr::null()
+}
+
+#[op2(async)]
+pub async fn op_async_void() {}
+
+#[op2(async)]
+pub async fn op_async_yield() {
+  tokio::task::yield_now().await
+}
+
+#[op2(async(lazy), fast)]
+pub async fn op_async_yield_lazy() {
+  tokio::task::yield_now().await
+}
+
+#[op2(async(lazy), nofast)]
+pub async fn op_async_yield_lazy_nofast() {
+  tokio::task::yield_now().await
+}
+
+#[op2(async(deferred), fast)]
+pub async fn op_async_yield_deferred() {
+  tokio::task::yield_now().await
+}
+
+#[op2(async(deferred), nofast)]
+pub async fn op_async_yield_deferred_nofast() {
+  tokio::task::yield_now().await
+}
+
+#[op2(async(lazy), fast)]
+pub async fn op_async_void_lazy() {}
+
+#[op2(async(lazy), nofast)]
+pub async fn op_async_void_lazy_nofast() {}
+
+#[op2(async(deferred), fast)]
+pub async fn op_async_void_deferred() {}
+
+#[op2(async(deferred), nofast)]
+pub async fn op_async_void_deferred_nofast() {}
+
+fn bench_op(
+  b: &mut Bencher,
+  count: usize,
+  op: &str,
+  arg_count: usize,
+  call: &str,
+) {
+  #[cfg(not(feature = "unsafe_runtime_options"))]
+  unreachable!(
+    "This benchmark must be run with --features=unsafe_runtime_options"
+  );
+
+  let tokio = tokio::runtime::Builder::new_current_thread().build().unwrap();
+  let mut runtime = JsRuntime::new(RuntimeOptions {
+    extensions: vec![testing::init_ops_and_esm()],
+    // We need to feature gate this here to prevent IDE errors
+    #[cfg(feature = "unsafe_runtime_options")]
+    unsafe_expose_natives_and_gc: true,
+    ..Default::default()
+  });
+  let err_mapper =
+    |err| generic_error(format!("{op} test failed ({call}): {err:?}"));
+
+  let args = (0..arg_count)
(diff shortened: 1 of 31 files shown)


core/benches/ops/async.rs
```diff
@@ -0,0 +1,222 @@
+use bencher::*;
```
bartlomieju, line 1: Add copyright?

core/benches/ops/async.rs
```diff
@@ -0,0 +1,222 @@
...
+    return accum;
+  }}
+  
+  %PrepareFunctionForOptimization(bench);
+  await bench();
+  %OptimizeFunctionOnNextCall(bench);
+  await bench();
+
+  return bench;
+}}
```
bartlomieju, line 141: How about making this a separate file that uses `include_str!` here? I find it hard for some of the tests to figure out what's going on without syntax highlighting and this one doesn't use any format strings, which makes it easy to move to a separate file.

core/benches/ops/async.rs
```diff
@@ -0,0 +1,222 @@
...
+  let bench = v8::Global::new(&mut scope, bench);
+  drop(scope);
+  drop(guard);
+  b.iter(move || {
+    tokio.block_on(async {
+      let guard = tokio.enter();
+      runtime.call_and_await(&bench).await.unwrap();
+      drop(guard);
+    });
+  });
```
bartlomieju, line 164: Not a blocker, but have you profiled it using `samply`? One time I did that for other benchmark I saw a surprising profile where most of the time is spent in runtime creation/dropping.

core/ops.rs
```diff
@@ -156,6 +156,7 @@ pub struct OpCtx {
   pub(crate) context_state: Rc<RefCell<ContextState>>,
   /// If the last fast op failed, stores the error to be picked up by the slow op.
   pub(crate) last_fast_error: UnsafeCell<Option<AnyError>>,
+  // pub(crate) last_fast_result: UnsafeCell<Option<Box<dyn std::any::Any>>>,
```
bartlomieju, line 159: Remove?

core/runtime/ops.rs
```diff
@@ -173,16 +174,22 @@ pub fn map_async_op_infallible<R: 'static>(
...
-    Poll::Ready(res) => {
-      if ctx.metrics_enabled() {
-        dispatch_metrics_async(ctx, OpMetricsEvent::Completed);
+  let pinned =
+    match pinned.poll_unpin(&mut Context::from_waker(noop_waker_ref())) {
+      Poll::Pending => pinned,
+      Poll::Ready(res) => {
+        // TODO(mmastrac): optimize this so we don't double-allocate
+        if deferred {
+          ready(res).boxed_local()
```
bartlomieju, line 183: Are deferred ops tracking metrics?

core/runtime/ops.rs
```diff
@@ -239,20 +247,26 @@ pub fn map_async_op_fallible<R: 'static, E: Into<Error> + 'static>(
...
-      if ctx.metrics_enabled() {
-        if res.2.is_err() {
-          dispatch_metrics_async(ctx, OpMetricsEvent::Error);
+  let pinned =
+    match pinned.poll_unpin(&mut Context::from_waker(noop_waker_ref())) {
+      Poll::Pending => pinned,
+      Poll::Ready(res) => {
+        // TODO(mmastrac): optimize this so we don't double-allocate
+        if deferred {
+          ready(res).boxed_local()
```
bartlomieju, line 256: Ditto

bartlomieju (commented):
Could you add a test for `async(deferred)` to `core/ops2/op2/test_cases/async/`?

core/benches/ops/async.rs
mmastrac: This one does use a format string -- see `{call}`, `{op}`, and `{count}`. I am not sure of a good way to organize this code externally because it's pretty dynamic and we specifically need to generate JS.

I can try some experiments to clean it up a bit.

core/benches/ops/async.rs
mmastrac: I haven't in this case, but I included the baseline so we can see the overhead of running the same overhead code but with no work. It really comes down to spinning the event loop being slow :/

```
test baseline                             ... bench:       1,194 ns/iter (+/- 34)
test bench_op_async_void                  ... bench:     149,005 ns/iter (+/- 3,891)
```

core/ops.rs
mmastrac: Good spot, thanks

core/runtime/ops.rs
mmastrac: They work but I added a test to confirm because it's not obvious.

core/benches/ops/async.rs
mmastrac: Ack, we need copyright_checker here

core/benches/ops/async.rs
mmastrac: I replaced these with a template approach -- the files are valid JS so they can be formatted w/`deno fmt` and are somewhat readable

bartlomieju (approved):
LGTM, can you update the markdown docs before landing?

PR #274, opened 2023-10-23
feat(ops): support fast(op_other) for fastcall alternatives

mmastrac:
This is the final op2 feature to replace the last op1 ops in deno.

```rust
// Unused scope would normally make this a slow-only op
#[op2(fast(op_fast))]
fn op_slow(_scope: &v8::HandleScope, a: u32, b: u32) -> u32 {
  a + b
}

#[op2(fast)]
fn op_fast(a: u32, b: u32) -> u32 {
  a + b
}
```

### DIFF
--- a/ops/op2/README.md
+++ b/ops/op2/README.md
@@ -71,6 +71,17 @@ used if you really know what you are doing.
 Lazy `async(deferred)` calls _may_ be fastcalls, though the resolution will
 still happen on a slow path.
 
+## fastcalls
+
+`op2` requires fastcall-compatible ops to be annotated with `fast`. If you wish to avoid fastcalls
+for some reason (this is unlikely), you can specify `nofast` instead.
+
+You may also choose an alternate op function to use as the fastcall equivalent to a slow function. In
+this case, you can specify `fast(op_XYZ)`. The other op must be decorated with `#[op2(fast)]`, and does
+not need to be registered. When v8 optimized the slow function to a fastcall, it will switch the implementation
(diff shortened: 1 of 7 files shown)


ops/op2/README.md
```diff
@@ -71,6 +71,20 @@ used if you really know what you are doing.
...
 still happen on a slow path.
 
+## fastcalls
+
+`op2` requires fastcall-compatible ops to be annotated with `fast`. If you wish
+to avoid fastcalls for some reason (this is unlikely), you can specify `nofast`
+instead.
+
+You may also choose an alternate op function to use as the fastcall equivalent
+to a slow function. In this case, you can specify `fast(op_XYZ)`. The other op
```
bartlomieju, line 81: Could we go with `fast(slow_fallback = op_XYZ)` instead?

bartlomieju (commented):
Looks okay, but I have one suggestion and unfortunately tests are failing too :(

ops/op2/README.md
mmastrac: Slow ops can have multiple fast versions (rather than the other way around), so this way makes more sense, especially if we get multiple overrides.

mmastrac: Also, I struggled a bit with getting `x = y` parameters to work with the current setup. I think the parser needs a bit of work if we want to support that style of parameter.

Given that there's just the one op right now, I don't know if it's worth working on that until we have a more useful use case for op args of that style.

PR #279, opened 2023-10-24
chore: mark op1 as deprecated and port most core ops

mmastrac:
Before we remove `#[op]`, give embedders some notice to switch over to `#[op2]`. Note that I could not port the wasm ops yet -- we'll address that in #275


core/examples/hello_world.rs
```diff
@@ -2,21 +2,12 @@
...
-use deno_core::op;
-use deno_core::Extension;
-use deno_core::JsRuntime;
-use deno_core::Op;
-use deno_core::RuntimeOptions;
-
-// This is a hack to make the `#[op]` macro work with
-// deno_core examples.
-// You can remove this:
 use deno_core::*;
```
mmastrac, line 5: This is a more convenient import for most of the examples, so leaving it here (it's no longer required though)

core/examples/http_bench_json_ops/main.rs
```diff
@@ -1,26 +1,13 @@
...
-use deno_core::op;
-use deno_core::AsyncRefCell;
-use deno_core::AsyncResult;
-use deno_core::JsBuffer;
-use deno_core::JsRuntimeForSnapshot;
-use deno_core::Op;
-use deno_core::OpState;
-use deno_core::Resource;
-use deno_core::ResourceId;
+use deno_core::*;
```
mmastrac, line 3: Ditto

core/runtime/tests/ops.rs
```diff
@@ -288,42 +289,6 @@ fn test_op_detached_buffer() {
     .unwrap();
 }
 
-#[test]
```
mmastrac, line 291: Unused feature, just removing for now

core/runtime/tests/snapshot.rs
```diff
@@ -1,4 +1,5 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+use crate as deno_core;
```
mmastrac, line 2: I might remove the `op2(core)` and just do this in the rare files in deno_core that have ops.

bartlomieju: I like this idea

core/examples/http_bench_json_ops/main.rs
bartlomieju, line 1: We can remove this file altogether I think. It's named `http_bench_json_ops` - it has nothing to do with JSON ops that we did in 2020 or 2021 :D

bartlomieju (approved):

core/runtime/tests/snapshot.rs
mmastrac: I'll follow up with it in a later pass!

PR #280, opened 2023-10-24
build: run select examples in ci

nayeemrmn:


bartlomieju (approved):
LGTM, thanks bringing this.

For posterity, added because this test was removed in: https://github.com/denoland/deno/pull/20966

PR #286, opened 2023-10-26
perf: Use V8 Deno builtins

bartlomieju:


core/runtime/bindings.rs
```diff
@@ -179,17 +180,26 @@ pub(crate) fn initialize_context<'s>(
...
-      "ExtrasBindingObject.console",
-    );
+    let console_obj: v8::Local<v8::Object> =
+      get(scope, extra_binding_obj, b"console", "%console");
+    let is_one_byte_fn: v8::Local<v8::Object> =
+      get(scope, extra_binding_obj, b"isOneByte", "%isOneByte");
+    let from_utf8_fn: v8::Local<v8::Object> =
+      get(scope, extra_binding_obj, b"fromUtf8", "%fromUtf8");
+    let to_utf8_fn: v8::Local<v8::Object> =
+      get(scope, extra_binding_obj, b"toUtf8", "%toUtf8");
```
mmastrac, line 190: I wouldn't worry about it for this patch, but I think we could just pass the entire extra_binding_obj into the function and allow the function to extract these various fields.

core/runtime/bindings.rs
```diff
@@ -130,10 +130,11 @@ pub(crate) fn initialize_context<'s>(
   codegen.push_str(include_str!("bindings.js"));
   _ = writeln!(
     codegen,
-    "Deno.__op__ = function(opFns, callConsole, console) {{"
+    "Deno.__op__ = function(opFns, callConsole, console, isOneByte, fromUtf8, toUtf8) {{"
   );
   if init_mode == InitMode::New {
     _ = writeln!(codegen, "Deno.__op__console(callConsole, console);");
+    _ = writeln!(codegen, "Deno.core = {{ ops: {{}}, asyncOps: {{}}, isOneByte, fromUtf8, toUtf8 }};");
```
mmastrac, line 137: We can probably remove the code from bindings.js that does the ops/asyncOps init.

core/01_core.js
```diff
@@ -853,8 +853,8 @@ for (let i = 0; i < 10; i++) {
       specifier,
     ) => ops.op_eval_context(source, specifier),
     createHostObject: () => ops.op_create_host_object(),
-    encode: (text) => ops.op_encode(text),
-    decode: (buffer) => ops.op_decode(buffer),
+    encode: (text) => new Uint8Array(toUtf8(text)),
+    decode: (buffer, ignoreBOM = false) => fromUtf8(buffer, ignoreBOM),
```
mmastrac, line 857: 🙏

core/01_core.js
```diff
@@ -853,8 +853,8 @@ for (let i = 0; i < 10; i++) {
       specifier,
     ) => ops.op_eval_context(source, specifier),
     createHostObject: () => ops.op_create_host_object(),
-    encode: (text) => ops.op_encode(text),
```
mmastrac, line 856: Remove these ops as well?

core/runtime/bindings.rs
bartlomieju: I agree, I spoke with @littledivy about this today and the whole `Deno.__op__ = ...` bit seems like something that should be looked at again and cleaned up.

core/runtime/bindings.rs
bartlomieju: I'm not sure I understand

mmastrac: https://github.com/denoland/deno_core/blob/main/core/runtime/bindings.js#L3

core/runtime/bindings.rs
mmastrac: Agreed...  op binding is pretty rough.

PR #288, opened 2023-10-29
chore: remove `core` from `op2` and just `use crate`

mmastrac:
This simplifies a lot of code as we no longer need to pass the `deno_core` namespace into _everything_.

There are only a handful of places we actually use ops in deno_core and it's pretty easy just to add `use crate as deno_core`.


mmastrac:
Works locally, but failing on CI for some reason. Digging in.

bartlomieju (approved):
LGTM, great clean up!

PR #290, opened 2023-10-30
Tweak documentation of v8_set_flags

mlafeldt:
We experienced some weird V8 segfault by not adding a fake first argument as done in https://github.com/denoland/deno/blob/v1.37/cli/util/v8.rs

There's also a note about it in the documentation of `V8::set_flags_from_command_line`, which I copied here.


littledivy (approved):
LGTM, thanks

PR #295, opened 2023-11-01
perf(ops): trim async op cruft

mmastrac:
Removes a lot of the old async machinery that is no longer used, and provides a path forward for op tracking, used in Deno test. The metrics object has been trimmed down significantly.

Before:
```
test bench_op_async_void                  ... bench:     104,595 ns/iter (+/- 4,252)
```

After:
```
test bench_op_async_void                  ... bench:      75,782 ns/iter (+/- 3,128)
```

Changes:

 - Don't create a scope for async ops unless or until we really need one
 - Start using the new metrics functions
 - Replace OpTracker with a new metrics tracker that is optional and zero-cost when unused


core/ops.rs
```diff
@@ -177,9 +158,9 @@ impl OpCtx {
     // version and require a slightly different set of arguments (for example, it may need the fastcall
     // callback information to get the `OpCtx`).
     let fast_fn = if metrics_fn.is_some() {
-      &decl.fast_fn
-    } else {
       &decl.fast_fn_with_metrics
+    } else {
+      &decl.fast_fn
```
bartlomieju, line 163: Ooops, a previous mistake?

core/ops_metrics.rs
```diff
@@ -1,92 +1,136 @@
...
-pub struct OpMetrics {
-  pub ops_dispatched: u64,
+pub struct OpMetricsSummary {
   pub ops_dispatched_sync: u64,
   pub ops_dispatched_async: u64,
-  // TODO(bartlomieju): this field is never updated
-  pub ops_dispatched_async_unref: u64,
-  pub ops_completed: u64,
-  pub ops_completed_sync: u64,
   pub ops_completed_async: u64,
```
bartlomieju, line 60: We should make sure the shape of the object returned from `Deno.metrics()` API stays the same - the fields can be zeroed out, but the shape should still be correct until the API is present on the `Deno` namespace.

core/ops_metrics.rs
```diff
@@ -1,92 +1,136 @@
...
 
-// TODO(@AaronO): track errors
-#[derive(Default, Debug)]
-pub struct OpsTracker {
-  ops: RefCell<Vec<OpMetrics>>,
+impl OpMetricsSummary {
+  /// Does this op have outstanding async op dispatches?
+  pub fn has_outstanding_ops(&self) -> bool {
+    self.ops_dispatched_async > self.ops_completed_async
+  }
```
bartlomieju, line 67: Doesn't it need to take ref/unref ops into account?

core/runtime/ops.rs
```diff
@@ -26,117 +23,6 @@ use v8::WriteOptions;
...
-  match op {
-    Err(err) => Either::Left(ready(OpResult::Err(OpError::new(
-      ctx.get_error_class_fn,
-      err,
-    )))),
-    Ok(fut) => Either::Right(fut.map(|r| OpResult::Ok(r.into()))),
-  }
-}
-
-pub fn queue_async_op<'s>(
```
bartlomieju, line 97: Lovely to see all of these go!

bartlomieju (approved):

core/ops.rs
mmastrac: Yeah 😓  We never actually enabled metrics in fast ops. I added the fast ops to the benchmark so it'll have some coverage.

core/ops_metrics.rs
mmastrac: It never did in the past.. those were always zero 🤷

PR #301, opened 2023-11-03
feat(core): typed externals

mmastrac:


core/external.rs
```diff
@@ -1,53 +1,258 @@
...
+    impl $crate::external::Externalizable for $type {
+      fn external_marker() -> usize {
+        // Use the address of a static mut as a way to get around lack of const TypeId. Because it is mutable, the
+        // compiler cannot collapse multiple definitions into one.
+        static mut DEFINITION: $crate::external::ExternalDefinition =
+          $crate::external::ExternalDefinition::new($name);
+        // Wash the pointer through black_box so the compiler cannot see what we're going to do with it and needs
+        // to assume it will be used for valid purposes.
+        let ptr = std::hint::black_box(unsafe { &mut DEFINITION } as *mut _);
+        ptr as usize
```
lucacasonato, line 17: I don't understand this comment. This code isn't `const` either? Why is this preferrable over `TypeId::of`?

lucacasonato (commented):
Wow, this is pretty scary stuff. We should really package this up into a totally safe API that is exposed to op2 via `#[external]` macro and is lifetime bound to the v8::External via GC.

lucacasonato (commented):
Wow, this is pretty scary stuff. We should really package this up into a totally safe API that is exposed to op2 via `#[external]` macro and is lifetime bound to the v8::External via GC.

core/external.rs
mmastrac: Ack, I didn't update this comment when the design changed. TypeId isn't really a great fit for this because the implementation details change fairly often and we don't necessarily want to be testing a u128 for identity.

mmastrac:
> Wow, this is pretty scary stuff. We should really package this up into a totally safe API that is exposed to op2 via `#[external]` macro and is lifetime bound to the v8::External via GC.

As far as I can see, that's just not possible here. We have a raw pointer and that's it. If we turn on the finalization registry, we take a big perf hit.

There's not a great way to do this in an impact-free way without Divy's cppgc stuff. If we want to use externals we need to ensure that lifetimes are controlled.

The code will panic in either debug or release mode if it detects a failure mode, so we'll very quickly discover any problems. We've been doing something similar on the HTTP side with slab IDs, so the approach works well in shaking out problems.

Unfortunately we need to go down to this level to extract the perf we need from HTTP and get rid of things like the slab allocator. I need a way to land the http external code without having to wait for that code to make its way through.

core/external.rs
```diff
@@ -1,53 +1,258 @@
...
+      );
+    }
+    self.ptr as _
+  }
+
+  /// Unsafely retrieves the underlying object from this pointer after validating it.
+  ///
+  /// # Safety
+  ///
+  /// This method is inherantly unsafe because we cannot know if the underlying memory has been deallocated at some point.
```
bartlomieju, line 107:
```suggestion
  /// This method is inherently unsafe because we cannot know if the underlying memory has been deallocated at some point.
```

core/external.rs
```diff
@@ -1,53 +1,258 @@
...
+    let validated_ptr = self.validate_pointer();
+    let external = std::ptr::addr_of!((*validated_ptr).external);
+    &*external
+  }
+
+  /// Unsafely takes the object from this external.
+  ///
+  /// # Safety
+  ///
+  /// This method is inherantly unsafe because we cannot know if
```
bartlomieju, line 121:
```suggestion
  /// This method is inherently unsafe because we cannot know if
```

core/external.rs
```diff
@@ -1,53 +1,258 @@
...
+
+  struct External1(u32);
+  external!(External1, "external 1");
+
+  struct External2(u32);
+  external!(External2, "external 2");
+
+  // Use the same name as External 1
+  struct External1b(u32);
+  external!(External1b, "external 1");
```
bartlomieju, line 151: Why do we allow same name for multiple externals? Is it meant to describe a "kind" (eg. `HTTPResource`)?

core/external.rs
```diff
@@ -1,53 +1,258 @@
...
+    let external = ExternalPointer::new(External1(1));
+    assert_eq!(unsafe { external.unsafely_deref() }.0, 1);
+    let ptr = external.into_raw();
+
+    let external = ExternalPointer::<External1>::from_raw(ptr);
+    assert_eq!(unsafe { external.unsafely_deref() }.0, 1);
+    assert_eq!(unsafe { external.unsafely_take() }.0, 1);
+  }
+
+  // If this test ever fails then our "pseudo type ID" system is not working as expected.
```
bartlomieju, line 184: Could you update this comment to explain why it's supposed to fail?

core/external.rs
```diff
@@ -1,53 +1,258 @@
...
+  pub fn test_external_markers() {
+    let m1 = External1::external_marker();
+    let m2 = External2::external_marker();
+    let m1b = External1b::external_marker();
+
+    assert_ne!(m1, m2);
+    assert_ne!(m1, m1b);
+  }
+
+  // If this test ever fails then our "pseudo type ID" system is not working as expected.
```
bartlomieju, line 195: Ditto

core/external.rs
mmastrac: This is to prevent accidentally allowing cross-external contamination if someone names an external with the same name in another module. We can't easily detect two externals with the same name, but we prevent them from being mixed up between multiple APIs and causing security/crash issues down the line.

bartlomieju (approved):

Issue #302, opened 2023-11-03
Removing core from op2 breaks downstream usage of deno_core containing crates


rscarson:
Referring to this change:
https://github.com/denoland/deno_core/commit/16970b70f3640cc0b3023eb342dcfbee93d1d112

`::deno_core` assumes that deno_core is directly in Cargo.toml

So anything using op2 from a deno_core containing crate will not work unless it also includes the exact same version of deno_core as a direct dependency as well. See her for an example:
https://github.com/rscarson/rustyscript/issues/30

This change more or less breaks rustyscript's support for op2

PR #305, opened 2023-11-05
fix(ops): partial revert of 16970b7 to fix issue #302

rscarson:
Very large looking change that effectively just replaces `::deno_core` with `deno_core` across `_dispatch` and `*.out` files

This is required for software using deno_core to compile if deno_core is being provided through a wrapper crate, such as rustyscript or js_sandbox

Below is an example of a sample that would not compile without this change:

```rust
use core_wrap::deno_core::{self, extension, op2};

#[op2]
#[string]
fn test_op2(#[string] _name: String) -> String {
    "test".to_string()
}
```

### DIFF
--- a/core/modules/tests.rs
+++ b/core/modules/tests.rs
@@ -21,9 +21,7 @@ use url::Url;
 use super::*;
 
 // deno_ops macros generate code assuming deno_core in scope.
-mod deno_core {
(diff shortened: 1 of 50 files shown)


mmastrac (approved):
LGTM!

PR #314, opened 2023-11-08
chore: trim deps and add dep checker step

mmastrac:


.github/workflows/ci.yml
```diff
@@ -52,6 +56,14 @@ jobs:
           cargo clippy --locked --release --all-features --all-targets -- -D clippy::all
           deno lint
 
+      - name: Lint deps
+        if: contains(matrix.os, 'ubuntu')
+        run: |
+          cargo install cargo-machete --locked
+          cargo machete
+          cargo install cargo-udeps --locked
+          cargo +nightly udeps
```
dsherret, line 65: Consider putting this after cargo test so that people get faster feedback on failing tests.

dsherret (approved):
Nice!

.github/workflows/ci.yml
mmastrac: Good call, swapped.

mmastrac: Actually, have to revert it because the tests cause a dep lint failure (because of compile tests :/)

PR #323, opened 2023-11-09
feat(core): speedier async stubs + template opAsync

mmastrac:
Adds a new template system for op async dispatch generation, and starts generating opAsync from it as well so we can play with performance more easily.

First pass at optimization: 5% improvement in async dispatch by avoiding allocation of promise IDs and generally doing less work in the async handlers.

Before:

```
test bench_op_async_void                  ... bench:      80,853 ns/iter (+/- 2,123)
```

After:

```
test bench_op_async_void                  ... bench:      76,799 ns/iter (+/- 1,158)
```

### DIFF
--- a/core/rebuild_async_stubs.js
+++ b/core/rebuild_async_stubs.js
@@ -6,7 +6,7 @@ const doNotModify =
 
 // The template function we build opAsync and op_async_N functions from
 function __TEMPLATE__(__ARGS_PARAM__) {
-  const id = (nextPromiseId + 1) & 0xffffffff;
+  const id = nextPromiseId;
   try {
     const maybeResult = __OP__(__ARGS__);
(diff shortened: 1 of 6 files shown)


core/rebuild_async_stubs.js
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env deno run --allow-read --allow-write
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
```
bartlomieju, line 2: Would be nice to add some tests for this template system

bartlomieju: Also if you'd prefer to do it in Rust, I'd recommend `tinytemplate` for nicer templating :)

bartlomieju (approved):
LGTM

PR #324, opened 2023-11-11
perf(core): try to fast path every realm case

mmastrac:


bartlomieju (approved):
LGTM, this makes sense

Issue #326, opened 2023-11-14
Reject unsupported import attributes


bartlomieju:
Ref https://github.com/denoland/deno/issues/20843

`deno_core` should accept a list of known attributes and default to what WPT recognizes if the setting is not provided.

bartlomieju:
Failing tests:
```

promise_test(async test => {
    const result = await import("./export-hello.js", { with: { } });
    assert_equals(result.default, "hello");
}, "Dynamic import with an empty with clause should succeed");

promise_test(async test => {
    return promise_rejects_js(test, TypeError,
        import("./export-hello.js", { with: { unsupportedAssertionKey: "unsupportedAssertionValue"} }),
        "Dynamic import with an unsupported import attribute should fail");
}, "Dynamic import with an unsupported import attribute should fail");

promise_test(test => {
    return promise_rejects_js(test, TypeError,
        import("./export-hello.js", { assert: { type: "notARealType"} } ),
        "Dynamic import with an unsupported type attribute should fail");
}, "Dynamic import with an unsupported type attribute should fail");

```

PR #327, opened 2023-11-14
feat: Import attributes validation callback

bartlomieju:
Closes https://github.com/denoland/deno_core/issues/326

This commit adds `RuntimeOptions.validate_import_attributes_cb` that can be
used to customize which import attributes are allowed.

### DIFF
--- a/core/modules/mod.rs
+++ b/core/modules/mod.rs
@@ -51,7 +51,8 @@ pub type ModuleName = FastString;
 
 /// Callback to validate import attributes. If the validation fails and exception
 /// should be thrown using `scope.throw_exception()`.
-pub type ValidateImportAttributesCb = Box<dyn Fn(&mut v8::HandleScope, &HashMap<String, String>)>;
+pub type ValidateImportAttributesCb =
+  Box<dyn Fn(&mut v8::HandleScope, &HashMap<String, String>)>;
 
 const SUPPORTED_TYPE_ASSERTIONS: &[&str] = &["json"];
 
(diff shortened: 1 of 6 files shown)


core/modules/mod.rs
```diff
@@ -49,10 +49,16 @@ pub(crate) type ModuleLoadId = i32;
...
+pub type ValidateImportAttributesCb =
+  Box<dyn Fn(&mut v8::HandleScope, &HashMap<String, String>)>;
+
 const SUPPORTED_TYPE_ASSERTIONS: &[&str] = &["json"];
 
-/// Throws V8 exception if assertions are invalid
-pub(crate) fn validate_import_assertions(
+/// Throws a `TypeError` if `type` attribute is not equal to "json". Allows
+/// all other attributes.
+pub(crate) fn validate_import_attributes(
```
dsherret, line 61: Should the default validator error when the key is not `type`?

core/runtime/jsruntime.rs
```diff
@@ -454,6 +456,10 @@ pub struct RuntimeOptions {
   /// An optional instance of `FeatureChecker`. If one is not provided, the
   /// default instance will be created that has no features enabled.
   pub feature_checker: Option<Arc<FeatureChecker>>,
+
+  /// A callback that can be used to validate import attributes received at
+  /// the import site.
```
dsherret, line 461: Maybe this should say something about how providing `None` will use the default validator and describe how the default validator works? Otherwise it kind of seems like providing `None` will cause import attributes to not be validated.

dsherret (approved):
LGTM

core/modules/mod.rs
bartlomieju: I'm not sure - @lucacasonato thoughts? I was thinking of doing that in the CLI.

core/runtime/jsruntime.rs
bartlomieju: Sure, I'll update that

PR #330, opened 2023-11-16
build(core): add feature flag include_icu_data

jirutka:
Embedding ICU data into Deno binary is undesirable when linking with system-provided ICU (in Linux distributions) - it adds 10 MiB of unnecessary bloat.

This patch is already used in deno package for Alpine Linux.


mmastrac (approved):
LGTM!

mmastrac:
@jirutka Linking this with https://github.com/denoland/deno/issues/21153 as we should automatically make this work in unvendored mode.

PR #331, opened 2023-11-17
feat: Add a re-entrancy flag for ops

mmastrac:
Ops now generate a re-entrancy guard in debug mode that checks to see if they are illegally re-entrant. Ideally we should limit op re-entrancy because it can cause issues with locks, v8 states, etc.

Tests pass in Deno locally with these changes, and the given ops marked as re-entrant.

Note that this check is compiled out in release mode for now.

### DIFF
--- a/core/ops_builtin_v8.rs
+++ b/core/ops_builtin_v8.rs
@@ -360,7 +360,8 @@ pub struct SerializeDeserializeOptions<'a> {
   for_storage: bool,
 }
 
-#[op2]
+// May be reentrant in the case of errors.
+#[op2(reentrant)]
 #[buffer]
 pub fn op_serialize(
   scope: &mut v8::HandleScope,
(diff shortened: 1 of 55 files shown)


core/ops_builtin_v8.rs
```diff
@@ -1,3 +1,4 @@
+use crate::OpState;
```
bartlomieju, line 1: Move under the copyright :) `rust-analyzer` could do a better job here

bartlomieju (approved):
LGTM with a nit

core/ops_builtin_v8.rs
mmastrac: Ugh we need the copyright checker here.

aapoalas (approved):
Looks good to me aside for that one copyright/import issue :)

PR #245, opened 2023-09-29
Adds `clear_modules` to JsRealm and JsRuntime to allow the unloading of modules.

rscarson:
Added `clear_modules` to JsRealm and JsRuntime to allow the unloading of modules.

This change does not affect loaded extensions.

The use case is a very quick reset for an active runtime that does not require reconstructing the runtime, or reloading extensions/ops

The code below was used to check for memory not being cleared from v8 resources:
- Performs 1000 iterations of
    - Load 1000 modules, each storing 1024bytes of data.
    - Clear the modules
```rust
#[cfg(test)]
mod test_module_unload {
  fn get_numbered_module(
    i: usize,
  ) -> (crate::url::Url, Option<crate::FastString>) {
    let name = crate::resolve_path(
      &format!("module{i}.js"),
      &std::env::current_dir()
        .expect("Unable to get current working directory"),
    )
    .expect("Unable to get module url");

    let code = Some(crate::FastString::from(
      "
      let value = `
      Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla mollis facilisis ex a pretium. Pellentesque pharetra mollis ipsum, laoreet dapibus turpis. Ut volutpat hendrerit justo ac hendrerit. Vivamus id ligula nisl. Proin pretium tincidunt hendrerit. Vivamus ultrices magna neque, id elementum nisl luctus id. Sed finibus, arcu vitae porta ornare, felis nibh convallis neque, eu auctor risus diam ut neque. Sed euismod nisl ut nibh tincidunt interdum. Mauris aliquet mauris odio, vitae eleifend ante imperdiet at. Phasellus dictum sodales mi sit amet fringilla. Aliquam erat volutpat. Ut eget neque eu lorem aliquet molestie. In a purus sit amet velit dapibus semper a ac quam. Ut hendrerit aliquam tempus. Maecenas justo ipsum, venenatis id ipsum nec, euismod rutrum ante. Sed et urna porta dolor dictum pulvinar ut eu enim.
      Vestibulum justo nisl, lacinia consequat quam ut, tristique dapibus arcu. Mauris rhoncus, lectus nec mattis pharetra, lectus neque placerat dolor, vel commodo nunc ligula sit amet justo. Nunc mi.
      `;
      ".to_string(),
    ));

    (name, code)
  }

  #[test]
  fn test_unload() {
    let mut js_runtime = crate::JsRuntime::new(Default::default());

    for iteration in 0..1000 {
      let runtime = &mut js_runtime;
[...]

PR #335, opened 2023-11-20
refactor(core): refactoring ModuleMap and breaking down locks

mmastrac:
Moves most of the module-related code from JsRealm to ModuleMap. This also removes the top-level RefCell and adds lower-level ones.

This paves the way for future refactoring.

Performance is not a consideration for this refactoring, but it's approximately on par with performance before.


core/runtime/tests/jsrealm.rs
```diff
@@ -732,101 +732,6 @@ async fn test_realms_concurrent_module_evaluations() {
   assert!(matches!(other_realm_receiver.await, Ok(Ok(()))));
 }
 
-// Make sure that loading and evaluating dynamic imported modules in different
```
mmastrac, line 735: Deleting any multi-realm tests that get in the way for now.

core/modules/tests.rs
```diff
@@ -691,32 +686,6 @@ async fn dyn_import_ok() {
   .await;
 }
 
-#[tokio::test]
```
mmastrac, line 694: Was never exported and should not be used regardless.

core/runtime/jsruntime.rs
```diff
@@ -305,9 +305,6 @@ pub type CompiledWasmModuleStore = CrossIsolateStore<v8::CompiledWasmModule>;
 pub struct JsRuntimeState {
   known_realms: Vec<JsRealmInner>,
   pub(crate) has_tick_scheduled: bool,
-  /// A counter used to delay our dynamic import deadlock detection by one spin
```
mmastrac, line 308: This lives in ModuleMap for now, but it will likely move around a bit as we refactor.

core/runtime/jsrealm.rs
```diff
@@ -71,7 +62,6 @@ pub(crate) struct ContextState {
   pub(crate) js_wasm_streaming_cb: Option<Rc<v8::Global<v8::Function>>>,
   pub(crate) pending_promise_rejections:
     VecDeque<(v8::Global<v8::Promise>, v8::Global<v8::Value>)>,
-  pending_dyn_mod_evaluate: Vec<DynImportModEvaluate>,
   pub(crate) pending_mod_evaluate: Option<ModEvaluate>,
```
mmastrac, line 65: This will be moved in a follow-up PR.

core/runtime/jsruntime.rs
```diff
@@ -1533,20 +1521,23 @@ impl JsRuntime {
...
-          assert!(poll_imports.is_ready());
-
-          if main_realm.evaluate_dyn_imports(&mut self.inner.v8_isolate) {
-            has_evaluated = true;
-          } else {
-            break;
+        {
+          let main_realm = self.inner.main_realm.as_ref().unwrap();
+          let mut scope = main_realm.handle_scope(&mut self.inner.v8_isolate);
+          let modules = main_realm.0.module_map();
```
mmastrac, line 1527: We will clean this up in a future PR. Right now we just moved methods and tweaked the scope a bit.

core/modules/mod.rs
```diff
@@ -725,7 +708,6 @@ pub(crate) struct ModuleRequest {
 pub(crate) struct ModuleInfo {
   #[allow(unused)]
   pub id: ModuleId,
-  // Used in "bindings.rs" for "import.meta.main" property value.
```
bartlomieju, line 728: Why remove this comment?

core/modules/tests.rs
bartlomieju: Can you elaborate how it was not safe?

core/runtime/mod.rs
```diff
@@ -1,5 +1,5 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
-mod bindings;
+pub(crate) mod bindings;
```
bartlomieju, line 2: Why does it need to public for a crate now?

core/runtime/snapshot_util.rs
```diff
@@ -187,7 +187,7 @@ pub(crate) fn get_snapshotted_data(
 
   let next_module_id = {
     let info_data: v8::Local<v8::Array> =
-      val.get_index(&mut scope, 1).unwrap().try_into().unwrap();
+      val.get_index(&mut scope, 2).unwrap().try_into().unwrap();
```
bartlomieju, line 190: Can you add a comment why it's using index 2?

core/modules/map.rs
```diff
@@ -55,46 +61,79 @@ pub(crate) enum SymbolicModule {
...
-    HashMap<ModuleLoadId, v8::Global<v8::PromiseResolver>>,
-  pub(crate) preparing_dynamic_imports:
-    FuturesUnordered<Pin<Box<PrepareLoadFuture>>>,
-  pub(crate) pending_dynamic_imports:
-    FuturesUnordered<StreamFuture<RecursiveModuleLoad>>,
-
-  // This store is used temporarily, to forward parsed JSON
-  // value from `new_json_module` to `json_module_evaluation_steps`
+  // TODO(mmastrac): we should not be swapping this loader out
+  pub(crate) loader: RefCell<Rc<dyn ModuleLoader>>,
```
bartlomieju, line 75: I don't understand this change... Why is it `RefCell<Rc<_>>`? This seems very fishy

core/modules/map.rs
```diff
@@ -55,46 +61,79 @@ pub(crate) enum SymbolicModule {
...
+    let id = data.next_load_id;
+    data.next_load_id += 1;
+    id + 1
+  }
+
+  fn collect_modules(
     &self,
-  ) -> Vec<(AssertedModuleType, &ModuleName, &SymbolicModule)> {
-    let mut output = vec![];
+    mut f: impl FnMut(usize, AssertedModuleType, &ModuleName, &SymbolicModule),
```
bartlomieju, line 123: Can you add a comment about this param? It's not obvious what it's meant to do. Maybe this method should be renamed from `collect_modules()` if it doesn't return anything?

core/modules/map.rs
```diff
@@ -55,46 +61,79 @@ pub(crate) enum SymbolicModule {
...
+  // TODO(mmastrac): we should not be swapping this loader out
+  pub(crate) loader: RefCell<Rc<dyn ModuleLoader>>,
+
+  dynamic_import_map:
+    RefCell<HashMap<ModuleLoadId, v8::Global<v8::PromiseResolver>>>,
+  preparing_dynamic_imports:
+    RefCell<FuturesUnordered<Pin<Box<PrepareLoadFuture>>>>,
+  pending_dynamic_imports:
+    RefCell<FuturesUnordered<StreamFuture<RecursiveModuleLoad>>>,
+  pending_dyn_mod_evaluate: RefCell<Vec<DynImportModEvaluate>>,
```
bartlomieju, line 83: `pending_dyn_mod_evaluations`?

core/modules/mod.rs
mmastrac: It's no longer used there... we use the main_module_id property in ModuleMap now.

core/runtime/mod.rs
mmastrac: There's a utility function in `bindings` required for ModuleMap 😢 

```
pub fn create_empty_fn<'s>(
  scope: &mut v8::HandleScope<'s>,
) -> Option<v8::Local<'s, v8::Function>> {
  let empty_fn = v8::FunctionTemplate::new(scope, empty_fn);
  empty_fn.get_function(scope)
}
```

mmastrac: I could probably just duplicate it as it's not really that complex

core/modules/map.rs
mmastrac: We need to swap out loader during extension loads because of how things are currently done. It's definitely fishy and I'd like to fix that at some point.

Originally we just used the outer RefCell to swap it out, but that's gone.

core/modules/tests.rs
mmastrac: It just cleared the module map without clearing the pending modules or any other important stuff. The comments stated that it wasn't safe, and it isn't actually used anywhere so I'm just removing a potential footgun.

The API was only ever tested, so I'm not sure what it was intended for, but we definitely shouldn't be doing anything like this.

core/runtime/snapshot_util.rs
mmastrac: (this is one of the reasons I dislike all this manual serialization)

core/runtime/mod.rs
mmastrac: Actually, this requires an external so it does need to be pub(crate) for now. I can clean this up in a later refactoring but for now it's too painful.

core/modules/mod.rs
bartlomieju: Okay, so why keep this field?

core/runtime/mod.rs
bartlomieju: I agree, seems better to duplicate that export that module.

core/modules/map.rs
bartlomieju: I see. That makes sense. Could you update a comment and remove `pub(crate)` and then add `get_loader()` (public) and `set_loader()` that only `deno_core` can use?

bartlomieju (approved):

core/modules/mod.rs
mmastrac: We use it as part of tests 😢

core/modules/tests.rs
aapoalas: https://github.com/denoland/deno_core/pull/245

This was a very recent community contribution.

bartlomieju: I completely forgot about this. IIRC they had a use case for testing and were aware of the footgun-like nature of this API. I think we should keep it after all.

PR #336, opened 2023-11-21
refactor(core): stripping out non-main realms

mmastrac:
We no longer support multiple realms (aka contexts) in a runtime. 

Allows us to use a single scope for the whole polling operation. This yields a 1% bump on http tests in deno.

Follow-up work will move much of the jsruntime polling code into the realm.


core/runtime/jsruntime.rs
```diff
@@ -232,15 +233,10 @@ pub struct JsRuntime {
   pub(crate) inner: InnerIsolateState,
   pub(crate) allocations: IsolateAllocations,
   extensions: Vec<Extension>,
-  preserve_snapshotted_modules: Option<&'static [&'static str]>,
   event_loop_middlewares: Vec<EventLoopMiddlewareFn>,
-  global_template_middlewares: Vec<GlobalTemplateMiddlewareFn>,
-  global_object_middlewares: Vec<GlobalObjectMiddlewareFn>,
```
bartlomieju, line 238: Why are these fields gone?

core/runtime/jsruntime.rs
```diff
@@ -853,134 +816,52 @@ impl JsRuntime {
...
   #[inline]
   pub fn handle_scope(&mut self) -> v8::HandleScope {
-    self.main_realm().handle_scope(self.v8_isolate())
+    let isolate = &mut self.inner.v8_isolate;
+    self.inner.main_realm.handle_scope(isolate)
+  }
+
+  #[inline(always)]
+  /// Create a scope on the stack with the given context
+  fn with_runtime_scope<'s, T>(
```
bartlomieju, line 841: Shouldn't this be called `with_context_scope` since you are creating `v8::ContextScope`?

core/runtime/jsruntime.rs
```diff
@@ -1410,45 +1307,60 @@ impl JsRuntime {
...
     cx: &mut Context,
+    scope: &mut v8::HandleScope,
+    global: &v8::Global<v8::Value>,
   ) -> Poll<Result<v8::Global<v8::Value>, Error>> {
+    let promise = v8::Local::new(scope, global);
+
+    // TODO(mmastrac): this would be better if we just type-tested at the top level
+    let Ok(promise) = v8::Local::<v8::Promise>::try_from(promise) else {
+      return Poll::Ready(Ok(global.clone()));
+    };
```
bartlomieju, line 1331: I'd be in favor of that, makes sense to panic here if it's not a `v8::Promise`.

core/runtime/jsruntime.rs
mmastrac: They were no longer used when I removed multiple realms. They are easy enough to add back if needed in a later pass, but I didn't want to add `allow(unused)`.

core/runtime/jsruntime.rs
mmastrac: For sure... I want to leave it as a TODO because we don't have a way to cast/type test `v8::Global<...>` in rusty_v8 like we do with `v8::Local<...>` uet. I suspect it's a bit of a hole in our rusty_v8 API.

core/runtime/jsruntime.rs
mmastrac: Good call, I'll rename it.

core/runtime/jsruntime.rs
bartlomieju: Surprising - I thought they were used in `ext/node` in Deno repository. Maybe they are just one-off now and don't need to be stored anymore 👍

bartlomieju (approved):

PR #340, opened 2023-11-22
feat(core): add support for other assertion types to ModuleMap

mmastrac:
`ModuleMap` now supports any type of module assertion (including arbitrary strings), however this has not been threaded through beyond this code.

### DIFF
--- a/core/modules/map.rs
+++ b/core/modules/map.rs
@@ -823,14 +823,9 @@ impl ModuleMap {
     specifier: impl AsRef<str>,
     asserted_module_type: impl AsRef<AssertedModuleType>,
   ) -> bool {
-    if let Some(id) =
-      self.get_id(specifier.as_ref(), asserted_module_type.as_ref())
-    {
-      return asserted_module_type.as_ref()
-        == &self.data.borrow().info.get(id).unwrap().module_type;
-    }
-
-    false
+    self
+      .get_id(specifier.as_ref(), asserted_module_type.as_ref())
+      .is_some()
   }
 
   pub(crate) fn alias(
--- a/core/modules/map.rs
+++ b/core/modules/map.rs
@@ -88,6 +88,95 @@ pub(crate) struct ModuleMap {
   pub(crate) dyn_module_evaluate_idle_counter: Cell<u32>,
 }
 
+/// Map of [`ModuleName`] and [`AssertedModuleType`] to a data field.
+struct ModuleNameTypeMap<T> {
+  submaps: Vec<HashMap<ModuleName, T>>,
+  map_index: HashMap<AssertedModuleType, usize>,
+  len: usize,
+}
+
+impl<T> Default for ModuleNameTypeMap<T> {
+  fn default() -> Self {
+    Self {
+      submaps: Default::default(),
+      map_index: Default::default(),
+      len: 0,
+    }
+  }
+}
+
+impl<T> ModuleNameTypeMap<T> {
+  pub fn len(&self) -> usize {
+    self.len
+  }
+
+  fn map_index(&self, ty: &AssertedModuleType) -> Option<usize> {
+    self.map_index.get(ty).copied()
+  }
+
+  pub fn get<Q>(&self, ty: &AssertedModuleType, name: &Q) -> Option<&T>
+  where
+    ModuleName: std::borrow::Borrow<Q>,
+    Q: std::cmp::Eq + std::hash::Hash + ?Sized,
+  {
+    let Some(index) = self.map_index(ty) else {
+      return None;
+    };
+    let Some(map) = self.submaps.get(index) else {
+      return None;
+    };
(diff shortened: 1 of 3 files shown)


core/modules/map.rs
```diff
@@ -88,6 +88,95 @@ pub(crate) struct ModuleMap {
   pub(crate) dyn_module_evaluate_idle_counter: Cell<u32>,
 }
 
+/// Map of [`ModuleName`] and [`AssertedModuleType`] to a data field.
+struct ModuleNameTypeMap<T> {
```
mmastrac, line 92: We use a vec of maps rather than a HashMap with multiple keys to simplify lookups here.

core/modules/map.rs
```diff
@@ -901,7 +925,7 @@ impl ModuleMap {
     let load = RecursiveModuleLoad::dynamic_import(
       specifier,
       referrer,
-      asserted_module_type,
+      asserted_module_type.clone(),
```
mmastrac, line 928: All of these clones will be cheap, except if using a non-standard assertion type.

core/modules/mod.rs
```diff
@@ -594,43 +599,32 @@ impl Stream for RecursiveModuleLoad {
           Ok(url) => url,
           Err(error) => return Poll::Ready(Some(Err(error))),
         };
-        let load_fut = if let Some(_module_id) = inner.root_module_id {
```
mmastrac, line 597: Cleaned up this giant TODO here.

aapoalas:
**nitpick:** Shouldn't this be `import attributes`? Or is this still about `import "foo" assert { type: "foo" }`?

core/modules/map.rs
```diff
@@ -770,50 +821,25 @@ impl ModuleMap {
...
-    asserted_module_type: AssertedModuleType,
-    f: impl FnOnce(&mut HashMap<ModuleName, SymbolicModule>) -> T,
-  ) -> T {
-    match asserted_module_type {
-      AssertedModuleType::Json => f(&mut self.data.borrow_mut().by_name_json),
-      AssertedModuleType::JavaScriptOrWasm => {
-        f(&mut self.data.borrow_mut().by_name_js)
-      }
-    }
+    self
```
mmastrac, line 826: This was overtesting -- if it's in `ModuleInfo` with the given ID, the type is correct.

mmastrac:
@aapoalas In this case it's just allowing for other assertions -- we don't currently allow import assertions with parameters but this will open the door (slightly) for import assertions of other types.

aapoalas:
Just as long as we're not adding more features to the deprecated `assert { type: foo }` syntax only expanding the possibilities of the new/fixed `with { type?: foo, ... }` syntax.

mmastrac:
@aapoalas Likely this is still a bit of a ways off full support for `with` but this gets us a bit closer.

core/modules/map.rs
bartlomieju: Do we have a unit test for that? If not could you add one?

core/modules/map.rs
bartlomieju: What do you mean by non-standard assertion type?

core/modules/mod.rs
bartlomieju: Very nice 👍

core/modules/mod.rs
```diff
@@ -669,11 +663,64 @@ impl Stream for RecursiveModuleLoad {
...
+    scope: &mut v8::HandleScope<'s>,
+  ) -> v8::Local<'s, v8::Value> {
+    match self {
+      AssertedModuleType::JavaScriptOrWasm => v8::Integer::new(scope, 0).into(),
+      AssertedModuleType::Json => v8::Integer::new(scope, 1).into(),
+      AssertedModuleType::Other(ty) => {
+        v8::String::new(scope, ty).unwrap().into()
+      }
+    }
+  }
```
bartlomieju, line 691: Where do we even use this method? Ideally we could remove it all together (in another PR)

bartlomieju (approved):

core/modules/map.rs
mmastrac: Anything other than JSON or JavaScriptOrWasm. Those two (the only ones we actually use right now) will be cheap clones.

core/modules/mod.rs
mmastrac: It's in the snapshot serialization -- if we move to serde_v8 we can get rid of this at that point. :)

core/modules/map.rs
mmastrac: Unless I'm mistaken, all the tests technically cover this functionality because it's just not possible for the type not to match. The modules of different types are all independent, so the module returned by `get_id` _must_ have the same module type. I suspect this code was leftover from when I split the maps up ages ago and just wasn't cleaned up.

Issue #338, opened 2023-11-22
Add ability to call function with args on `JsRuntime`


yorinasub17:
Hello!

I'm working on a tool that embeds Deno as a runtime (after following [this blog post](https://deno.com/blog/roll-your-own-javascript-runtime)). I'm working on a feature that allows users to pass context arguments to a function they are writing, and found that there isn't an easy way to call a function with args (unless I am missing something). Is there any interest in adding a `call_with_args_and_await` method to `JsRuntime` that is basically `call_and_await`, but takes args?

E.g., https://github.com/yorinasub17/deno_core/commit/1675219c0e04d0ba2310aa8554c14bc4ca2f6630 (which I am happy to open as a PR if it is of interest)

Thanks for reading and considering this!

bartlomieju:
Hey, sounds like a good idea and a PR would be more than welcome!

yorinasub17:
Great! Took a stab at this here: https://github.com/denoland/deno_core/pull/341

PR #341, opened 2023-11-23
feat(core): add function to call a JS function with args

yorinasub17:
Closes https://github.com/denoland/deno_core/issues/338

This implements a new function `call_with_args_and_await` on `JsRuntime` that allows users to call a JS function from the Rust runtime with args (unlike `call_and_await`).

### DIFF
--- a/core/runtime/jsruntime.rs
+++ b/core/runtime/jsruntime.rs
@@ -1208,14 +1208,18 @@ impl JsRuntime {
   ) -> Result<v8::Global<v8::Value>, Error> {
     let promise = {
       let scope = &mut self.handle_scope();
-      let mut local_args: SmallVec<[v8::Local<v8::Value>; 32]> =
-        SmallVec::with_capacity(args.len());
-      for v in args {
-        local_args.push(v8::Local::new(scope, v));
-      }
       let cb = function.open(scope);
       let this = v8::undefined(scope).into();
-      let promise = cb.call(scope, this, &local_args);
+      let promise = if args.is_empty() {
+        cb.call(scope, this, &[])
+      } else {
+        let mut local_args: SmallVec<[v8::Local<v8::Value>; 32]> =
+          SmallVec::with_capacity(args.len());
+        for v in args {
+          local_args.push(v8::Local::new(scope, v));
+        }
+        cb.call(scope, this, &local_args)
+      };
       if promise.is_none() || scope.is_execution_terminating() {
         let undefined = v8::undefined(scope).into();
         return exception_to_err_result(scope, undefined, false);
--- a/core/runtime/jsruntime.rs
+++ b/core/runtime/jsruntime.rs
@@ -1195,11 +1195,27 @@ impl JsRuntime {
     &mut self,
     function: &v8::Global<v8::Function>,
   ) -> Result<v8::Global<v8::Value>, Error> {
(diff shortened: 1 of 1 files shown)


core/runtime/jsruntime.rs
```diff
@@ -1304,12 +1304,28 @@ impl JsRuntime {
...
+    function: &v8::Global<v8::Function>,
+    args: &[v8::Global<v8::Value>],
   ) -> Result<v8::Global<v8::Value>, Error> {
     let promise = {
       let scope = &mut self.handle_scope();
+      let mut local_args: SmallVec<[v8::Local<v8::Value>; 32]> =
+        SmallVec::with_capacity(args.len());
+      for v in args {
+        local_args.push(v8::Local::new(scope, v));
+      }
```
yorinasub17, line 1325: Please let me know if there is a better way to do this! I'm new to Rust and I copied this from similar code in the same file (choosing `32` arbitrarily), so not sure if this looks right.

core/runtime/jsruntime.rs
```diff
@@ -1304,12 +1304,28 @@ impl JsRuntime {
...
       let cb = function.open(scope);
       let this = v8::undefined(scope).into();
-      let promise = cb.call(scope, this, &[]);
+      let promise = cb.call(scope, this, &local_args);
```
mmastrac, line 1328: I think you can change this to avoid the `SmallVec` when args is empty. It would look something like this:

```
let cb = function.open(scope);
let this = v8::undefined(scope).into();
let promise = if args.is_empty() {
      let mut local_args: SmallVec<[v8::Local<v8::Value>; 32]> =
        SmallVec::with_capacity(args.len());
      for v in args {
        local_args.push(v8::Local::new(scope, v));
      }
      cb.call(scope, this, &local_args)
} else {
  cb.call(scope, this, &[]);
}
```

yorinasub17: Makes sense! Updated in https://github.com/denoland/deno_core/pull/341/commits/6d48ae327413faf7d3dc9f0ba6c56023c39a3349

bartlomieju: 32 seems excessive here - this is copied from "op code" that expects there's gonna be up to 16 ops returned in one go. In reality I don't expect many situation where embedders would pass more than 8 args to a function.

core/runtime/jsruntime.rs
```diff
@@ -1194,12 +1194,32 @@ impl JsRuntime {
...
   ) -> Result<v8::Global<v8::Value>, Error> {
     let promise = {
       let scope = &mut self.handle_scope();
       let cb = function.open(scope);
       let this = v8::undefined(scope).into();
-      let promise = cb.call(scope, this, &[]);
+      let promise = if args.is_empty() {
+        cb.call(scope, this, &[])
+      } else {
+        let mut local_args: SmallVec<[v8::Local<v8::Value>; 32]> =
```
mmastrac, line 1216:
```suggestion
        let mut local_args: SmallVec<[v8::Local<v8::Value>; 8]> =
```

mmastrac (approved):
LGTM!

core/runtime/jsruntime.rs
yorinasub17: Makes sense. Thanks for the explanation!

PR #342, opened 2023-11-23
refactor(core): remove cloning in event loop poll

mmastrac:


core/runtime/jsruntime.rs
```diff
@@ -83,6 +83,7 @@ pub(crate) struct IsolateAllocations {
 pub(crate) struct ManuallyDropRc<T>(ManuallyDrop<Rc<T>>);
 
 impl<T> ManuallyDropRc<T> {
+  #[allow(unused)]
```
mmastrac, line 86: Keeping to avoid future footgun.

bartlomieju: What's the future footgun here? :D

bartlomieju (approved):
LGTM

core/runtime/jsruntime.rs
mmastrac: `ManuallyDrop<Rc<T>>` can be cloned..  but when you drop the clone it doesn't release the refcount 😢

mmastrac: ... and because ManuallyDrop<T> derefs to T, that means that a "clone" method on T is called, even if that clone method is on the Rc.

PR #343, opened 2023-11-23
refactor(core): remove one lock by moving next_tick to context

mmastrac:


core/ops_builtin_v8.rs
```diff
@@ -74,14 +74,18 @@ pub fn op_run_microtasks(isolate: *mut v8::Isolate) {
   };
 }
 
-#[op2(fast)]
-pub fn op_has_tick_scheduled(state: &JsRuntimeState) -> bool {
-  state.has_tick_scheduled
+#[op2]
```
mmastrac, line 77: These become slow for now but it should not matter a huge amount.

bartlomieju (approved):

PR #345, opened 2023-11-23
refactor: removing locks on JsRuntimeState

mmastrac:
Removes most locks from the poll loop. We no longer need to lock JsRuntimeState at all, as the inner mutable items are moved into RefCell or Cell.

The waker that was previously accessed via OpState is cloned JsRuntimeState, and will likely move to ContextState at a later time.


bartlomieju (approved):
LGTM, nice cleanup!

PR #347, opened 2023-11-23
refactor(core): split module map data and recursive load out

mmastrac:
Splitting `RecursiveModuleLoad` and `ModuleMapData` out.


bartlomieju (approved):
Looks like this is just moving the code around so I won't be nitpicking on anything here, LGTM!

PR #350, opened 2023-11-24
feat: V8TaskSpawner

mmastrac:
Provides an `OpState`-stored (for now) `V8TaskSpawner` and `V8CrossThreadTaskSpawner` that allow for submission of v8 tasks to a queue that is run alongside ops.

```
let task_spawner: &V8TaskSpawn = op_state.borrow();
task_spawner.spawn(|scope| {
  /* use scope */
});
```

This allows you submit jobs into a queue that's run just before op dispatch (timing TBD and subject to change). The task is given a scope for the current context, and may do whatever it wishes.

This will likely turn into microtask submission API in the future.

### DIFF
--- a/core/runtime/jsruntime.rs
+++ b/core/runtime/jsruntime.rs
@@ -623,6 +623,21 @@ impl JsRuntime {
 
     let weak = Rc::downgrade(&state_rc);
     let context_state = Rc::new(RefCell::new(ContextState::default()));
+
+    // Add the task spawners to the OpState
+    let spawner = context_state
+      .borrow()
+      .task_spawner_factory
+      .clone()
+      .new_same_thread_spawner();
+    op_state.borrow_mut().put(spawner);
+    let spawner = context_state
+      .borrow()
+      .task_spawner_factory
+      .clone()
+      .new_cross_thread_spawner();
+    op_state.borrow_mut().put(spawner);
+
     let count = ops.len();
     let mut op_ctxs = ops
       .into_iter()
@@ -1772,12 +1787,14 @@ impl EventLoopPendingState {
   ) -> Self {
     let num_unrefed_ops = state.unrefed_ops.len();
     let num_pending_ops = state.pending_ops.len();
+    let has_pending_tasks = state.task_spawner_factory.has_pending_tasks();
     let has_pending_dyn_imports = modules.has_pending_dynamic_imports();
     let has_pending_dyn_module_evaluation =
       modules.has_pending_dyn_module_evaluation();
     let has_pending_module_evaluation = state.pending_mod_evaluate.is_some();
     EventLoopPendingState {
-      has_pending_refed_ops: num_pending_ops > num_unrefed_ops,
+      has_pending_refed_ops: has_pending_tasks
+        || num_pending_ops > num_unrefed_ops,
       has_pending_dyn_imports,
       has_pending_dyn_module_evaluation,
       has_pending_module_evaluation,
@@ -1961,8 +1978,20 @@ impl JsRuntime {
     scope: &mut v8::HandleScope,
     context_state: &RefCell<ContextState>,
   ) -> Result<bool, Error> {
-    let mut context_state = context_state.borrow_mut();
     let mut dispatched_ops = false;
+
+    // Poll any pending task spawner tasks. Note that we need to poll separately because otherwise
+    // Rust will extend the lifetime of the borrow longer than we expect.
+    let tasks = context_state.borrow().task_spawner_factory.poll_inner(cx);
+    if let Poll::Ready(tasks) = tasks {
+      // TODO(mmastrac): we are using this flag
+      dispatched_ops = true;
+      for task in tasks {
+        task(scope);
+      }
+    }
+
+    let mut context_state = context_state.borrow_mut();
 
     // We return async responses to JS in unbounded batches (may change),
     // each batch is a flat vector of tuples:
--- a/core/tasks.rs
+++ b/core/tasks.rs
@@ -16,6 +16,11 @@ static_assertions::assert_not_impl_any!(V8TaskSpawnerFactory: Send);
 static_assertions::assert_not_impl_any!(V8TaskSpawner: Send);
 static_assertions::assert_impl_all!(V8CrossThreadTaskSpawner: Send);
 
+/// The [`V8TaskSpawnerFactory`] must be created on the same thread as the thread that runs tasks.
+///
+/// This factory is not [`Send`] because it may contain `!Send` tasks submitted by a
+/// [`V8CrossThreadTaskSpawner`]. It is only safe to send this object to another thread if you plan on
+/// submitting [`Send`] tasks to it, which is what [`V8CrossThreadTaskSpawner`] does.
 #[derive(Default)]
 pub(crate) struct V8TaskSpawnerFactory {
   // TODO(mmastrac): ideally we wouldn't box if we could use arena allocation and a max submission size
@@ -26,9 +31,7 @@ pub(crate) struct V8TaskSpawnerFactory {
   has_tasks: AtomicBool,
   /// The polled waker, woken on task submission.
   waker: AtomicWaker,
-  /// This cannot be Send because it may contain `!Send` tasks submitted by a [`V8TaskSpawner`]. It is
-  /// only safe to send this object to another thread if you plan on submitting [`Send`] tasks to it,
-  /// which is what [`V8CrossThreadTaskSpawner`] does.
+  /// Mark as `!Send`. See note above.
   _unsend_marker: PhantomData<*const ()>,
 }
 
@@ -44,6 +47,10 @@ impl V8TaskSpawnerFactory {
     V8CrossThreadTaskSpawner { tasks: self }
   }
 
+  /// `false` guarantees that there are no queued tasks, while `true` means that it is likely (but not guaranteed)
+  /// that tasks exist.
+  ///
+  /// Calls should prefer using the waker, but this is left while we rework the event loop.
   pub fn has_pending_tasks(&self) -> bool {
     self.has_tasks.load(std::sync::atomic::Ordering::SeqCst)
   }
@@ -54,7 +61,7 @@ impl V8TaskSpawnerFactory {
     // Check the flag first -- if it's false we definitely have no tasks
     if !self
       .has_tasks
-      .swap(false, std::sync::atomic::Ordering::SeqCst)
+      .swap(false, std::sync::atomic::Ordering::Acquire)
     {
       self.waker.register(cx.waker());
       return Poll::Pending;
@@ -78,10 +85,10 @@ impl V8TaskSpawnerFactory {
 
   fn spawn(&self, task: SendTask) {
     self.tasks.lock().unwrap().push(task);
-    // TODO(mmastrac): can we use a looser ordering here?
+    // TODO(mmastrac): can we skip the mutex here?
     self
       .has_tasks
-      .store(true, std::sync::atomic::Ordering::SeqCst);
+      .store(true, std::sync::atomic::Ordering::Release);
     self.waker.wake();
   }
 }
@@ -103,7 +110,7 @@ impl V8TaskSpawner {
   /// The task is handed off to be run the next time the event loop is polled, and there are
   /// no guarantees as to when this may happen.
   ///
-  /// # Important Notes
+  /// # Safety
   ///
   /// The task shares the same [`v8::HandleScope`] as the core event loop, which means that it
   /// must maintain the scope in a valid state to avoid corrupting or destroying the runtime.
@@ -169,7 +176,7 @@ impl V8CrossThreadTaskSpawner {
   /// no guarantees as to when this may happen, however the function will not return until the
   /// task has been fully run to completion.
   ///
-  /// # Important Notes
+  /// # Safety
   ///
   /// The task shares the same [`v8::HandleScope`] as the core event loop, which means that it
   /// must maintain the scope in a valid state to avoid corrupting or destroying the runtime.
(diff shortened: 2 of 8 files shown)


core/runtime/jsruntime.rs
```diff
@@ -1964,6 +1981,18 @@ impl JsRuntime {
...
 
+    // Poll any pending task spawner tasks
+    // TODO(mmastrac): we are polling this here because we already have a lock on context_state
+    if let Poll::Ready(tasks) =
+      context_state.task_spawner_factory.poll_inner(cx)
+    {
+      // TODO(mmastrac): we are using this flag
+      dispatched_ops = true;
+      for task in tasks {
+        task(scope);
```
aapoalas, line 1992: From V8's API documentation for `ThrowException`:

> When an exception has been scheduled it is illegal to invoke any JavaScript operation; the caller must return immediately and only after the exception has been handled does it become legal to invoke JavaScript operations.

I believe that this also extends to using scopes in C++ / Rust. Effectively, since now this `task` can trigger any extension provided Rust code that can then do anything it wants with the scope we'll need to prepare for the eventuality that some code being called will throw an error.

If a JS error thrown in a task breaks the event loop then we're in a bad state.

core/tasks.rs
```diff
@@ -0,0 +1,98 @@
...
+    let task: Box<dyn FnOnce(&mut v8::HandleScope<'_>)> = Box::new(f);
+    let task: Box<dyn FnOnce(&mut v8::HandleScope<'_>) + Send> =
+      unsafe { std::mem::transmute(task) };
+    self.tasks.spawn(task)
+  }
+}
+
+/// Allows for submission of v8 tasks on any thread.
+#[derive(Clone)]
+pub struct V8CrossThreadTaskSpawner {
```
aapoalas, line 87: After thinking on this a bit, I don't think FFI callbacks can use the same-thread `V8TaskSpawner` at all: Same thread callbacks are always necessarily responses to some JS calls going into native code and that native code then synchronously calling an FFI callback C function pointer and expecting a synchronous return value. So that just needs to be a sort of manually handled call with a `CallbackScope` etc.

I wonder if the same-thread task spawner even makes much sense. From a callback perspective at least it doesn't seem to make much sense as it cannot be used to do synchronous callbacks which is the only thing you want to do when on the same thread.

aapoalas (changes requested):
This is going to need a fair bit of tests to make sure that misbehaving tasks cannot break the event loop, I think.

mmastrac:
> This is going to need a fair bit of tests to make sure that misbehaving tasks cannot break the event loop, I think.

I may need to be clearer about this as it's designed to replace the middleware system which is already a bit of a wild west. Effectively we need a way to run code _in_ the event loop right now that is free.

core/tasks.rs
mmastrac: There's a few cases where the same-thread spawner works, but it's mainly for newer functionality like @littledivy's WASM rewrite where we need to call the `finish` method from a tokio task that doesn't have a scope associated with it.

aapoalas:
> > This is going to need a fair bit of tests to make sure that misbehaving tasks cannot break the event loop, I think.
> 
> I may need to be clearer about this as it's designed to replace the middleware system which is already a bit of a wild west. Effectively we need a way to run code _in_ the event loop right now that is free.

Yeah, I understand. Just, since we're reusing the same scope it means that user code can start affecting the event loop code's validity.

core/runtime/jsruntime.rs
mmastrac: Interesting. How do we currently handle this in middleware? We may just need to copy the same requirements over to this task spawner.

aapoalas: At least FFI middleware creates its own CallbackScope and then a HandleScope from that. If it throws it only poisons itself.

core/tasks.rs
aapoalas: That sounds like a damn nice use case then! Really nice <3

mmastrac: Thinking on it more, I suspect the same-thread task spawner wouldn't be useful for FFI as it would be better to just go right into the callback function, but it does make me wonder if we should consider a specific "v8 callback" API in the future.

core/runtime/jsruntime.rs
```diff
@@ -1961,9 +1978,21 @@ impl JsRuntime {
...
     context_state: &RefCell<ContextState>,
   ) -> Result<bool, Error> {
-    let mut context_state = context_state.borrow_mut();
     let mut dispatched_ops = false;
 
+    // Poll any pending task spawner tasks. Note that we need to poll separately because otherwise
+    // Rust will extend the lifetime of the borrow longer than we expect.
+    let tasks = context_state.borrow().task_spawner_factory.poll_inner(cx);
+    if let Poll::Ready(tasks) = tasks {
+      // TODO(mmastrac): we are using this flag
```
bartlomieju, line 1987: What does that TODO mean?

core/runtime/jsruntime.rs
```diff
@@ -1961,9 +1978,21 @@ impl JsRuntime {
     scope: &mut v8::HandleScope,
     context_state: &RefCell<ContextState>,
   ) -> Result<bool, Error> {
-    let mut context_state = context_state.borrow_mut();
     let mut dispatched_ops = false;
 
+    // Poll any pending task spawner tasks. Note that we need to poll separately because otherwise
+    // Rust will extend the lifetime of the borrow longer than we expect.
+    let tasks = context_state.borrow().task_spawner_factory.poll_inner(cx);
```
bartlomieju, line 1985: I assume this is free if there are no tasks?

mmastrac: Correct. Just one boolean check (atomic but cheap). Will also go away with wakers

littledivy:
Just tried it out. Works great with rust wasm streaming pipeline!

core/tasks.rs
```diff
@@ -0,0 +1,193 @@
...
+    self.has_tasks.load(std::sync::atomic::Ordering::SeqCst)
+  }
+
+  /// Poll this set of tasks, returning a non-empty set of tasks if there have
+  /// been any queued, or registering the waker if not.
+  pub fn poll_inner(&self, cx: &mut Context) -> Poll<Vec<UnsendTask>> {
+    // Check the flag first -- if it's false we definitely have no tasks
+    if !self
+      .has_tasks
+      .swap(false, std::sync::atomic::Ordering::SeqCst)
```
littledivy, line 55: What would happen if a task is scheduled after this `poll_inner` atomic check? `has_pending_task` would spin the event loop?

core/tasks.rs
```diff
@@ -0,0 +1,193 @@
...
+  tasks: Arc<V8TaskSpawnerFactory>,
+}
+
+// SAFETY: the underlying V8TaskSpawnerFactory is not Send, but we always submit Send tasks
+// to it from this spawner.
+unsafe impl Send for V8CrossThreadTaskSpawner {}
+
+impl V8CrossThreadTaskSpawner {
+  /// Spawn a task that runs within the [`crate::JsRuntime`] event loop, potentially (but not
+  /// required to be) from a different thread than the runtime is running on.
```
littledivy, line 135: Where will we be using `V8CrossThreadTaskSpawner`? I'm wary of adding it unless we want to use it right now.

littledivy (commented):
Looks good, left some comments

core/tasks.rs
mmastrac: This will replace middleware for FFI and NAPI

core/tasks.rs
mmastrac: Let me look closer at it again but there should only be the following cases as we're performing an atomic swap to false here:

1. has_tasks is true, we have tasks -- they are returned and has_tasks is set to false
2. has_tasks is true, but we lost a race and have no tasks -- we return pending and has_tasks becomes false. The waker is registered.
3. has_tasks is false -- we return pending and register our waker. If we are racing a cross-thread submission, the waker will be woken and we'll be awoken on the next time through

In all cases, has_tasks ends up as false unless a new submission is made. There should be no way for has_tasks to get set to false without tasks being extracted and run, and there should be no way for has_tasks to remain true while we have no tasks (causing a spin).

mmastrac: I added some tests for serial and parallel spawning in a multi-threaded runtime to confirm that this cannot deadlock, even under Miri.

littledivy: Sounds good, thanks

core/tasks.rs
```diff
@@ -0,0 +1,273 @@
+use futures::task::AtomicWaker;
```
littledivy, line 1:
```suggestion
// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.

use futures::task::AtomicWaker;
```

core/tasks.rs
```diff
@@ -0,0 +1,273 @@
...
+  tasks: Mutex<Vec<SendTask>>,
+  /// A flag we can poll without any locks.
+  has_tasks: AtomicBool,
+  /// The polled waker, woken on task submission.
+  waker: AtomicWaker,
+  /// This cannot be Send because it may contain `!Send` tasks submitted by a [`V8TaskSpawner`]. It is
+  /// only safe to send this object to another thread if you plan on submitting [`Send`] tasks to it,
+  /// which is what [`V8CrossThreadTaskSpawner`] does.
+  _unsend_marker: PhantomData<*const ()>,
+}
```
littledivy, line 31:
```suggestion
/// V8TaskSpawnerFactory cannot be Send because it may contain `!Send` tasks 
/// submitted by a [`V8TaskSpawner`]. Use [`V8CrossThreadTaskSpawner`] to
/// submit [`Send`] tasks.
#[derive(Default)]
pub(crate) struct V8TaskSpawnerFactory {
  // TODO(mmastrac): ideally we wouldn't box if we could use arena allocation and a max submission size
  // TODO(mmastrac): we may want to split the Send and !Send tasks
  /// The set of tasks, non-empty if `has_tasks` is set.
  tasks: Mutex<Vec<SendTask>>,
  /// A flag we can poll without any locks.
  has_tasks: AtomicBool,
  /// The polled waker, woken on task submission.
  waker: AtomicWaker,
  _unsend_marker: PhantomData<*const ()>,
}
```

core/tasks.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+  }
+
+  pub fn new_cross_thread_spawner(self: Arc<Self>) -> V8CrossThreadTaskSpawner {
+    V8CrossThreadTaskSpawner { tasks: self }
+  }
+
+  pub fn has_pending_tasks(&self) -> bool {
+    self.has_tasks.load(std::sync::atomic::Ordering::SeqCst)
+  }
```
littledivy, line 47:
```suggestion

  /// False guarantees that there are no queued tasks.
  pub fn has_pending_tasks(&self) -> bool {
    self.has_tasks.load(std::sync::atomic::Ordering::SeqCst)
  }
```

core/tasks.rs
```diff
@@ -0,0 +1,273 @@
...
+  /// event loop's [`v8::HandleScope`].
+  pub fn spawn<F>(&self, f: F)
+  where
+    F: FnOnce(&mut v8::HandleScope) + 'static,
+  {
+    // SAFETY: we are transmuting Send into a !Send handle but we can guarantee this object will never
+    // leave the current thread because `V8TaskSpawner` is !Send.
+    let task: Box<dyn FnOnce(&mut v8::HandleScope<'_>)> = Box::new(f);
+    let task: Box<dyn FnOnce(&mut v8::HandleScope<'_>) + Send> =
+      unsafe { std::mem::transmute(task) };
```
littledivy, line 120:
```suggestion
    let task: Box<dyn FnOnce(&mut v8::HandleScope<'_>)> = Box::new(f);
    // SAFETY: we are transmuting Send into a !Send handle but we guarantee this object will never
    // leave the current thread because `V8TaskSpawner` is !Send.
    let task: Box<dyn FnOnce(&mut v8::HandleScope<'_>) + Send> =
      unsafe { std::mem::transmute(task) };
```

Moves safety comment above `unsafe` usage.

core/tasks.rs
```diff
@@ -0,0 +1,273 @@
...
+unsafe impl Send for V8CrossThreadTaskSpawner {}
+
+impl V8CrossThreadTaskSpawner {
+  /// Spawn a task that runs within the [`crate::JsRuntime`] event loop, potentially (but not
+  /// required to be) from a different thread than the runtime is running on.
+  ///
+  /// The task is handed off to be run the next time the event loop is polled, and there are
+  /// no guarantees as to when this may happen.
+  ///
+  /// # Important Notes
```
littledivy, line 142:
```suggestion
  /// # Safety
```

It is potentially unsafe if a V8 exception is not handled by the user.

core/tasks.rs
```diff
@@ -0,0 +1,273 @@
...
+  /// there are no checks for this case.
+  ///
+  /// As this function blocks until the task has run to completion (or panics/deadlocks), it is
+  /// safe to borrow data from the local environment and use it within the closure.
+  ///
+  /// The task is handed off to be run the next time the event loop is polled, and there are
+  /// no guarantees as to when this may happen, however the function will not return until the
+  /// task has been fully run to completion.
+  ///
+  /// # Important Notes
```
littledivy, line 170: Ditto

core/tasks.rs
mmastrac: I really need to set up copyright checker

core/tasks.rs
mmastrac: Tweaked comment with a similar message

bartlomieju (approved):
LGTM, I really like that it's an additive change and we can migrate code in CLI gradually 👍

PR #351, opened 2023-11-25
chore: prune some deps and bump rust toolchain

mmastrac:


bartlomieju:
Looks like we're using Rust v1.37.1 here and we need something newer.

mmastrac:
The tool for unused deps unexpectedly bumped its min version.

bartlomieju (approved):

PR #344, opened 2023-11-23
[WIP] feat: generic handling of import attributes

bartlomieju:
PoC for importing text and css modules using import attributes:

```
cargo run --example fs_module_loader ./import_attrs.js
Run ./import_attrs.js
styles .foo aYTs9q_foo
styles .bar aYTs9q_bar

```

PR #358, opened 2023-11-27
refactor: add support for generic synthetic modules

bartlomieju:
This commit refactors handling of "synthetic modules". Instead of having
a very specialized method for creating JSON modules, we now have
a more generic "ModuleMap::create_synthetic_module" method that
will allow to support other synthetic modules like text/bytes/url.

This will be useful for https://github.com/denoland/deno_core/pull/344.

### DIFF
--- a/core/modules/map.rs
+++ b/core/modules/map.rs
@@ -267,51 +267,36 @@ impl ModuleMap {
     Ok(module_id)
   }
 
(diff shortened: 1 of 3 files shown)


mmastrac (approved):
Nice and surgical. 👍

Issue #329, opened 2023-11-14
Rewrite `JsRealm::evaluate_pending_module`


bartlomieju:
Discussed with @mmastrac. In an effort to decrease complexity and a number of actions we need to do in each event loop tick we want to rewrite `JsRealm::evaluate_pending_module`.

Instead of storing checking if a promise stored in `ContextState.pending_mod_evaluate` on each tick (if there's a pending module evaluation) we'd like to use [`Promise::then2`](https://docs.rs/v8/latest/v8/struct.Promise.html#method.then2) in `JsRealm::mod_evaluate`.

More precisely, instead of doing this:
https://github.com/denoland/deno_core/blob/79c876c85529d9ed9ab2aa372c878c99347e4a1b/core/runtime/jsrealm.rs#L628-L645

we would like to use `Promise::then2` to attach a handler that would perform steps from `JsRealm::evaluate_pending_module`. This should save a bunch of `HandleScope` creations as well as `v8::Global` cloning, which in turn might yield further simplifications of this part of the codebase.

PR #359, opened 2023-11-27
refactor(core): extract exception_state [WIP]

mmastrac:


core/modules/map.rs
```diff
@@ -701,155 +687,87 @@ impl ModuleMap {
...
-      });
-    }
-
-    let maybe_value = module.evaluate(tc_scope);
+    let Some(value) = module.evaluate(tc_scope) else {
+      if tc_scope.has_terminated() || tc_scope.is_execution_terminating() {
+        // Just drop the sender to send the generic closed error
+      } else {
+        // TODO(mmastrac): is this even posible?
+        assert!(module.get_status() == v8::ModuleStatus::Errored);
```
ry, line 701: assert_eq ?

core/modules/map.rs
```diff
@@ -83,6 +83,8 @@ pub(crate) struct ModuleMap {
   pub(crate) loader: RefCell<Rc<dyn ModuleLoader>>,
   pub(crate) import_meta_resolve_cb: ImportMetaResolveCallback,
 
+  // TODO(mmastrac): temporarily crate-public to avoid RefCell on ContextState
+  pub(crate) exception_state: Rc<ExceptionState>,
```
bartlomieju, line 87: I might be nitpicking here, but seems trange to store `exception_state` on `ModuleMap`. Is it duplicated in `JsRuntimeState`?

core/modules/map.rs
```diff
@@ -48,14 +51,11 @@ use super::AssertedModuleType;
 type PrepareLoadFuture =
   dyn Future<Output = (ModuleLoadId, Result<RecursiveModuleLoad, Error>)>;
 
-use super::default_import_meta_resolve_cb;
```
bartlomieju, line 51: What happened to this guy?

core/modules/map.rs
```diff
@@ -145,16 +133,18 @@ impl ModuleMap {
       for m in not_evaluated {
         msg.push_str(&format!("  - {}\n", m));
       }
-      panic!("{}", msg);
+      // panic!("{}", msg);
```
bartlomieju, line 136: Uncomment before landing, this is a useful debugging tool.

core/modules/map.rs
```diff
@@ -670,15 +665,18 @@ impl ModuleMap {
   /// Implementors must manually call [`JsRuntime::run_event_loop`] to drive
   /// module evaluation future.
   ///
-  /// `Error` can usually be downcast to `JsError` and should be awaited and
-  /// checked after [`JsRuntime::run_event_loop`] completion.
+  /// Rejection errors from module evaluation are treated as unhandled rejections, and
```
bartlomieju, line 668: "Rejection errors" could be worded better, I don't really understand what it means here.

core/modules/map.rs
```diff
@@ -702,155 +700,115 @@ impl ModuleMap {
...
-    }
-
-    let maybe_value = module.evaluate(tc_scope);
+    let Some(value) = module.evaluate(tc_scope) else {
+      if tc_scope.has_terminated() || tc_scope.is_execution_terminating() {
+        let undefined = v8::undefined(tc_scope).into();
+        _ = sender
+          .send(exception_to_err_result(tc_scope, undefined, true, false));
+      } else {
+        debug_assert_eq!(module.get_status(), v8::ModuleStatus::Errored);
```
bartlomieju, line 715: Why `debug_assert_eq` and not `assert_eq`?

core/modules/map.rs
```diff
@@ -702,155 +700,115 @@ impl ModuleMap {
...
-      pending_mod_evaluate
-        .sender
-        .send(exception_to_err_result(tc_scope, exception, false))
+      sender
+        .send(exception_to_err_result(tc_scope, exception, false, false))
         .expect("Failed to send module evaluation error.");
-    } else if let Some(value) = maybe_value {
-      assert!(
+    } else {
+      debug_assert!(
```
bartlomieju, line 732: Ditto

core/modules/map.rs
```diff
@@ -702,155 +700,115 @@ impl ModuleMap {
...
+         _rv: v8::ReturnValue| {
+          let mut sender = get_sender(args.data());
+          sender.module_map.pending_mod_evaluation.set(false);
+          sender.module_map.module_waker.wake();
+          _ = sender.sender.take().unwrap().send(Ok(()));
+          scope.throw_exception(args.get(0));
+        },
+      )
+      .data(evaluation.into())
+      .build(tc_scope);
```
bartlomieju, line 778: IIRC we discussed to hoist these functions to top-level. Maybe consider still doing that?

core/runtime/exception_state.rs
```diff
@@ -0,0 +1,127 @@
+use crate::error::exception_to_err_result;
```
bartlomieju, line 1: Copyright 🙃

core/runtime/exception_state.rs
```diff
@@ -0,0 +1,127 @@
...
+use anyhow::Error;
+use std::cell::Cell;
+use std::cell::RefCell;
+use std::collections::VecDeque;
+use std::rc::Rc;
+
+#[derive(Default)]
+pub(crate) struct ExceptionState {
+  // TODO(nayeemrmn): This is polled in `exception_to_err_result()` which is
+  // flimsy. Try to poll it similarly to `pending_promise_rejections`.
```
bartlomieju, line 11: Is this comment still valid?

core/runtime/exception_state.rs
```diff
@@ -0,0 +1,127 @@
...
+  pub(crate) js_build_custom_error_cb:
+    RefCell<Option<Rc<v8::Global<v8::Function>>>>,
+  pub(crate) js_promise_reject_cb:
+    RefCell<Option<Rc<v8::Global<v8::Function>>>>,
+  pub(crate) js_format_exception_cb:
+    RefCell<Option<Rc<v8::Global<v8::Function>>>>,
+}
+
+impl ExceptionState {
+  pub fn destroy(&self) {
```
bartlomieju, line 25: Maybe `reset()` since the struct is still valid after calling this fn

core/runtime/exception_state.rs
```diff
@@ -0,0 +1,127 @@
...
+    self.dispatched_exception.set(Some(exception));
+    self.dispatched_exception_is_promise.set(promise);
+  }
+
+  pub(crate) fn check_exception_condition(
+    &self,
+    scope: &mut v8::HandleScope,
+  ) -> Result<(), Error> {
+    // TODO(mmastrac): rewrite before landing
+    if let Some(exception) = self.get_dispatched_exception_as_local(scope) {
```
bartlomieju, line 64: Do you want to rewrite it before landing? How should it be rewritten?

core/runtime/exception_state.rs
```diff
@@ -0,0 +1,127 @@
...
+        .dispatched_exception
+        .as_ptr()
+        .as_ref()
+        .unwrap_unchecked()
+    }
+    .as_ref()
+    .map(|global| v8::Local::new(scope, global))
+  }
+
+  /// Calls the project rejection callback, if one exists.
```
bartlomieju, line 96: What's "project"?

core/runtime/jsruntime.rs
```diff
@@ -1942,8 +1962,13 @@ impl JsRuntime {
   /// Implementors must manually call [`JsRuntime::run_event_loop`] to drive
   /// module evaluation future.
   ///
-  /// `Error` can usually be downcast to `JsError` and should be awaited and
-  /// checked after [`JsRuntime::run_event_loop`] completion.
+  /// Rejection errors from module evaluation are treated as unhandled rejections, and
```
bartlomieju, line 1965: Ditto about "rejection errors". Is it on purpose that this comment is duplicated?

core/modules/map.rs
mmastrac: I want to avoid too many asserts in release mode -- we assert quite a bit and our debug tests should be sufficient to catch these. Generally we should avoid `assert` because it bloats the codebase and inhibits the optimizer.

core/modules/map.rs
mmastrac: I tried to do that but the way that the rusty_v8 mapping stuff works makes this extremely difficult to do. I removed as much duplication as I could.

I think we'll need to upstream changes to make these functions more ergonomic as it's generally painful to use them w/closures and I don't think it needs to be.

core/runtime/exception_state.rs
mmastrac: typo -- should be promise.

core/modules/map.rs
mmastrac: I removed the `Default` for ModuleMap as part of this change which means it is no longer necessary.

core/modules/map.rs
mmastrac: Ack, missed this. Good catch

core/runtime/exception_state.rs
mmastrac: This is a pre-teardown method so I cannot guarantee that it can be re-used after calling destroy. I think if we do add a runtime reset function we can rename it and test it as such but I wouldn't want to give that impression as-is.

core/runtime/exception_state.rs
mmastrac: I _think_ we can just check to see if there's a dispatched exception and call `exception_to_err_result` if so. 

I am not a fan of `exception_to_err_result` in general, but it does a lot of work that I don't want to re-write right now. I will rework this code a bit.

Cargo.toml
```diff
@@ -31,6 +31,7 @@ bytes = "1.4.0"
 cooked-waker = "5"
 derive_more = "0.99.17"
 futures = "0.3.21"
+itertools = "0"
```
mmastrac, line 34: TODO: remove this

core/modules/map.rs
bartlomieju: So what if I use `FsModuleLoader`? `import.meta.resolve` will not work anymore?

core/runtime/exception_state.rs
bartlomieju: Okay I see - in this case can it accept `self`? The types currently don't really communicate well that it shouldn't be used anymore.

bartlomieju:
Closes https://github.com/denoland/deno_core/issues/329?

core/modules/map.rs
mmastrac: Sorry, I should be clearer -- the default value was already handled in JsRuntime. The ModuleMap default was never really necessary:

```
    let import_meta_resolve_cb = options
      .import_meta_resolve_callback
      .unwrap_or_else(|| Box::new(default_import_meta_resolve_cb));
```

core/runtime/exception_state.rs
mmastrac: Because of the `Rc<...>` we're kind of stuck with `&self`. It's challenging because we need to ensure that all these callbacks are torn down before the snapshot is taken, and we still have a lot of intertwined classes.

I can improve the docs and maybe write a TODO on it. As I continue to tear out JsRealm, we may actually be able to get to a point where we can manually release all the isolate resources before the snapshot is taken. Right now it's pretty ugly.

core/modules/map.rs
mmastrac: This is some leftover cruft that hasn't been fully refactored yet...

It's on the ContextState and ModuleMap for now, but ContextState still has a RefCell which makes it more expensive to access in the event loop. The ModuleMap technically only needs this because we're still threading `reportError` errors through it. 

Once I remove that path (especially if we can get to a single unhandled rejection reporting channel), we might be able to get rid of this entirely.

core/runtime/exception_state.rs
mmastrac: Renamed to "prepare_to_destroy" to make it a little clearer.

core/runtime/exception_state.rs
mmastrac: Fixed, and renamed this method to `track_promise_rejection` since it's not actually calling a callback any more as well.

core/modules/map.rs
mmastrac: Removed this comment -- JsRuntime has the primary one now.

core/modules/map.rs
mmastrac: Looked at this briefly and I think it's better to do in a follow-up pass. Even if I can remove it, the RefCell on ContextState will be a problem.

core/runtime/exception_state.rs
mmastrac: Unfortunately yes 😢 

We should be able to make this work like a waker, but that will need a fair bit of work.

core/runtime/exception_state.rs
mmastrac: Simplified it

core/runtime/jsruntime.rs
mmastrac: Reworked this comment.. Let me know what you think

core/modules/map.rs
bartlomieju: Ah, got it. My bad then

bartlomieju (approved):
LGTM

PR #263, opened 2023-10-11
feat: lazy loaded esm sources

crowlKats:
This commit adds a new lazy to lazy load internal runtime code
authored in ES modules using "op_lazy_load_esm" op.

The code needs to embedded in the binary to be eliglble for lazy-loading,
and that can be achieved using "lazy_loaded_esm" option on the
"extension!" macro.

Prerequisite for https://github.com/denoland/deno/pull/20812

PR #366, opened 2023-12-05
feat: add ext:core/mod.js

bartlomieju:
This commit adds "ext:core/mod.js" built-in ES module that reexports
"core", "internals" and "primordials" properties of the 
"globalThis.__bootstrap" namespace. This is very convenient for
embedders that author runtime code using ES modules instead of
scripts, because it allows to import these props directly instead of
capturing "globalThis.__bootstrap" namespace.

To achieve that a new "ModuleMap::lazy_load_es_module_from_code"
method was added that accepts a specifier and source code; instantiates
and evaluates the provided code as ES module.

This will be very useful for https://github.com/denoland/deno_core/pull/263
and https://github.com/denoland/deno/pull/21422.


bartlomieju:
Probably worth to run some benchmarks to see startup time difference, but I expect it'll be <0.1ms.

core/modules/map.rs
```diff
@@ -1301,6 +1302,58 @@ impl ModuleMap {
...
+  /// possible to evaluate it synchronously).
+  ///
+  /// It is caller's responsibility to ensure that not duplicate specifiers are
+  /// passed to this method.
+  pub(crate) fn lazy_load_es_module_from_code(
+    &self,
+    scope: &mut v8::HandleScope,
+    module_specifier: &str,
+    source_code: ModuleCode,
+  ) -> Result<v8::Global<v8::Value>, Error> {
```
littledivy, line 1318: Very nice, infact this could also be used for a code caching implementation

core/modules/map.rs
```diff
@@ -1301,6 +1302,58 @@ impl ModuleMap {
...
+    let promise = v8::Local::<v8::Promise>::try_from(value).unwrap();
+    let result = promise.result(scope);
+    if !result.is_undefined() {
+      return Err(
+        exception_to_err_result::<()>(scope, result, false).unwrap_err(),
+      );
+    }
+
+    let status = module_local.get_status();
+    assert_eq!(status, v8::ModuleStatus::Evaluated);
```
littledivy, line 1351: Maybe return an `Err` if module has to be asyncronously evaluated (i.e contains top level async)

littledivy (approved):
LGTM

PR #367, opened 2023-12-06
refactor: combine LoggingModuleLoader and CountingModuleLoader

bartlomieju:
This commit combines "LoggingModuleLoader" and "CountingModuleLoader"
into "TestingModuleLoader". 

They serve very similar purpose and it's better to get more data for assertions
in tests.


mmastrac (approved):
LGTM

PR #377, opened 2023-12-07
feat(core): Adding deno_core timers API

mmastrac:
This is a new fundamental timer API that supports the "run steps after a timeout" specification of https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html


core/01_core.js
```diff
@@ -201,8 +202,17 @@
       }
     }
 
+    const timers = arguments[arguments.length - 2];
+    if (timers) {
+      for (let i = 0; i < timers.length; i += 2) {
+        timerDepth = timers[i];
+        timers[i + 1]();
```
mmastrac, line 209: Currently leaving try/catch to embedders but this may change.

core/web_timeout.rs
```diff
@@ -0,0 +1,635 @@
...
+
+      eprintln!("count={count} ref_count={ref_count}");
+
+      let v = poll_all(&timers).await;
+      assert_eq!(v.len(), count);
+
+      assert!(timers.is_empty());
+      assert!(!timers.has_pending_timers());
+    });
+  }
```
bartlomieju, line 634: Very nice tests 👍

bartlomieju (approved):
LGTM, this is a great first pass. I'm sure we'll do some back-and-forth so let's land it as is for now.

PR #384, opened 2023-12-12
refactor(core): clean up timer and promise code 

mmastrac:
Deno PR: https://github.com/denoland/deno/pull/21519

Note: `PollEventLoopOptions` was changed so the defaults are now the ones used in the majority of cases.

Started working on cleaning up the call_and_await and resolve_value code in JsRuntime. I managed to turn these into pure futures that will resolve independently of the event loop (though still requiring someone to poll it), which means that you can now do something like this:

```
#[op2]
fn call(scope: &mut v8::HandleScope, #[global] f: v8::Global<v8::Function>) -> impl Future<...> {
  JsRuntime::scoped_call(f).await
}
```

This should allow for some simplification of any code that's currently using `call_and_await`.  

`with_event_loop` is split into a "promise" and "future" version. The promise version will throw if the event loop terminates without the promise being resolved. The future version will continue polling until the future resolves (even if the event loop completes). This matches the current behaviour that we are relying on in deno.


core/runtime/bindings.rs
```diff
@@ -538,10 +538,10 @@ pub(crate) fn watch_promise<'s, F>(
 ) -> Option<v8::Local<'s, v8::Promise>>
 where
   F: FnOnce(
-    &mut v8::HandleScope,
-    v8::ReturnValue,
-    Result<v8::Local<v8::Value>, v8::Local<v8::Value>>,
-  ),
+      &mut v8::HandleScope,
```
mmastrac, line 541: This was actually wrong and could cause crashes. Fixed by adding 'static

core/runtime/jsruntime.rs
```diff
@@ -481,7 +512,7 @@ pub struct PollEventLoopOptions {
 impl Default for PollEventLoopOptions {
   fn default() -> Self {
     Self {
-      wait_for_inspector: true,
+      wait_for_inspector: false,
```
mmastrac, line 515: Default is false because we mostly poll the event loop with false and this cleans up a lot of code

core/web_timeout.rs
```diff
@@ -402,7 +407,7 @@ impl<T: Clone> WebTimers<T> {
   }
 
   pub fn has_pending_timers(&self) -> bool {
-    self.timers.borrow().len() > self.unrefd_count.get()
+    self.len() > self.unref_len()
```
mmastrac, line 410: This was wrong -- we didn't omit tombstones from the count

core/runtime/jsruntime.rs
bartlomieju: I think in most cases it doesn't make any difference, but I'm fine with this change.

core/runtime/jsruntime.rs
```diff
@@ -1240,89 +1271,123 @@ impl JsRuntime {
     )
   }
 
+  /// Call a function and returns a future resolving with the return value of the
```
bartlomieju, line 1274:
```suggestion
  /// Call a function and return a future resolving with the return value of the
```

core/runtime/jsruntime.rs
```diff
@@ -1240,89 +1271,123 @@ impl JsRuntime {
...
+
+  /// Call a function and returns a future resolving with the return value of the
+  /// function. If the function returns a promise, the future will resolve only once the
+  /// event loop resolves the underlying promise. If the future rejects, the future will
+  /// resolve with the underlying error.
+  ///
+  /// The event loop must be polled seperately for this future to resolve. If the event loop
+  /// is not polled, the future will never make progress.
+  pub fn scoped_call(
+    scope: &mut v8::HandleScope,
```
bartlomieju, line 1296: Not a fan of this name - maybe `call_with_scope`/`call_in_scope`? What's the use case for it? Just to pass scope explicitly?

core/runtime/jsruntime.rs
```diff
@@ -1419,93 +1484,68 @@ impl JsRuntime {
     ));
   }
 
-  pub fn poll_value(
+  /// Waits for the given value to resolve while polling the event loop.
+  ///
+  /// This future resolves when either the value is resolved or the event loop runs to
+  /// completion.
+  pub fn resolve(
```
bartlomieju, line 1491: `await_promise` or `promise_into_future`?

core/runtime/tests/error.rs
```diff
@@ -76,37 +76,32 @@ if (errMessage !== "higher-level sync error: original sync error") {
...
-      _ => panic!(),
-    }
-    Poll::Ready(())
-  })
-  .await;
+  let resolve = runtime.resolve(promise);
+  futures::executor::block_on(
+    runtime.with_event_loop_promise(resolve, PollEventLoopOptions::default()),
+  )
+  .unwrap();
```
bartlomieju, line 104: Since it's an async test fn, then maybe `.await` it instead?

bartlomieju (approved):
Feel free to ignore naming nitpicks, we can address them another time

core/runtime/jsruntime.rs
mmastrac: I couldn't really come up with a better name, `call_with_scope` might work -- it allows you to make a call without needing a `&mut self` for the `JsRuntime`. To be honest, I expect to pull most of these methods off into some sort of `JsRuntimeHandle` but that needs a bit of experimentation work.

core/runtime/tests/error.rs
mmastrac: Good call.. I unwrapped the inner poll_fn later on and missed this one.

core/runtime/jsruntime.rs
mmastrac: It broke a few tests when I was porting it over and using ::default() everywhere, but I think I'd like to circle back on this later on as we might be able to move it out of poll_event_loop somehow (not sure how yet)

PR #387, opened 2023-12-19
feat: Add RuntimeOptions::wait_for_inspector_disconnect_callback

bartlomieju:
This commit adds "RuntimeOptions::wait_for_inspector_disconnect_callback"
that allows to pass a callback that will be called when the event loop runs out of
work, but there are active, non-blocking inspector sessions - that is sessions
from "Chrome DevTools". This allows to print a message notifying user that
the program has finished, but is waiting for DevTools to disconnect to exit
the process.

PR in `deno`: https://github.com/denoland/deno/pull/21633

### DIFF
--- a/core/runtime/jsruntime.rs
+++ b/core/runtime/jsruntime.rs
@@ -70,6 +70,7 @@ use std::task::Poll;
 use std::task::Waker;
 use v8::Isolate;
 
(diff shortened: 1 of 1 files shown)


mmastrac (approved):
LGTM

PR #390, opened 2023-12-20
 refactor(core): add arenas and op driver infrastructure

mmastrac:
Building blocks for a new `OpSet`.

This commit introduces three distinct types of arenas:

1. Raw Arena: Provides basic functionality for memory allocation and deallocation within an arena structure. It manages raw memory blocks and allows recycling of allocated data.

2. Shared Atomic Arena: Offers thread-safe allocation and deallocation capabilities within the arena. It incorporates a mutex for exclusive access to the internal data, enabling multiple threads to safely allocate, share, and deallocate objects within the arena.

3. Unique Arena: Enables exclusive ownership semantics similar to a Box, maintaining the sole ownership of allocated data within the arena. It manages the cleanup of allocated data when the ArenaUnique instance is dropped, ensuring exclusive access and efficient memory management.

In addition, this updates our old JoinSet-based op system with the new arenas. Futures that resolve immediately are about 25% faster, and deferred ops are ~10% faster.

Before:

```
test bench_op_async_void                  ... bench:      79,904 ns/iter (+/- 4,763)
test bench_op_async_void_deferred         ... bench:   1,147,430 ns/iter (+/- 73,283)
test bench_op_async_void_deferred_nofast  ... bench:   1,120,793 ns/iter (+/- 40,374)
test bench_op_async_void_deferred_return  ... bench:   1,111,990 ns/iter (+/- 36,113)
test bench_op_async_void_indirect         ... bench:     139,757 ns/iter (+/- 3,190)
```

After:

```
test bench_op_async_void                  ... bench:      58,617 ns/iter (+/- 1,916)
test bench_op_async_void_deferred         ... bench:   1,049,621 ns/iter (+/- 49,606)
test bench_op_async_void_deferred_nofast  ... bench:   1,048,056 ns/iter (+/- 26,152)
test bench_op_async_void_deferred_return  ... bench:   1,027,673 ns/iter (+/- 41,257)
test bench_op_async_void_indirect         ... bench:     117,257 ns/iter (+/- 2,621)
```


core/arena/unique_arena.rs
```diff
@@ -0,0 +1,359 @@
...
+struct ArenaBoxData<T> {
+  #[cfg(debug_assertions)]
+  signature: usize,
+  arena_data: NonNull<ArenaUniqueData<T>>,
+  data: T,
+}
+
+impl<T: 'static> ArenaBox<T> {
+  /// Offset of the `ptr` field within the `ArenaBox` struct.
+  const PTR_OFFSET: usize = memoffset::offset_of!(ArenaBox<T>, ptr);
```
bartlomieju, line 37: I wonder if instead of using this crate, we could use newly stabilized APIs from the standard library? https://blog.rust-lang.org/2023/12/28/Rust-1.75.0.html#pointer-byte-offset-apis

core/arena/unique_arena.rs
```diff
@@ -0,0 +1,359 @@
...
+use std::future::Future;
+use std::pin::Pin;
+use std::ptr::NonNull;
+
+use crate::arena::raw_arena::RawArena;
+
+use super::alloc;
+
+#[cfg(debug_assertions)]
+const SIGNATURE: usize = 0x8877665544332211;
```
bartlomieju, line 12: Can you add a comment explaining what it this magic value?

core/arena/shared_atomic_arena.rs
```diff
@@ -0,0 +1,544 @@
...
+use std::sync::atomic::Ordering;
+
+use parking_lot::lock_api::RawMutex;
+
+use crate::arena::raw_arena::RawArena;
+
+use super::alloc;
+
+#[cfg(debug_assertions)]
+const SIGNATURE: usize = 0x1122334455667788;
```
bartlomieju, line 14: Please add a comment about this magic value.

core/benches/ops/opset.rs
```diff
@@ -0,0 +1,141 @@
...
+      0,
+    ))));
+  }
+  drop(v);
+}
+
+// fn bench_opset(b: &mut Bencher) {
+//   let runtime = tokio::runtime::Builder::new_current_thread()
+//     .enable_time()
+//     .build()
```
bartlomieju, line 30: Remove before landing?

core/runtime/jsruntime.rs
```diff
@@ -3,7 +3,10 @@ use super::bindings;
 use super::bindings::watch_promise;
 use super::exception_state::ExceptionState;
 use super::jsrealm::JsRealmInner;
+use super::op_driver::OpDriver;
 use super::snapshot_util;
+use crate::OpMetricsEvent;
+use crate::_ops::dispatch_metrics_async;
```
bartlomieju, line 9: Is it on purpose that this is now imported from `crate::_ops` and not `crate::ops_metrics`?

bartlomieju (commented):
First pass review. There's a lot of moving pieces here so I will need at least one more pass. Pinging @littledivy for additional review.

core/runtime/op_driver/erased_future.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+pub struct ErasedFuture<const MAX_SIZE: usize, Output> {
+  erased: TypeErased<MAX_SIZE>,
+  poll: unsafe fn(
+    this: Pin<&mut TypeErased<MAX_SIZE>>,
+    cx: &mut Context,
+  ) -> Poll<Output>,
+  _output: PhantomData<Output>,
+}
```
littledivy, line 80: A better name would be `StackFuture`.

But I think we should use the `stackfuture` crate from Microsoft https://github.com/microsoft/stackfuture - the implementation looks similar. `stackfuture` also has compile-time assertions when future does not fit.

littledivy (changes requested):

core/arena/unique_arena.rs
mmastrac: I didn't see those new APIs, sweet. We still need the memoffset crate to get the offsets, but we can simplify the pointer math below.

core/arena/unique_arena.rs
mmastrac: Sure -- these are all purely debug-build sanity checks on raw pointers.

core/arena/shared_atomic_arena.rs
```diff
@@ -0,0 +1,544 @@
...
+  /// Clones the `ArenaArc` reference, increments its reference count, and returns a raw pointer to the contained data.
+  ///
+  /// This function increments the reference count of the `ArenaArc`.
+  #[inline(always)]
+  pub fn clone_into_raw(arc: &ArenaArc<T>) -> NonNull<T> {
+    unsafe {
+      arc.ptr.as_ref().ref_count.fetch_add(1, Ordering::Relaxed);
+      Self::ptr_from_data(arc.ptr)
+    }
+  }
```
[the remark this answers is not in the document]
mmastrac: I thought about this, but I preferred to duplicate the code for now until I'm certain about the shape of the arena APIs. 

The trait is definitely the right way to go long-term.

core/runtime/jsruntime.rs
mmastrac: Nope, `rust-analyzer` picked the wrong import site, will fix.

core/runtime/op_driver/futuresunordered_driver.rs
```diff
@@ -0,0 +1,353 @@
...
+    }
+  }
+}
+
+impl OpDriver for FuturesUnorderedDriver {
+  fn submit_op_fallible<
+    R: 'static,
+    E: Into<Error> + 'static,
+    const LAZY: bool,
+    const DEFERRED: bool,
```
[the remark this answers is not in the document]
mmastrac: I could probably turn this into a u8 const param, but these bools are not new and the callsites are already a bit vague. I suspect I'll be removing one of these bools in the future after the OpSet work is complete as it doesn't make a lot of sense to have both.

core/runtime/op_driver/erased_future.rs
mmastrac: Thanks for the pointer to the `stackfuture` crate, I haven't seen that but it could be useful in other places.

In our case we cannot use the crate because we also need the `TypeErased` struct to erase the value types as well.

We don't need compile-time size assertions because we do a compile-time check and allocate on the heap if the future or return value doesn't fit (see the ValueLarge enum member, for instance).

`stackfuture` isn't really a great name for this as these don't live in the stack -- they live in the Arena. They are copied around like value types only at the very beginning, after which they are pinned in place inside the Arena.

littledivy: I see, sounds good.

core/arena/shared_atomic_arena.rs
```diff
@@ -0,0 +1,544 @@
...
+    panic!("A reservation must be completed or forgotten")
+  }
+}
+
+/// Represents an atomic reference-counted pointer into an arena-allocated object.
+pub struct ArenaArc<T> {
+  ptr: NonNull<ArenaArcData<T>>,
+}
+
+impl<T> ArenaArc<T> {
```
littledivy, line 29: Where is shader atomic arena used in this PR?

core/arena/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+    handle_alloc_error(layout);
+  };
+  alloc
+}
+
+#[inline(always)]
+fn alloc<T>() -> NonNull<T> {
+  let alloc = unsafe { std::alloc::alloc(Layout::new::<T>()) } as *mut _;
+  let Some(alloc) = NonNull::new(alloc) else {
+    handle_alloc_error(Layout::new::<T>());
```
littledivy, line 30: These should be marked unsafe. 0 sized layout is UB

mmastrac: Good catch. I'll assert instead as the compiler will elide it for valid types.

core/arena/shared_atomic_arena.rs
mmastrac: This is used in the OpSet changes that are not part of this PR

core/arena/unique_arena.rs
mmastrac: I'm actually going to punt on this one until we're certain of how reliable 1.75 is -- the APIs don't buy us a lot right now and I want to avoid locking in 1.75 right now. It'll be easy enough to fix with a future revision.

mmastrac: In the end I moved all the byte-manipulation functions to one spot to make it easier to swap out in the future

bartlomieju (approved):

PR #391, opened 2023-12-20
feat(serde_v8): From ExternalPointer to *mut and *const c_void

lrowe:
Also mark ExternalPointer as repr(transparent).


mmastrac (approved):
LGTM!

Issue #223, opened 2023-09-20
Extensions should expose their ops at `ext:<extname>/ops`


lucacasonato:
This will help reduce property lookups for ops significantly.

bartlomieju:
Could you elaborate a bit more what this would entail?

bartlomieju:
We should proceed with this approach. My work in https://github.com/denoland/deno_core/pull/402 should make this much more approachable and I can do a prototype for it.

PR #396, opened 2023-12-24
feat(core): add type functions

petamoriken:
related: https://github.com/denoland/deno/pull/21669


mmastrac:
These seem reasonable. Can you add unit tests to a new `testing/unit/type_test.ts`, please?

core/01_core.js
```diff
@@ -843,7 +843,31 @@
     deserialize: (buffer, options) => ops.op_deserialize(buffer, options),
     getPromiseDetails: (promise) => ops.op_get_promise_details(promise),
     getProxyDetails: (proxy) => ops.op_get_proxy_details(proxy),
+    isAnyArrayBuffer: (value) => ops.op_is_any_array_buffer(value),
```
mmastrac, line 846: New ops should use `ensureFastOps` so we can eventually land a fix for #223

core/ops_builtin.rs
```diff
@@ -43,9 +44,33 @@ deno_core::extension!(
     op_write_type_error,
     op_shutdown,
     op_format_file_name,
-    op_is_proxy,
     op_str_byte_length,
     op_panic,
+    ops_builtin_types::op_is_any_array_buffer,
```
mmastrac, line 49: We should add `op_is_array_buffer_view`, `op_is_data_view` and `op_is_typed_array` -- those three ops could be very useful for fast-op dispatch selection.

mmastrac (approved):
LGTM assuming tests pass

PR #397, opened 2023-12-27
fix(ops): support returning non-`anyhow` `Result` from fast sync op

0f-0b:
```rs
// compiles
#[op2]
pub fn op_slow_anyhow_result(_: &mut v8::HandleScope) -> anyhow::Result<()> {
    Ok(())
}

// compiles
#[op2(fast)]
pub fn op_fast_anyhow_result() -> anyhow::Result<()> {
    Ok(())
}

// compiles
#[op2]
pub fn op_slow_std_io_result(_: &mut v8::HandleScope) -> std::io::Result<()> {
    Ok(())
}

// does not compile (fixed by this PR)
#[op2(fast)]
pub fn op_fast_std_io_result() -> std::io::Result<()> {
    Ok(())
}
```

### DIFF
--- a/ops/op2/README.md
+++ b/ops/op2/README.md
@@ -15,6 +15,12 @@ string. At this time there is no way to avoid this copy, though the `op` code
 does attempt to avoid any allocations where possible by making use of a stack
 buffer.
 
+## Fallible `op`s
+
+An `op` function may be declared to return `Result` to indicate that the `op` is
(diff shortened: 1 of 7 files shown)


mmastrac:
Looks good to me. Could you add a corresponding section to the ops/op2 README with a note that return values need to be `into anyhow::Error`?

mmastrac (approved):

PR #357, opened 2023-11-27
[WIP] Attribute types css

bartlomieju:
Follow up to #344. And blocked by it.

PR #398, opened 2023-12-27
refactor: ModuleLoader::load can return string or bytes

bartlomieju:
This commit changes signature of "ModuleLoader::load" and "ModuleSourceFuture"
to allow returning either string or bytes. When modules are evaluated and we received
bytes, we try to parse them as UTF-8 and error out if they're not valid UTF-8.

This is done in preparation for WASM and bytes imports.

Ref #344 
Ref #357


core/modules/loaders.rs
```diff
@@ -22,6 +23,8 @@ use std::future::Future;
 use std::pin::Pin;
 use std::rc::Rc;
 
+use super::ModuleSourceCode;
```
bartlomieju, line 26: Refactor to use crate::

core/modules/loaders.rs
```diff
@@ -155,7 +158,11 @@ impl ModuleLoader for ExtModuleLoader {
     let result = source.load();
     match result {
       Ok(code) => {
-        let res = ModuleSource::new(ModuleType::JavaScript, code, specifier);
+        let res = ModuleSource::new(
+          ModuleType::JavaScript,
+          ModuleSourceCode::String(code),
+          specifier,
+        );
```
bartlomieju, line 165: It would be great if I could somehow codify that `extension!` only support string source code

core/modules/loaders.rs
```diff
@@ -211,7 +218,11 @@ impl ModuleLoader for LazyEsmModuleLoader {
...
```
bartlomieju, line 225: Similarly here. Maybe Rust types are enough to communicate it though?

core/modules/loaders.rs
```diff
@@ -293,8 +304,13 @@ impl ModuleLoader for FsModuleLoader {
...
 
-      let code = std::fs::read_to_string(path)?.into();
-      let module = ModuleSource::new(module_type, code, module_specifier);
+      // TODO(bartlomieju): handle bytes as well
+      let code: FastString = std::fs::read_to_string(path)?.into();
+      let module = ModuleSource::new(
+        module_type,
+        ModuleSourceCode::String(code),
+        module_specifier,
+      );
```
bartlomieju, line 313: In here I'm actually erring on the side of always returning bytes - since we have a helper that tries to convert provided source code to a string this should be all fine (and would handle other types of modules like WASM and bytes easily. TBD

core/modules/loaders.rs
```diff
@@ -347,7 +363,7 @@ impl ModuleLoader for StaticModuleLoader {
     let res = if let Some(code) = self.map.get(module_specifier) {
       Ok(ModuleSource::new(
         ModuleType::JavaScript,
-        code.try_clone().unwrap(),
+        ModuleSourceCode::String(code.try_clone().unwrap()),
```
bartlomieju, line 366: Not sure about this one 🤷

core/modules/map.rs
```diff
@@ -248,10 +248,10 @@ impl ModuleMap {
     module_source: ModuleSource,
   ) -> Result<ModuleId, ModuleError> {
     let ModuleSource {
-      code,
       module_type,
       module_url_found,
       module_url_specified,
+      code,
```
bartlomieju, line 254: Revert

core/modules/map.rs
```diff
@@ -271,22 +271,24 @@ impl ModuleMap {
...
+    if let Some(module_id) = maybe_module_id {
+      debug!(
+        "Already-registered module fetched again: {:?}",
+        module_url_found
+      );
+      return Ok(module_id);
+    }
+
+    let code = ModuleSource::get_string_source(code)
+      .map_err(|e| ModuleError::Other(e.into()))?;
```
bartlomieju, line 284: This could probably use some `.context()`

core/modules/mod.rs
```diff
@@ -33,6 +34,13 @@ pub(crate) use map::ModuleMap;
 
 pub type ModuleId = usize;
 pub(crate) type ModuleLoadId = i32;
+
+#[derive(Debug)]
+pub enum ModuleSourceCode {
+  String(FastString),
+  Bytes(Box<[u8]>),
+}
+// TODO(bartlomieju): remove
```
bartlomieju, line 43: I think I still want to remove `ModuleCode`, or replace what `ModuleSourceCode` is right now

core/modules/mod.rs
```diff
@@ -253,12 +261,24 @@ impl ModuleSource {
...
     }
   }
+
+  pub fn get_string_source(
+    code: ModuleSourceCode,
+  ) -> Result<ModuleCode, FromUtf8Error> {
+    match code {
+      ModuleSourceCode::String(code) => Ok(code),
+      ModuleSourceCode::Bytes(bytes) => {
+        let str_ = String::from_utf8(bytes.to_vec())?;
```
bartlomieju, line 277: This performs a clone. Maybe I should store `Vec<u8>` instead in `ModuleSourceCode`?

PR #402, opened 2023-12-29
feat: Allow support for Wasm modules

bartlomieju:
More generic solution than #344 and #357.

It allows embedders to load any module type they want.

Still some rough edges that need to be fixed. I will probably
split this PR into a few more.


core/examples/fs_module_loader.rs
```diff
@@ -1,12 +1,87 @@
...
+  module_name: &FastString,
+  code: ModuleSourceCode,
+) -> Result<v8::Global<v8::Value>, Error> {
+  // FsModuleLoader always returns bytes.
+  let ModuleSourceCode::Bytes(buf) = code else {
+    unreachable!()
+  };
+
+  let code = String::from_utf8(buf).with_context(|| {
+    format!("Can't convert {:?} source code to string", module_name)
```
littledivy, line 62: nit: `std::str::from_utf8` and pass the borrowed value to `v8::String::new`

bartlomieju: Thanks, applied. I'm still debating what should be the signature of `ModuleSourceCode::Bytes`. Given that it will be stored in the `deno_graph`'s graph it should probably be `Arc<[u8]>`. I'll discuss this with @dsherret.

core/modules/mod.rs
```diff
@@ -41,13 +42,60 @@ pub(crate) type ModuleLoadId = i32;
...
+  fn from(value: Box<[u8]>) -> Self {
+    Self::Boxed(value)
+  }
+}
+
+impl From<&'static [u8]> for ModuleCodeBytes {
+  fn from(value: &'static [u8]) -> Self {
+    Self::Static(value)
+  }
+}
```
bartlomieju, line 96: @mmastrac does this look fine for you?

mmastrac: Looks good!

petamoriken:
I disagree with adding `type: "wasm"` and `type: "url"`.

## WebAssembly Modules

WebAssembly Modules are just now being standardized. For now it is the same as regular JavaScript modules, `import { fib } from "./fib.wasm";`, and there is no consensus to add a special `type` attribute. Pleas see https://github.com/WebAssembly/esm-integration/issues/42.

In addition, [Stage 3 Source Phase Imports](https://github.com/tc39/proposal-source-phase-imports) is in progress as a preliminary step. If it is to proceed, this should be implemented first.

```js
import source fooModule from "./foo.wasm";

const fooInstance = WebAssembly.instantiate(fooModule, { /* imports */ });
```

## URL Modules

This only resolves URLs, does not fetch anything, does not evaluate anything. It is beyond the scope of Import Attributes, which should be resolved in the proposed [Stage 1 Asset References](https://github.com/tc39/proposal-asset-references/). Please refer to [the Module Harmony note](https://github.com/tc39/notes/blob/main/meetings/2023-05/may-15.md#module-harmony-interaction-semantics-of-the-different-proposals) for more information on each phase of the import.

Import Phase Modifiers

@lucacasonato Any thoughts?

bartlomieju:
Thanks for the comment @petamoriken, we are still discussing wasm imports - "bytes" import will solve the biggest pain point of using WASM in Deno already - that is a need to `fetch()` the WASM file and it not being part of the module graph.

With "bytes" imports we can have a small helper function:
```js
export async function wasmShim(wasmBytes) {
  const wasmMod = await WebAssembly.compile(wasmBytes);
  const requestedImports = WebAssembly.Module.imports(wasmMod);
  const importedModules = await Promise.all(
    requestedImports.map((i) => i.module).map((m) => import(m)),
  );
  const importsObject = {};
  for (let i = 0; i < requestedImports.length; i++) {
    const importedModule = importedModules[i];
    const requestedImport = requestedImports[i];
    if (typeof importsObject[requestedImport.module] === "undefined") {
      importsObject[requestedImport.module] = {};
    }
    const import_ = importedModule[requestedImport.name];
    importsObject[requestedImport.module][requestedImport.name] = import_;
  }
  const result = await WebAssembly.instantiate(wasmMod, importsObject);
  return result;
}
```
and then load a WASM file like so:
```js
import wasmBytes from "./ffmpeg.wasm" with { type: "bytes" };
const ffmpeg = await wasmShim(wasmBytes);
```
This will make `ffmpeg.wasm` part of the module graph which will make this file be cached and very cheap to load.

petamoriken:
No objections to `type: "bytes"` or `type: "text"`. However, for WebAssembly, I think the standard Source Phase Imports are sufficient.

```js
import source wasmModule from "./ffmpeg.wasm";
const ffmpeg = await wasmShim(wasmModule);
```

JOTSR:
I think that using MIME type as module type assertion is a safer and more future-proof option than arbitrary string type. MIME type is more web-compliant and is more like future in-browser module assertion approach. Additionally, MIME type can prevent conflicts if a new alternative for an existing type emerges, and it can allow deno to verify the MIME type of remote imports and ensure host security.

|Type|MIME|
|:-|:-|
|wasm|application/wasm|
|bytes|application/octet-stream (any format), image/png (specific format)|
|text|text/html, text/plain (eg: markdown)|
|url|text/x-url|

bartlomieju:
> I have no objections to `type: "bytes"` or `type: "text"`.
> 
> However, for WebAssembly, I think the standard Source Phase Imports are sufficient.
> 
> ```js
> import source wasmModule from "./ffmpeg.wasm";
> const ffmpeg = await wasmShim(wasmModule);
> ```

Yes, that would be the best solution, however this is still a proposal that is not implemented by V8. Once this is supported we can discourage the use of "bytes" import and suggest to use `import source`.

> I think that using MIME type as module type assertion is a safer and more future-proof option than arbitrary string type. MIME type is more web-compliant and is more like future in-browser module assertion approach. Additionally, MIME type can prevent conflicts if a new alternative for an existing type emerges, and it can allow deno to verify the MIME type of remote imports and ensure host security.

I'm not sure I understand what you are suggesting here @JOTSR. This PR is just a proof-of-concept to get these types of imports working. The actual validation and decision if a particular import should work will be left to the embedder - in case of Deno CLI we already do checks against file extensions and MIME types.

petamoriken:
> > I have no objections to `type: "bytes"` or `type: "text"`.
> > However, for WebAssembly, I think the standard Source Phase Imports are sufficient.
> > ```js
> > import source wasmModule from "./ffmpeg.wasm";
> > const ffmpeg = await wasmShim(wasmModule);
> > ```
> 
> Yes, that would be the best solution, however this is still a proposal that is not implemented by V8. Once this is supported we can discourage the use of "bytes" import and suggest to use `import source`.

Can we achieve this somehow by transpiling using swc? Likely soon we will have source phase imports in swc.
https://github.com/swc-project/swc/pull/8279

bartlomieju:
> > > I have no objections to `type: "bytes"` or `type: "text"`.
> > > However, for WebAssembly, I think the standard Source Phase Imports are sufficient.
> > > ```js
> > > import source wasmModule from "./ffmpeg.wasm";
> > > const ffmpeg = await wasmShim(wasmModule);
> > > ```
> > 
> > 
> > Yes, that would be the best solution, however this is still a proposal that is not implemented by V8. Once this is supported we can discourage the use of "bytes" import and suggest to use `import source`.
> 
> Can we achieve this somehow by transpiling using swc? Likely soon we will have source phase imports in swc. [swc-project/swc#8279](https://github.com/swc-project/swc/pull/8279)

That could work, but we never transpile JS sources in Deno. It would be a big change (and have a noticeable impact on performance) to transpile JS sources as well. I will raise this topic within the team.

littledivy:
`type: "url"` is quite useful for publishing FFI libraries.

```js
import lib from "../libfoo.so";

const { symbols } = Deno.dlopen(new URL(lib).pathname, {});
```

JOTSR:
> I'm not sure I understand what you are suggesting here @JOTSR. This PR is just a proof-of-concept to get these types of imports working. The actual validation and decision if a particular import should work will be left to the embedder - in case of Deno CLI we already do checks against file extensions and MIME types.

I said that one possible benefit of allowing the developer to specify the MIME type (mostly for remote URLs) can enhance security and clarity. This is also consistent with many web APIs that require the MIME type for linked resources.

jsejcksn:
> Also using MIME type can allow "smart" imports in the future.
> 
> ```ts
> import htmlRaw from 'html' with { type: "text/plain" } //originally supported type
> //htmlRaw as string
> 
> import htmlDoc from 'html' with { type: "text/html" } //a future support
> //htmlDoc as Document
> ```

There's some prior discussion about that here:

- https://github.com/tc39/proposal-import-attributes/issues/114

core/runtime/jsruntime.rs
```diff
@@ -352,6 +353,7 @@ pub struct JsRuntimeState {
   /// It will be retrieved by `exception_to_err_result` and used as an error
   /// instead of any other exceptions.
   pub(crate) validate_import_attributes_cb: ValidateImportAttributesCb,
+  pub(crate) custom_module_evaluation_cb: Option<CustomModuleEvaluationCb>,
```
bartlomieju, line 356: After consideration I think this should be a method on the `ModuleLoader` trait.

core/modules/mod.rs
```diff
@@ -41,13 +42,60 @@ pub(crate) type ModuleLoadId = i32;
...
+  /// the storage overhead.
+  Boxed(Box<[u8]>),
+
+  /// Code loaded from the `deno_graph` infrastructure.
+  Arc(Arc<[u8]>),
+}
+
+impl ModuleCodeBytes {
+  pub fn as_bytes(&self) -> &[u8] {
+    match self {
```
bartlomieju, line 64: All of this can be upstreamed in a separate, small PR

core/modules/map.rs
```diff
@@ -436,11 +502,116 @@ impl ModuleMap {
...
 
+  /// Create and compile an ES module.
+  pub(crate) fn new_wasm_module(
+    &self,
+    scope: &mut v8::HandleScope,
+    main: bool,
+    name: ModuleName,
+    source: ModuleCodeString,
+    is_dynamic_import: bool,
+  ) -> Result<ModuleId, ModuleError> {
```
bartlomieju, line 513: This should be abstracted into a separate function and upstreamed in a separate PR. Maybe `new_es_module` should accept a `ModuleType` to define how the module introduces itself.

core/runtime/jsruntime.rs
```diff
@@ -648,6 +652,7 @@ impl JsRuntime {
       // SAFETY: we just asserted that layout has non-0 size.
       unsafe { std::alloc::alloc(layout) as *mut _ };
 
+    let custom_module_evaluation_cb = options.custom_module_evaluation_cb;
     let validate_import_attributes_cb = options
```
bartlomieju, line 656: Discussed with @lucacasonato and we decided we should remove it. It should be up to embedder's `ModuleLoader` to deny loading a module on unknown attributes. Upstream in a separate PR.

core/modules/recursive_load.rs
```diff
@@ -196,7 +196,9 @@ impl RecursiveModuleLoad {
     module_source: ModuleSource,
   ) -> Result<(), ModuleError> {
     let requested_module_type = module_request.requested_module_type.clone();
-    if requested_module_type != module_source.module_type {
+    if requested_module_type == RequestedModuleType::Json
+      && requested_module_type != module_source.module_type
```
bartlomieju, line 200: Ditto - this should be completely removed from here.

core/modules/mod.rs
```diff
@@ -156,22 +215,24 @@ pub(crate) fn get_requested_module_type_from_attributes(
...
+/// For non-`JavaScript` modules, this value suggests
+/// how to interpret the module. For JSON modules it's used to validate
+/// against actually discovered MIME type.
+#[derive(Clone, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
 #[repr(u32)]
 pub enum ModuleType {
   JavaScript,
   Json,
+  /// Non-well-known module type.
+  Other(Cow<'static, str>),
```
bartlomieju, line 227: Upstream in a separate PR along with `CustomModuleEvaluationCb`

wasm_example/import.mjs
```diff
@@ -0,0 +1,6 @@
+import wasmBytes from "./import.wasm" with { type: "bytes" };
```
mmastrac, line 1: This should probably be an integration test

core/runtime/jsruntime.rs
mmastrac: I think ModuleLoader is doing a little bit too much right now TBH -- it does fetching, transpiling and now we'd be adding evaluation.

Module loading is something like a pipeline and we might want to consider how we can structure the callbacks to be more composable. For example, I might want to stack a URL+file fetching component with a TypeScript compilation transpiler, along with wasm/bytes/url module conversion types.

Right now everything is pretty much jammed into module loaders that aren't really re-usable, but we could be building a library of components instead.

bartlomieju: I'm not sure how feasible it is - the crux is that we want loading to be as efficient as possible and it's happening in an async context. Which somewhat forces us to do all of that stuff in a single function. I'll be happy to revisit it, but for now I'd like to get this stuff landed. I can leave this evaluation callback as is and not move it to `ModuleLoader`.

core/modules/recursive_load.rs
bartlomieju: https://github.com/denoland/deno_core/pull/435

core/runtime/jsruntime.rs
bartlomieju: https://github.com/denoland/deno_core/pull/435

core/examples/fs_module_loader.rs
```diff
@@ -1,12 +1,88 @@
...
+fn custom_module_evaluation_cb(
+  scope: &mut v8::HandleScope,
+  module_type: Cow<'_, str>,
+  module_name: &FastString,
+  code: ModuleSourceCode,
+) -> Result<v8::Global<v8::Value>, Error> {
+  match &*module_type {
+    "bytes" => Ok(bytes_module(scope, code)),
+    "text" => text_module(scope, module_name, code),
+    "url" => Ok(url_module(scope, module_name)),
```
petamoriken, line 25: As I mentioned earlier, I think `type: "url"` would violate the spec and would like to remove it.

bartlomieju: @petamoriken, this is still work-in-progress PR and it's just an example in `deno_core`. Deno will not have support for URL imports.

bartlomieju: Removed now.

core/01_core.js
```diff
@@ -358,11 +358,12 @@
...
-      } else if (TypedArrayPrototypeSymbolToStringTag(arg) === "Uint8Array") {
-        `Uint8Array(${TypedArrayPrototypeGetLength(arg)}) [${
-          TypedArrayPrototypeJoin(TypedArrayPrototypeSlice(arg, 0, 10), ", ")
-        }]`;
       }
+      //  else if (TypedArrayPrototypeSymbolToStringTag(arg) === "Uint8Array") {
+      //   `Uint8Array(${TypedArrayPrototypeGetLength(arg)}) [${
+      //     TypedArrayPrototypeJoin(TypedArrayPrototypeSlice(arg, 0, 10), ", ")
+      //   }]`;
+      // }
```
bartlomieju, line 366: This is broken, fix it in a separate PR

core/modules/map.rs
```diff
@@ -324,16 +324,13 @@ impl ModuleMap {
           ))));
         };
 
-        let ctx = JsRealm::get_custom_module_evaluation_ctx(scope);
```
bartlomieju, line 327: Revert https://github.com/denoland/deno_core/pull/528

wasm_example/import.mjs
bartlomieju, line 1: Turn these into a unit test, by updating `StaticModuleLoader` to accept wasm files as well

core/modules/map.rs
```diff
@@ -209,17 +213,17 @@ impl ModuleMap {
...
-  }
-
   pub(crate) fn get_type_by_module(
     &self,
     global: &v8::Global<v8::Module>,
   ) -> Option<ModuleType> {
     self.data.borrow().get_type_by_module(global)
   }
 
+  pub(crate) fn get_name_by_id(&self, id: ModuleId) -> Option<String> {
```
mmastrac, line 223: Some spurious changes in this file?

core/modules/map.rs
```diff
@@ -1617,3 +1670,187 @@ pub fn module_origin<'a>(
...
+    exports: vec![],
+  };
+  let rendered = render_js_wasm_module("./foo.wasm", deps);
+  pretty_assertions::assert_eq!(
+    rendered,
+    r#"import wasmMod from "./foo.wasm" with { type: "$$deno-core-internal-wasm-module" };
+const modInstance = await import.meta.wasmInstantiate(wasmMod);"#,
+  );
+
+  let deps = WasmDeps {
```
mmastrac, line 1771: We could probably store some basic WASM modules in the testdata and simplify these tests.

core/modules/map.rs
```diff
@@ -1617,3 +1670,187 @@ pub fn module_origin<'a>(
...
+        name: "default",
+        index: 6,
+        export_type: wasm_dep_analyzer::ExportType::Function,
+      },
+    ],
+  };
+  let rendered = render_js_wasm_module("./foo.wasm", deps);
+  pretty_assertions::assert_eq!(
+    rendered,
+    r#"import wasmMod from "./foo.wasm" with { type: "$$deno-core-internal-wasm-module" };
```
mmastrac, line 1840: Store as a .out file?

testing/integration/wasm_imports/add.wat
bartlomieju, line 1: Add a comment and instruction how to regenerate `add.wasm` file

core/modules/map.rs
bartlomieju: `WasmDeps` doesn't implement `Serialize`/`Deserialize` so there's no easy way to store intermediate representation. I agree that it could be improved, but maybe it's not a blocker for this PR?

dsherret (approved):
LGTM

PR #407, opened 2024-01-03
 fix(core): clamp V8Slice slice to bounds 

mmastrac:
We should always clamp the expected buffer range to that of the underlying storage.

### DIFF
--- a/testing/checkin/runner/testing.rs
+++ b/testing/checkin/runner/testing.rs
@@ -1,4 +1,8 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+use std::any::Any;
+use std::any::TypeId;
+use std::collections::HashMap;
+
 use deno_core::v8;
 
 #[derive(Default)]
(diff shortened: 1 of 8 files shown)


testing/checkin/runner/testing.rs
```diff
@@ -10,3 +14,23 @@ pub struct Output {
 pub struct TestFunctions {
   pub functions: Vec<(String, v8::Global<v8::Function>)>,
 }
+
+#[derive(Default)]
+pub struct TestData {
+  pub data: HashMap<(String, TypeId), Box<dyn Any>>,
+}
```
bartlomieju, line 21: Nitpick: I think you could use `GothamState` from `deno_core` directly?

bartlomieju (approved):

testing/checkin/runner/testing.rs
mmastrac: I don't think so as this has a name/typeid pair, while `GothamState` is just typeid.

PR #414, opened 2024-01-04
perf(core): use FuturesUnordered instead of JoinSet

mmastrac:
Implements a new `OpDriver` based on `FuturesUnordered`.

As part of this work, we create a trait that allows us to test `OpDriver` without V8. This part of the code is a bit awkward, but we can likely remove it at a later date once the performance work slows down.

Other changes:

 - Add `OpScheduling` to simplify the lazy/deferred/eager traits.
 - Modify the batch size of dispatched async ops to 1024 to prevent stack overflow errors
 - Keeps errors in `anyhow::Error` until we're ready to dispatch them

Before:

```
test bench_op_async_void_deferred         ... bench:   1,034,629 ns/iter (+/- 74,835)
test bench_op_async_void_deferred_nofast  ... bench:   1,034,527 ns/iter (+/- 84,364)
test bench_op_async_void_deferred_return  ... bench:   1,031,853 ns/iter (+/- 66,991)
test bench_op_async_void_lazy             ... bench:     993,310 ns/iter (+/- 30,881)
test bench_op_async_void_lazy_nofast      ... bench:   1,011,262 ns/iter (+/- 37,490)
test bench_op_async_yield                 ... bench:   1,037,248 ns/iter (+/- 44,996)
test bench_op_async_yield_deferred        ... bench:   1,052,324 ns/iter (+/- 36,561)
test bench_op_async_yield_deferred_nofast ... bench:   1,048,557 ns/iter (+/- 55,217)
test bench_op_async_yield_lazy            ... bench:   1,546,403 ns/iter (+/- 83,743)
test bench_op_async_yield_lazy_nofast     ... bench:   1,541,818 ns/iter (+/- 51,707)
```

After:

```
test bench_op_async_void_deferred         ... bench:     799,585 ns/iter (+/- 21,081)
test bench_op_async_void_deferred_nofast  ... bench:     790,412 ns/iter (+/- 55,422)
test bench_op_async_void_deferred_return  ... bench:     793,884 ns/iter (+/- 24,304)
test bench_op_async_void_lazy             ... bench:     760,932 ns/iter (+/- 28,006)
test bench_op_async_void_lazy_nofast      ... bench:     759,421 ns/iter (+/- 17,968)
test bench_op_async_yield                 ... bench:     780,460 ns/iter (+/- 16,812)
test bench_op_async_yield_deferred        ... bench:     799,123 ns/iter (+/- 17,069)
test bench_op_async_yield_deferred_nofast ... bench:     791,680 ns/iter (+/- 39,701)
test bench_op_async_yield_lazy            ... bench:     994,607 ns/iter (+/- 19,175)
test bench_op_async_yield_lazy_nofast     ... bench:   1,000,546 ns/iter (+/- 28,675)
```


core/modules/tests.rs
```diff
@@ -659,54 +657,6 @@ fn test_validate_import_attributes_default() {
   }
 }
 
-#[tokio::test]
```
mmastrac, line 662: Moved to integration test

core/runtime/tests/ops.rs
```diff
@@ -354,41 +353,6 @@ fn ops_in_js_have_proper_names() {
   runtime.execute_script_static("test", src).unwrap();
 }
 
-/// Dipatch more promises than the ring can hold, but wait for them all to queue up first.
```
mmastrac, line 357: Moved to integration test

core/runtime/op_driver/mod.rs
```diff
@@ -45,24 +79,283 @@ pub(crate) trait OpDriver: Default {
...
-    Result<v8::Local<'s, v8::Value>, v8::Local<'s, v8::Value>>,
-  )>;
+  ) -> Poll<(PromiseId, OpId, OpResult<C>)>;
 
+  /// Return the number of futures currently being polled.
   fn len(&self) -> usize;
 }
+
+#[cfg(test)]
+mod tests {
```
mmastrac, line 122: These test that the two drivers work identically.

core/modules/tests.rs
bartlomieju: I'm not sure I'm a fan of that - now we got module loading tests split between `core/modules` and `testing/checking/integration`. Can you keep it here until I land `itest!` macro and comparing `.out` files in `deno_core`?

core/runtime/jsruntime.rs
```diff
@@ -670,10 +670,26 @@ impl JsRuntime {
...
+      .collect::<Vec<_>>();
+    let ops_with_metrics = op_metrics_fns
+      .iter()
+      .map(|o| o.is_some())
+      .collect::<Vec<_>>();
+    let context_state = Rc::new(ContextState::new(
+      isolate_ptr,
+      options.get_error_class_fn.unwrap_or(&|_| "Error"),
+      ops_with_metrics,
+    ));
```
bartlomieju, line 692: So does that mean that now metrics are always collected using the new mechanism? Could we abstract it away in `ContextState` maybe?

core/runtime/jsruntime.rs
```diff
@@ -2105,27 +2120,36 @@ impl JsRuntime {
...
-    // allocations.
     let mut args: SmallVec<[v8::Local<v8::Value>; 32]> =
       SmallVec::with_capacity(32);
 
     loop {
-      let Poll::Ready((promise_id, op_id, metrics_event, res)) =
-        context_state.pending_ops.poll_ready(cx, scope)
+      if args.len() >= MAX_VEC_SIZE_FOR_OPS {
+        // We have too many, bail for now but re-wake the waker
+        cx.waker().wake_by_ref();
```
bartlomieju, line 2140: Why do you bail out here and re-wake? Can't we return a batch immediately?

core/runtime/op_driver/futures_unordered_driver.rs
```diff
@@ -0,0 +1,349 @@
...
+  fn pending_op_success<R: 'static>(
+    info: PendingOpInfo,
+    rv_map: C::MappingFn<R>,
+    v: R,
+  ) -> PendingOp<C> {
+    PendingOp(info, OpResult::new_value(v, rv_map))
+  }
+
+  #[inline(always)]
+  fn pending_op_failure<E: Into<Error> + 'static>(
```
bartlomieju, line 155: Are these two methods just to satisfy a trait? This seems super verbose, instead of inlining in `pending_op_result`. Maybe they're here to enforce types?

core/runtime/op_driver/futures_unordered_driver.rs
```diff
@@ -0,0 +1,349 @@
...
+  type Output;
+
+  fn len(&self) -> usize;
+  fn spawn(&mut self, f: Self::Future);
+  fn poll_next_unpin(&mut self, cx: &mut Context) -> Poll<Self::Output>;
+}
+
+pub struct SubmissionQueueResults<F: SubmissionQueueFutures> {
+  queue: Rc<Queue<F>>,
+}
```
bartlomieju, line 313: This is really clever

core/runtime/op_driver/mod.rs
```diff
@@ -9,31 +8,66 @@ use std::task::Poll;
...
 
 /// `OpDriver` encapsulates the interface for handling operations within Deno's runtime.
 ///
 /// This trait defines methods for submitting ops and polling readiness inside of the
 /// event loop.
-pub(crate) trait OpDriver: Default {
+///
+/// The driver takes an optional [`OpMappingContext`] implementation, which defaults to
+/// one compatible with v8. This is used solely for testing purposes.
+pub(crate) trait OpDriver<C: OpMappingContext = V8OpMappingContext>:
```
bartlomieju, line 39: I think that complicates code a lot - in a follow up let's figure out another way to test this so we don't have to pass `OpMappingContext` everywhere.

core/runtime/op_driver/mod.rs
```diff
@@ -45,24 +79,283 @@ pub(crate) trait OpDriver: Default {
...
+  }
+
+  #[rstest]
+  #[case::joinset(JoinSetDriver::<TestMappingContext>::default())]
+  #[case::futures_unordered(FuturesUnorderedDriver::<TestMappingContext>::default())]
+  fn test_driver<D: OpDriver<TestMappingContext>>(
+    #[case] driver: D,
+    #[values(2, 16)] count: usize,
+    #[values(OpScheduling::Eager, OpScheduling::Lazy, OpScheduling::Deferred)]
+    scheduling: OpScheduling,
```
bartlomieju, line 222: Side note: I really don't like all this macro magic from `rstest` 😅 

That said, these tests are great.

testing/checkin/runtime/async.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+const { op_async_barrier_create, op_async_barrier_await, op_async_yield } = Deno
+  .core
+  .ensureFastOps();
```
bartlomieju, line 4: Side issue: let's rename `Deno.core.ensureFastOps()` to `Deno.core.generateOpBindings()`?

bartlomieju (approved):

core/modules/tests.rs
bartlomieju: Meh, I guess that's fine actually.

mmastrac: For sure -- I'll eventually be moving almost all of these into integration tests and we can tweak how they run at that point. There's nothing in the module tests that should prevent all of them being integration tests eventually.

core/runtime/jsruntime.rs
mmastrac: Nope, this just means we keep a vec of booleans so we can look up which ops have metrics enabled. This means we no longer need to feed the `metrics_enabled` flag through to the op driver as we can check for it when polling the event loop. 

I just realized I missed something here - we are double-collecting the metrics functions. I'll fix that before landing.

core/runtime/jsruntime.rs
mmastrac: We do return a batch immediately -- it just breaks out of the loop allowing the batch of 512 ops to be resolved. We pick up the next batch on the next time through the loop.

core/runtime/op_driver/futures_unordered_driver.rs
mmastrac: It's a bit verbose, I agree. I removed OpError from the driver which makes these kind of redundant now. The previous code was actually a little hairy and repetitive. I'll fix before landing.

core/runtime/jsruntime.rs
bartlomieju: I see, looks like I didn't read through precisely enough :)

core/runtime/op_driver/futures_unordered_driver.rs
bartlomieju: Thank you

core/runtime/op_driver/mod.rs
mmastrac: It would be great to remove this somehow, but I haven't figured out a way to do this yet. What really kills us is that this code needs to run under Miri :/

core/runtime/op_driver/mod.rs
mmastrac: Under the hood it just runs a combination of all those inputs which actually reduces the amount of macro code we need to effectively test all of these cases 😓 

Even though these macros are a bit noisy, the real value is that we can add a lot of code coverage with only a few lines of testing code.

testing/checkin/runtime/async.ts
mmastrac: I think we should bite the bullet and do https://github.com/denoland/deno_core/issues/223 which gets rid of this method. I never really liked having to ask for ops this way.

What do you think?

bartlomieju: Yes, that sounds like a better solution

core/runtime/op_driver/futures_unordered_driver.rs
mmastrac: Fixed

core/runtime/jsruntime.rs
mmastrac: Fixed.

Issue #412, opened 2024-01-04
Consider adding "bin" target


bartlomieju:
Currently `deno_core` is just a library (that means it cannot be built as a standalone binary tool).

Given all the `examples/` and work such as #411 or #402 I think it would make sense to provide a binary target for "deno_core". This would be a very similar situation to [V8's `d8`](https://v8.dev/docs/d8) which is a developer shell that allows to interact with engine. This would allow us to test various new changes to `deno_core` in an easy way as well as provide a useful tool for benchmarking different scenarios.

PR #415, opened 2024-01-04
[WIP] Add REPL example

### LINKED ISSUE #411: Add REPL example
Life the least amount of code possible from https://github.com/denoland/deno/tree/main/cli/tools/repl and provide a simple REPL example that can be used to test out code.

bartlomieju:
Closes #411


core/cdp.rs
```diff
@@ -0,0 +1,528 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+use deno_core::serde::Deserialize;
```
bartlomieju, line 3: Upstream it in a separate PR

core/examples/repl.rs
```diff
@@ -0,0 +1,465 @@
...
+}
+
+impl std::fmt::Display for EvaluationOutput {
+  fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    match self {
+      EvaluationOutput::Value(value) => f.write_str(value),
+      EvaluationOutput::Error(value) => f.write_str(value),
+    }
+  }
+}
```
ry, line 108: seems excessive for an example

bartlomieju: It's copy-pasted from `cli/tools/repl`. I'm trying to slim it down as much as possible so we can turn it into a runtime API to provide `node:repl` polyfill.

bartlomieju: Plus see #412 - Matt and I have been discussing this and we want to do a bin target to be able to test changes quickly without having to rebuilt `deno` itself (because iterating on `deno_core` is waaaaaay faster).

PR #418, opened 2024-01-05
fix(core): snapshot crash in windows

mmastrac:
Needs confirmation that this fixes things.


littledivy (approved):
Let's try it?

bartlomieju:
Can we land it?

Issue #417, opened 2024-01-05
fast calls generated for `#[anybuffer]` may create new V8 handles


littledivy:
Usage of `to_slice_buffer_any` in fast calls crashes tests with latest V8 (checkout `main` rusty_v8):

```
#
# Fatal error in v8::HandleScope::CreateHandle()
# Cannot create a handle without a HandleScope
#

Process 25068 stopped
* thread #14, name = 'runtime::ops::tests::test_op_buffer_any_length', stop reason = EXC_BREAKPOINT (code=1, subcode=0x100da39b8)
    frame #0: 0x0000000100da39b8 deno_core-09c55473f92ab90c`::Abort() at platform-posix.cc:699:7 [opt]
warning: deno_core-09c55473f92ab90c was compiled with optimization - stepping may behave oddly; variables may not be available.
(lldb) up
frame #1: 0x0000000100e4b468 deno_core-09c55473f92ab90c`::ReportApiFailure() at api.cc:322:5 [opt]
(lldb) up
frame #2: 0x0000000101178978 deno_core-09c55473f92ab90c`::Extend() at handles.cc:154:8 [opt]
(lldb) up
frame #3: 0x0000000100e4bdb4 deno_core-09c55473f92ab90c`::CreateHandle() at handles-inl.h:265:14 [opt]
(lldb) up
frame #4: 0x00000001015cf574 deno_core-09c55473f92ab90c`::GetBuffer() at js-array-buffer.cc:262:24 [opt]
(lldb) up
frame #5: 0x0000000100e7e990 deno_core-09c55473f92ab90c`::Buffer() at api.cc:8990:56 [opt]
(lldb) up
frame #6: 0x0000000100d64590 deno_core-09c55473f92ab90c`::v8__ArrayBufferView__Buffer__Data() at binding.cc:1676:31
(lldb) up
frame #7: 0x00000001003ea418 deno_core-09c55473f92ab90c`deno_core::runtime::ops::to_slice_buffer_any::h7d199eeb53a54155 [inlined] v8::array_buffer_view::_$LT$impl$u20$v8..data..ArrayBufferView$GT$::data::hcc1c7920a8140359(self=0x0000000170d57938) at array_buffer_view.rs:54:7
   51  	  #[inline(always)]
   52  	  pub fn data(&self) -> *mut c_void {
   53  	    unsafe {
-> 54  	      v8__ArrayBufferView__Buffer__Data(self)
[...]

PR #419, opened 2024-01-05
fix(core): disable fastcalls for typed buffers

mmastrac:
With the upcoming v8 upgrade, we cannot support fastcalls for typed buffers.

Fix for #417


ops/op2/dispatch_fast.rs
```diff
@@ -730,13 +730,21 @@ fn map_arg_to_v8_fastcall_type(
...
       BufferType::Slice(.., NumericArg::u8)
       | BufferType::Ptr(.., NumericArg::u8),
       _,
       BufferSource::Any,
-    ) => V8FastCallType::AnyArray,
+      // This cannot be fast at this time as we have no way of accessing
+      // the shared pointer to the backing store in a fastcall.
+      // https://github.com/denoland/deno_core/issues/417
+      // ) => V8FastCallType::AnyArray,
+    ) => return Ok(None),
```
littledivy, line 747: Can we partial support this with `V8FastCallType::Uint8Array`? Fast calls have native support for it

mmastrac: I think we definitely could but I'm not sure I have the bandwidth right now :/

I mostly need to unblock rusty_v8 so I can get some work in there.

littledivy (approved):

PR #424, opened 2024-01-08
Upgrade rusty_v8 to 0.83.0

littledivy:
"change-array-by-copy" is enabled by [default](https://chromium-review.googlesource.com/c/v8/v8/+/5035578)


mmastrac (approved):
LGTM

PR #425, opened 2024-01-08
[WIP] destrucutre AggregateError

crowlKats:
The second test, which uses `destructureError` (underlying op being `op_destructure_error`), adds a frame for unknown reasons.

### DIFF
--- a/core/runtime/tests/error.rs
+++ b/core/runtime/tests/error.rs
@@ -85,7 +85,7 @@ if (errMessage !== "higher-level sync error: original sync error") {
       r#"
 (async () => {
   let errMessage;
-  const { op_err_async } = Deno.core.ensureFastOps(); 
+  const { op_err_async } = Deno.core.ensureFastOps();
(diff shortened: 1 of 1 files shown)


core/runtime/tests/error.rs
```diff
@@ -115,3 +115,53 @@ fn syntax_error() {
...
+
+#[tokio::test]
+async fn aggregate_error() {
+  let mut runtime = JsRuntime::new(Default::default());
+  let src = r#"
+(async () => {
+  await Promise.any([]);
+})()
+"#;
+  let value_global = runtime
```
mmastrac, line 127: Any chance these could be integration and/or unit tests? I'd like to avoid adding more of these tests which are a bit awkward to maintain in Rust.

bartlomieju:
@littledivy did more digging into this problem and our current understanding is that it's a V8 problem - not sure if it's worth investing resources right now to fix it.

PR #428, opened 2024-01-08
feat(core): better diagnostics when extensions fail to load

mmastrac:


bartlomieju (approved):
LGTM once lint error is resolved.

PR #433, opened 2024-01-10
feat: create ops virtual module for each extension

bartlomieju:
This commit adds a "virtual ops module" for each extension with ops.

All the enabled ops provided by the module can be imported like so:
```
import { op_name1, op_name2 } from "ext:<name>/ops";
```

Closes https://github.com/denoland/deno_core/issues/223


core/extensions.rs
```diff
@@ -586,6 +586,51 @@ impl Extension {
...
+      .collect();
+
+    let mut output = Vec::with_capacity(op_names.len() * 2 + 4);
+    output.push("const {".to_string());
+
+    for op_name in &op_names {
+      output.push(format!("  {},", op_name));
+    }
+
+    output.push("} = Deno.core.ensureFastOps(true);".to_string());
```
bartlomieju, line 617: Ideally we could remove this `true`, but I think we're gonna face many crashes before we migrate all the code to use the new system.

bartlomieju: Also, it's unclear what would be interaction with enabled/disabled ops.

bartlomieju: Eg. if I remove the `true` I get failures in `testing/unit/ops_buffer_test.ts` which also uses `ensureFastOps(true)`. But I can use `import { ... } from "ext:/...` there because it's a test file that will fail a check in module loader. 

Ideas?

mmastrac: > Also, it's unclear what would be interaction with enabled/disabled ops.

I think we need to throw from disabled ops as there's no way to have an optional import is there?



> Eg. if I remove the `true` I get failures in `testing/unit/ops_buffer_test.ts` which also uses `ensureFastOps(true)`. But I can use `import { ... } from "ext:/...` there because it's a test file that will fail a check in module loader.
> 
> Ideas?

I think we'll just need to reexport here:

What about this?

```
import * as ops from "ext:...";
globalThis.testOps = ops;
```

bartlomieju:
A few more things to consider:
1. What is we have an op (`op_foo`) that is defined in `foo` ext in `deno` repo, but then an embedder disables that op and provides another extension that defines that op (is that actually even possible?) - in such case the origin of the op is different and JS code that imports the op function would crash
2. Still not sure what to do with enabled/disabled ops

core/extensions.rs
bartlomieju: This will become a non-issue once we land https://github.com/denoland/deno_core/pull/456

core/modules/loaders.rs
```diff
@@ -118,6 +118,12 @@ pub(crate) struct ExtModuleLoader {
 impl ExtModuleLoader {
   pub fn new(extensions: &[Extension]) -> Self {
     let mut sources = HashMap::new();
+    sources.extend(
+      extensions
+        .iter()
+        .flat_map(|e| e.get_ops_virtual_module())
+        .map(|s| (s.specifier.to_string(), s.clone())),
+    );
```
bartlomieju, line 126: Should be gone once we start using "new_synthetic_module"

bartlomieju:
Discussed offline with @mmastrac and we're going to start with a single "vritual module" for all ops. We need to measure impact on having multiple modules on startup time and memory usage.

littledivy (commented):
Should be ready for review. Summary of the changes I made:

- Store C-fn mapped fn pointer of the synthetic module eval function in external references table.
- Shortcut `get_stalled_top_level_await_message` when module is synthetic.
- Update internal module count assumptions from `2`.

core/modules/map.rs
```diff
@@ -247,8 +247,7 @@ impl ModuleMap {
     self.data.borrow().is_alias(name, requested_module_type)
   }
 
-  #[cfg(test)]
-  pub fn get_data(&self) -> &RefCell<ModuleMapData> {
+  pub(crate) fn get_data(&self) -> &RefCell<ModuleMapData> {
```
bartlomieju, line 250: Revert

core/runtime/bindings.rs
```diff
@@ -102,7 +110,7 @@ pub fn script_origin<'a>(
   )
 }
 
-fn get<'s, T>(
+pub(crate) fn get<'s, T>(
```
bartlomieju, line 113: Revert

core/runtime/bindings.rs
```diff
@@ -287,7 +295,7 @@ pub(crate) fn initialize_context<'s>(
   }
 }
 
-fn op_ctx_function<'s>(
+pub fn op_ctx_function<'s>(
```
bartlomieju, line 298: Revert

core/runtime/bindings.rs
```diff
@@ -698,3 +706,47 @@ where
...
+
+pub fn get_exports_for_ops_virtual_module<'s>(
+  op_ctxs: &[OpCtx],
+  scope: &mut v8::HandleScope<'s>,
+  global: v8::Local<v8::Object>,
+) -> Vec<(FastString, v8::Local<'s, v8::Value>)> {
+  let mut exports = Vec::with_capacity(op_ctxs.len());
+
+  // TODO(bartlomieju): duplicated in `bindings.rs`
+  // TODO(bartlomieju): called for every single extension - can be optimized away
```
bartlomieju, line 718: Not correct

core/runtime/bindings.rs
```diff
@@ -698,3 +706,47 @@ where
...
 }
+
+pub fn get_exports_for_ops_virtual_module<'s>(
+  op_ctxs: &[OpCtx],
+  scope: &mut v8::HandleScope<'s>,
+  global: v8::Local<v8::Object>,
+) -> Vec<(FastString, v8::Local<'s, v8::Value>)> {
+  let mut exports = Vec::with_capacity(op_ctxs.len());
+
+  // TODO(bartlomieju): duplicated in `bindings.rs`
```
bartlomieju, line 717: Remove

core/runtime/jsruntime.rs
```diff
@@ -1024,6 +1027,42 @@ impl JsRuntime {
...
+          get_exports_for_ops_virtual_module(&op_ctxs, scope, global);
+        let mod_id = module_map
+          .new_synthetic_module(
+            scope,
+            FastString::Owned("ext:core/ops".into()),
+            crate::ModuleType::JavaScript,
+            synthetic_module_exports,
+          )
+          .unwrap();
+        module_map.get_data().borrow_mut().next_load_id += 1;
```
bartlomieju, line 1055: Not needed

core/runtime/jsruntime.rs
```diff
@@ -1024,6 +1027,42 @@ impl JsRuntime {
...
     futures::executor::block_on(async {
+      // TODO(bartlomieju): this is somewhat duplicated in `bindings::initialize_context`,
+      // but for migration period we need to have ops available in both `Deno.core.ops`
+      // as well as have them available in "virtual ops module"
+      // if !matches!(
+      //   self.init_mode,
+      //   InitMode::FromSnapshot {
+      //     skip_op_registration: true
+      //   }
+      // ) {
```
bartlomieju, line 1038: Most likely this conditional should be used

core/modules/module_map_data.rs
```diff
@@ -516,11 +515,11 @@ impl ModuleMapData {
...
     let data = self;
-    // There's always one internal `deno_core` ES module loaded, so +1 here.
-    assert_eq!(data.handles.len(), modules.len() + 1);
-    assert_eq!(data.info.len(), modules.len() + 1);
-    assert_eq!(data.next_load_id as usize, modules.len());
-    assert_eq!(data.by_name.len(), modules.len() + 1);
+    // There's always two internal `deno_core` ES module loaded, so +2 here.
+    assert_eq!(data.handles.len(), modules.len() + 2);
+    assert_eq!(data.info.len(), modules.len() + 2);
+    assert_eq!(data.next_load_id as usize, modules.len() + 1);
```
littledivy, line 521:
```suggestion
    assert_eq!(data.next_load_id as usize, modules.len());
```

bartlomieju: I'm not sure about that. I think we're leaking some detail here. I want to investigate this `next_load_id` before I merge this PR.

littledivy: IIUC `next_load_id` is only tracked for non internal modules.

core/modules/module_map_data.rs
```diff
@@ -516,11 +515,11 @@ impl ModuleMapData {
   #[cfg(test)]
   pub fn assert_module_map(&self, modules: &Vec<ModuleInfo>) {
     let data = self;
-    // There's always one internal `deno_core` ES module loaded, so +1 here.
-    assert_eq!(data.handles.len(), modules.len() + 1);
-    assert_eq!(data.info.len(), modules.len() + 1);
+    // There's always two internal `deno_core` ES module loaded, so +2 here.
+    assert_eq!(data.handles.len(), modules.len() + 2);
```
mmastrac, line 519: Can we make this a `const`?

core/runtime/jsruntime.rs
```diff
@@ -1052,6 +1055,41 @@ impl JsRuntime {
...
+        let context_local = v8::Local::new(scope, context_global);
+        let context_state = JsRealm::state_from_scope(scope);
+        let op_ctxs = context_state.op_ctxs.borrow();
+        let global = context_local.global(scope);
+        let synthetic_module_exports =
+          get_exports_for_ops_virtual_module(&op_ctxs, scope, global);
+        let mod_id = module_map
+          .new_synthetic_module(
+            scope,
+            FastString::Owned("ext:core/ops".into()),
```
mmastrac, line 1078: FastString::StaticAscii?

core/runtime/tests/snapshot.rs
```diff
@@ -147,8 +147,8 @@ fn es_snapshot() {
       )
       .unwrap()
     };
-    // There's always one internal `deno_core` ES module loaded, so +1 here.
-    assert_eq!(i + 1, id);
+    // There's always two internal `deno_core` ES module loaded, so +2 here.
+    assert_eq!(i + 2, id);
```
mmastrac, line 151: `const`

testing/checkin/runtime/console.ts
```diff
@@ -1,8 +1,6 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
-const {
-  op_log_debug,
-  op_log_info,
-} = Deno.core.ensureFastOps();
+
+import { op_log_debug, op_log_info } from "ext:core/ops";
```
mmastrac, line 3: ❤️

mmastrac (approved):
LGTM once build passes.

PR #435, opened 2024-01-11
feat: Remove default import attributes and JSON modules validation

bartlomieju:
This commit removes default "import attributes" validation as
well as verification for JSON modules. It is now embedder's
responsibility to check if the "content type" of the module
matches "type": "json" attribute (if the embedder wants to do
that).

This is done in preparation to allowing embedder's to load 
and execute any module type (eg. "bytes", "text", etc).


mmastrac (approved):
LGTM

PR #438, opened 2024-01-11
refactor: Add ModuleMap::new_module_from_js_source

bartlomieju:
This commit adds "ModuleMap::new_module_from_js_source" method.

This method is pretty much the same as "ModuleMap::new_es_module", but allows to customize the 
type of the module - ie. it allows to create a module from JS source with any type - this allows for "shim"
modules. One might create a JS "shim" that acts as a proxy to a WASM module.

A drive by cleanup of "ModuleError" was done as well.


mmastrac (approved):
LGTM, nice cleanups

PR #439, opened 2024-01-12
feat: add 'requested_module_type' to ModuleLoader::load

bartlomieju:
This commit adds additional argument to "ModuleLoader::load" method that
is "RequestedModuleType". This argument can be used by the embedder to 
decide if the loaded module should be refused (eg. trying to load a module
with type "json", but the actual MIME type is not "application/json").

In a follow up I might extract `is_dyn_import` and `requested_module_type`
into a helper struct (maybe `LoadMetadata`).

### DIFF
--- a/core/modules/map.rs
+++ b/core/modules/map.rs
@@ -1459,7 +1459,9 @@ impl ModuleMap {
 
     let specifier = ModuleSpecifier::parse(module_specifier)?;
     let source = futures::executor::block_on(async {
-      loader.load(&specifier, None, false).await
+      loader
+        .load(&specifier, None, false, RequestedModuleType::None)
+        .await
     })?;
 
     self.lazy_load_es_module_from_code(
--- a/core/modules/map.rs
+++ b/core/modules/map.rs
@@ -279,6 +279,12 @@ impl ModuleMap {
       module_url_specified
     };
 
+    // TODO(bartlomieju): I have a hunch that this is wrong - write a test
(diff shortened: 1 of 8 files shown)


core/modules/map.rs
```diff
@@ -279,6 +279,12 @@ impl ModuleMap {
       module_url_specified
     };
 
+    // TODO(bartlomieju): I have a hunch that this is wrong - write a test
+    // that tries to "confuse" the type system, by first requesting a module
+    // with type `RequestedModuleType::Other("foo".into)``, and then the loader
+    // actually returns `ModuleType::Other("bar".into())`. See if it leads to
+    // unexpected result in how `ModuleMap` is structured and verify how
+    // querying the module map works (`ModuleMap::get_by_id`, `ModuleMap::get_by_name`).
     let requested_module_type = RequestedModuleType::from(module_type.clone());
```
mmastrac, line 288: Doesn't the recursive load code check this?

```
    if requested_module_type != module_source.module_type {
      return Err(ModuleError::Other(generic_error(format!(
        "Expected a \"{}\" module but loaded a \"{}\" module.",
        requested_module_type, module_source.module_type,
      ))));
    }
```

mmastrac (approved):
LGTM

core/modules/map.rs
bartlomieju: That code was removed in https://github.com/denoland/deno_core/commit/7fd84216ffdcb331f4ba38ca133b5c2ce31d0c3c.

PR #441, opened 2024-01-13
feat: Add import.meta.filename and import.meta.dirname

bartlomieju:
This commit adds two new properties to "import.meta" object:
- "import.meta.filename"
- "import.meta.dirname"

These properties are defined only for "file:///..." URLs and are
not present for any other schemes.


core/runtime/bindings.rs
```diff
@@ -356,6 +356,34 @@ pub extern "C" fn host_initialize_import_meta_object_callback(
...
+    let dirname = name
+      .strip_prefix("file://")
+      .unwrap()
+      .rsplit_once("/")
+      .map(
+        |(dirname, _)| {
+          if dirname.is_empty() {
+            "/"
+          } else {
+            dirname
```
dsherret, line 376: We need to add windows support here

bartlomieju: I believe this is unreachable on Windows? There's gonna be something like `file://C:/foo/bar/something.js`

dsherret: Sorry, I just meant in general (unless I'm misreading this). For example, something like `file:///C:/foo/bar/something.js` (note three forward slashes) would give `C:\foo\bar\something.js` similar to in Node.

I believe the output should be the same as:

```ts
import path from 'node:path';
import { fileURLToPath } from 'node:url';

const __filename = fileURLToPath(import.meta.url);
const __dirname = path.dirname(__filename);
```

bartlomieju: I see - so you want to change the file URL to be an actual path (with backwards slashes). Do you happen to know if we already have this utility somewhere in `deno`?

dsherret: Yeah, I think we'd need to parse it to a url then use `to_file_path()`

Just tried this out to confirm in Node:

```shellsession
> cat test.mjs
console.log(import.meta.filename);
console.log(import.meta.dirname);
> node test.mjs
V:\scratch\test.mjs
V:\scratch
```

bartlomieju: Thanks, I'll fix it in a moment

dsherret: That said, I kind of wonder if we could just do our own thing here without using the url crate.... might not be too hard and it would save us having to parse the file url, which would be faster. https://github.com/servo/rust-url/blob/f447500049cbfe42203ff16026842e74f3fd99ac/url/src/lib.rs#L2598-L2611

bartlomieju: I think it's not a big deal, this is called only once per module.

core/runtime/bindings.rs
```diff
@@ -368,6 +369,31 @@ pub extern "C" fn host_initialize_import_meta_object_callback(
...
+    let filename_key =
+      v8::String::new_external_onebyte_static(scope, b"filename").unwrap();
+    let dirname_key =
+      v8::String::new_external_onebyte_static(scope, b"dirname").unwrap();
+    // Use display() here so that Rust takes care of proper forward/backward slash
+    // formatting depending on the Os.
+    let filename_val =
+      v8::String::new(scope, &file_path.display().to_string()).unwrap();
+
+    let dir_path = file_path.parent().unwrap();
```
bartlomieju, line 390: This is probably wrong for URL like `file:///some_file.js` - ie. it works fine on Unix, but on Windows I'm not sure what we should do here 🤔

dsherret: A module wouldn't have that kind of URL on windows so that's ok.

bartlomieju: This is probably only a problem for testing code - should I just default to `/` here?

dsherret: In the test code we should do `file:///C:/some_file.js` on Windows.

bartlomieju: That would require changing pretty much all tests - can I just default to `/` in such case?

core/runtime/bindings.rs
```diff
@@ -368,6 +370,34 @@ pub extern "C" fn host_initialize_import_meta_object_callback(
   let resolve_key =
     v8::String::new_external_onebyte_static(scope, b"resolve").unwrap();
   meta.set(scope, resolve_key.into(), val.into());
+
+  // For `file:` URL we provide additional `filename` and `dirname` values
+  let Ok(name_url) = Url::parse(&name) else {
+    return;
```
dsherret, line 376: Nitpick: Using returns like this is great when the function does one thing, but this function does a lot (it can lead to bugs if someone goes and adds something to the bottom of this function that should still work when this returns). Maybe we should extract the code below to a new function or perhaps we should just nest the if statements?

bartlomieju: Done.

dsherret (approved):
LGTM!

core/runtime/bindings.rs
```diff
@@ -368,6 +369,45 @@ pub extern "C" fn host_initialize_import_meta_object_callback(
...
+  let dirname_key =
+    v8::String::new_external_onebyte_static(scope, b"dirname").unwrap();
+  // Use display() here so that Rust takes care of proper forward/backward slash
+  // formatting depending on the OS.
+  let filename_val =
+    v8::String::new(scope, &file_path.display().to_string()).unwrap();
+
+  assert!(file_path.pop());
+  let dir_path = file_path;
+  let dirname_val =
```
bartlomieju, line 407: @dsherret I think this is the problematic part - for `file:///C:/some_file.js` - `pop()` here will return nothing.

PR #442, opened 2024-01-14
feat: Change CustomModuleEvaluationCb to return an enum

bartlomieju:
This commit changes "CustomModuleEvaluationCb" to return an enum
that tells how to interpret the return value.

In addition to evaluating the returned value as a synthetic module like before
(using "CustomModuleEvaluationKind::Synthetic"), now users can evaluate a
synthetic module and a "computed" module that uses the synthetic module
(using "CustomModuleEvaluationKind::ComputedAndSynthetic").


dsherret (approved):
RSLGTM

PR #443, opened 2024-01-14
feat(example): add 'text' and 'bytes' imports to fs_module_loader.rs

bartlomieju:
A little cleanup and showcase how to wire up custom
module types in the `fs_module_loader.rs` example.

Ref #402


dsherret (approved):
LGTM

Issue #445, opened 2024-01-17
`initialize_context` rewrite


bartlomieju:
After several discussion with @mmastrac we established that `initialize_context` needs to be rewritten.

This function does several things, including codegen that is so unreadable and hard to reason about
that it makes it a hell to maintain.

From what I gather right now:
- `initialize_context` is initialized as first bit of JS code after a `JsRuntime` is created
- it includes `core/bindings.js` file as part of the codegen
- it sets up `Deno.core` namespace with `op` and asyncOps` objects - this can be done using native V8 APIs
- if this is a "new" context (ie. one not created from a snapshot), a call to `Deno.__op__console` is made to attach `callConsole` and `console` bindings to `Deno.core.callConsole` and `Deno.core.console` variables - this can be done using native V8 APIs
- For each to be available it calls `Deno.__op__registerOp` (which is part of `bindings.js`) that properly sets up an op binding in JS:
   -  for sync ops, the function is put into `Deno.core.ops`
   - for async ops, the function is temporarily stored in `Deno.core.asyncOps`, then calls `Deno.core.asyncStub` on first use (which is done by `Deno.core.ensureFastOps()`), that grabs the op from `Deno.core.asyncOps`, generated an optimized function (based on no of arguments) and puts the op in `Deno.core.ops`
 - Then initializes everything by calling `Deno.__op__` that registers each op, and then does a cleanup of all the `Deno.__*` APIs that were created during `initialize_context`

Example codegened file (from one of the test in `deno_core`):
```js
// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
if (!globalThis.Deno) {
  globalThis.Deno = {
    core: {
      ops: {},
      asyncOps: {},
    },
  };
}

Deno.__op__console = function (callConsole, console) {
  Deno.core.callConsole = callConsole;
  Deno.core.console = console;
};

Deno.__op__registerOp = function (isAsync, op, opName) {
  const core = Deno.core;
  if (isAsync) {
    if (core.ops[opName] !== undefined) {
      return;
    }
    core.asyncOps[opName] = op;
    const fn = function (...args) {
      if (this !== core.ops) {
        // deno-lint-ignore prefer-primordials
        throw new Error(
          "An async stub cannot be separated from Deno.core.ops. Use ???",
        );
      }
      return core.asyncStub(opName, args);
    };
    fn.name = opName;
    core.ops[opName] = fn;
  } else {
    core.ops[opName] = op;
  }
};

[...]

PR #448, opened 2024-01-17
refactor: simplify setup of JS bindings for async ops

bartlomieju:
This commit greatly simplifies how ops are bound into JS functions.

When a new context if created, for each op we're binding it to
`Deno.core.ops[<name>]` - additionally doing a set up for async ops
(by calling `Deno.core.setUpAsyncStub` that generates an optimized
version of an op, with promise accounting and error mapping).

Starting with this commit `00_primordials.js` and `00_infra.js` are
executed in `bindings::initialize_context`, instead of `JsRuntime::init_extension_js`.
These two files are necessary to do anything useful in `deno_core` -
including executing `01_core.js` (which relies on ops heavily).

String codegen was removed from `bindings::initialize_context` and replaced
with a direct for-loop in Rust that calls V8 APIs.

All the changes related to how context is initialized result in 0.2ms startup
time improvement in `fs_module_loader` example:
```
hyperfine --warmup 30 -S none './fs_module_loader_new empty.js' './fs_module_loader empty.js'
Benchmark 1: ./fs_module_loader_new empty.js
  Time (mean ± σ):       6.2 ms ±   0.2 ms    [User: 3.6 ms, System: 1.5 ms]
  Range (min … max):     5.9 ms …   8.2 ms    470 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 2: ./fs_module_loader empty.js
  Time (mean ± σ):       6.4 ms ±   0.2 ms    [User: 3.8 ms, System: 1.5 ms]
  Range (min … max):     6.1 ms …   8.1 ms    475 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Summary
  './fs_module_loader_new empty.js' ran
    1.03 ± 0.06 times faster than './fs_module_loader empty.js'

```

Closes https://github.com/denoland/deno_core/issues/445


bartlomieju:
Add a test that checks async stack trace with an async op - `ErrorCaptureStackTrace` setup is funky

core/00_infra.js
```diff
@@ -0,0 +1,473 @@
...
+      if (res.code) {
+        err.code = res.code;
+      }
+      // Strip unwrapOpResult() and errorBuilder() calls from stack trace
+      ErrorCaptureStackTrace(err, hideFunction);
+      throw err;
+    };
+  }
+
+  function setUpAsyncStub(opName) {
```
bartlomieju: `function asyncStub` is gone now

core/00_infra.js
```diff
@@ -0,0 +1,473 @@
...
+    // console.log("setUpAsyncStub", opName);
```
bartlomieju: Remove before landing

core/runtime/bindings.js
```diff
@@ -3,29 +3,21 @@
...
-      if (this !== core.ops) {
-        // deno-lint-ignore prefer-primordials
-        throw new Error(
-          "An async stub cannot be separated from Deno.core.ops. Use ???",
-        );
-      }
-      return core.asyncStub(opName, args);
-    };
-    fn.name = opName;
-    core.ops[opName] = fn;
```
bartlomieju: All of this was dead code

core/runtime/jsruntime.rs
```diff
@@ -992,6 +993,18 @@ impl JsRuntime {
...
+
+        // Execute 00_infra.js
+
+        // for each ext - execute "ext:<name>/ops"
+
+        // Execute 01_core.js
+
+        // Execute 02_error.js - probably can be merged into 01_core.js
+
+        // Execute mod.js
```
bartlomieju, line 1049: This is a TODO, so probably don't want to land it :)

core/runtime/bindings.rs
```diff
@@ -243,47 +242,42 @@ pub(crate) fn initialize_context<'s>(
 
   let global = context.global(scope);
 
-  let mut codegen = String::with_capacity(op_ctxs.len() * 200);
-  codegen.push_str(include_str!("bindings.js"));
-  _ = writeln!(codegen, "Deno.__op__ = function(opFns) {{");
```
bartlomieju, line 248: No more funky codegen :D

core/runtime/bindings.rs
```diff
@@ -243,47 +242,42 @@ pub(crate) fn initialize_context<'s>(
...
+  let deno_obj = get(scope, global, b"Deno", "Deno");
+  let deno_core_obj = get(scope, deno_obj, b"core", "Deno.core");
+  let deno_core_ops_obj: v8::Local<v8::Object> =
```
bartlomieju, line 247: Add comment what's going on in this function.

Also this function should be combined with `initialize_deno_namespace_object` and execute `00_primordials.js` and `00_infra.js`.

core/01_core.js
```diff
@@ -274,28 +271,15 @@
     );
   }
 
-  // Eagerly initialize ops for snapshot purposes
-  for (const opName of new SafeArrayIterator(ObjectKeys(asyncOps))) {
-    setUpAsyncStub(opName);
-  }
-
   function ensureFastOps(keep) {
```
bartlomieju, line 274: Sweet, this is now doing almost nothing

core/runtime/bindings.js
bartlomieju, line 1: Bye, bye

core/runtime/bindings.rs
```diff
@@ -169,29 +167,16 @@ pub(crate) fn initialize_deno_core_namespace<'s>(
...
   let deno_core_ops_key =
     v8::String::new_external_onebyte_static(scope, v8_static_strings::OPS)
       .unwrap();
 
-  let deno_core_async_ops_obj = v8::Object::new(scope);
-  let deno_core_async_ops_key = v8::String::new_external_onebyte_static(
-    scope,
-    v8_static_strings::ASYNC_OPS,
-  )
-  .unwrap();
```
bartlomieju, line 182: No more `Deno.core.asyncOps`

core/runtime/jsrealm.rs
```diff
@@ -255,7 +255,7 @@ impl JsRealm {
     unsafe { self.0.context.get_unchecked() as *const _ as _ }
   }
 
-  fn string_from_code<'a>(
+  pub(crate) fn string_from_code<'a>(
     scope: &mut HandleScope<'a>,
     code: &ModuleCodeString,
   ) -> Option<Local<'a, v8::String>> {
```
bartlomieju, line 261: This should probably be moved to `FastString`?

mmastrac (approved):
LGTM assuming deno tests pass

Issue #447, opened 2024-01-17
Disabling an op should result in replacing its underlying function


bartlomieju:
To remove some of the complexity around `Deno.__op__unregisterOp` and bifurcations in several parts of codebase we need to change the mechanism how "disabling" of ops work.

After talking with @mmastrac we decided we are gonna go with replacing the implementation of the op function with a stub function - either a noop or a function that throws.

This shouldn't be too hard to do, because op collection and registration is done before setting up context.

PR #456, opened 2024-01-18
feat: Disabling an op replaces it with a function that throws

bartlomieju:
This commit changes what "disabling" an op means. Before a disabled op
was never registered. This posed some challenges, especially as we
trying to provide "virtual ops modules" (https://github.com/denoland/deno_core/issues/223),
having ops conditionally available makes it super hard to properly
snapshot and set up the runtime.

With this change, "disabling" an op makes it throw when called.

Closes https://github.com/denoland/deno_core/issues/447

### DIFF
--- a/core/runtime/bindings.rs
+++ b/core/runtime/bindings.rs
@@ -24,7 +24,7 @@ pub(crate) fn external_references(
 ) -> v8::ExternalReferences {
   // Overallocate a bit, it's better than having to resize the vector.
   let mut references =
-    Vec::with_capacity(4 + (ops.len() * 4) + additional_references.len());
+    Vec::with_capacity(5 + (ops.len() * 4) + additional_references.len());
 
   references.push(v8::ExternalReference {
     function: call_console.map_fn_to(),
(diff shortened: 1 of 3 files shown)


mmastrac:
I prefer a default of throwing rather than a no-op, as some ops might have `null` or `undefined` as a return value with special meaning and we may end up in a case where code may appear to function correctly, but was just completely non-functional.

It might even be worth replacing the op with a panicking op by default as these ops are not intended to be reachable at all.

bartlomieju:
Discussed offline with Luca and Matt - we're gonna throw an exception from a disabled op in order not to bring the whole process down. This might change in the future.

core/runtime/bindings.rs
```diff
@@ -493,14 +496,24 @@ fn import_meta_resolve(
...
 
+pub(crate) fn op_disabled_fn(
+  scope: &mut v8::HandleScope,
+  _args: v8::FunctionCallbackArguments,
+  _rv: v8::ReturnValue,
+) {
+  let message = v8::String::new(scope, "op is disabled").unwrap();
+  let exception = v8::Exception::error(scope, message);
+  scope.throw_exception(exception);
+}
```
mmastrac, line 515: Nice!

mmastrac (approved):
LGTM

PR #457, opened 2024-01-19
feat: Allow synthetic modules to provide multiple exports

bartlomieju:
Prerequisite for "virtual ops modules".

This allows to create synthetic modules with multiple
exports. Previously only "default" export was supported,
which was fine for JSON modules and custom evaluated
modules, but lacks flexibility needed to generate modules
for ops in an extension.

### DIFF
--- a/core/modules/map.rs
+++ b/core/modules/map.rs
@@ -1,3 +1,4 @@
+use crate::FastString;
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 use crate::error::exception_to_err_result;
 use crate::error::generic_error;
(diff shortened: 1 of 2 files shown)


core/modules/map.rs
```diff
@@ -396,11 +400,12 @@ impl ModuleMap {
     scope: &mut v8::HandleScope,
     name: ModuleName,
     module_type: ModuleType,
-    value: v8::Local<v8::Value>,
+    exports: Vec<(v8::Local<v8::String>, v8::Local<v8::Value>)>,
```
mmastrac, line 403: Can we make this a `HashMap<FastString, v8::Local<v8::Value>>` here?

mmastrac (approved):
LGTM, two small suggestions

PR #458, opened 2024-01-19
Reduce ambiguous paths in macros

H1rono:
Current macros are not hygienic in path resolution. This causes compile error with codes like below.

```rust
deno_core::extension!(
    ext,
    js = [ "/path/to/ext.js" ]
);

mod std {}
```

error diagnostic:

```
error[E0433]: failed to resolve: could not find `borrow` in `std`
  --> src/lib.rs:14:1
   |
14 | / deno_core::extension!(
15 | |     ext,
16 | |     js = [ "/path/to/ext.js" ],
17 | | );
   | |_^ could not find `borrow` in `std`
   |
   = note: this error originates in the macro `deno_core::extension` (in Nightly builds, run with -Z macro-backtrace for more info)
```

This PR almost resolves this problem, by using unambiguous paths. There are still ambiguous paths remaining, but they won't couse such a issue because they only appear in unexported macros. I'm sorry if I missed something :pray:

references about hygienic macro:

- https://en.wikipedia.org/wiki/Hygienic_macro
- https://doc.rust-lang.org/reference/macros-by-example.html#hygiene
- https://gist.github.com/Kestrer/8c05ebd4e0e9347eb05f265dfb7252e1

### DIFF
--- a/core/extensions.rs
+++ b/core/extensions.rs
@@ -211,7 +211,7 @@ impl OpDecl {
 #[macro_export]
 macro_rules! ops {
   ($name:ident, parameters = [ $( $param:ident : $type:ident ),+ ], ops = [ $( $(#[$m:meta])* $( $op:ident )::+ $( < $op_param:ident > )?  ),+ $(,)? ]) => {
-    pub(crate) fn $name < $( $param : $type + 'static ),+ > () -> Vec<$crate::OpDecl> {
+    pub(crate) fn $name < $( $param : $type + 'static ),+ > () -> ::std::vec::Vec<$crate::OpDecl> {
       vec![
       $(
         $( #[ $m ] )*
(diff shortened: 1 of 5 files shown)


mmastrac (changes requested):
Thanks for the PR -- for unexported macros, I prefer to keep the hygiene to a minimum for readability purposes.

Can we revert everything outside of `extension!`, `external!`, `include_ascii_string!`, `located_script_name!` and `resources.rs`?

H1rono:
OK, I'll revert other macros.

mmastrac (approved):
LGTM, thanks! 

@H1rono did you have some way to automatically expand the macro paths, or was this done by hand?

H1rono:
By hand, unfortunately... I think [cargo-expand](https://github.com/dtolnay/cargo-expand) might be useful for automatic expanding.

PR #460, opened 2024-01-20
feat: Infrastructure for op sanitizer in deno_core

mmastrac:
This is the underlying infrastructure we need to get the op sanitizer working in Rust. There are no useful public-facing APIs from this yet, but it implements the op portion of the work.

Step 2 will be adding resources and timers to the API, then we'll be able to implement this in Deno itself.

### DIFF
--- a/core/runtime/op_driver/erased_future.rs
+++ b/core/runtime/op_driver/erased_future.rs
@@ -4,17 +4,20 @@ use std::marker::PhantomData;
 use std::marker::PhantomPinned;
 use std::mem::MaybeUninit;
 use std::pin::Pin;
+use std::ptr::addr_of_mut;
+use std::ptr::NonNull;
 use std::task::Context;
 use std::task::Poll;
 
 #[repr(C, align(16))]
 pub struct TypeErased<const MAX_SIZE: usize> {
   memory: [MaybeUninit<u8>; MAX_SIZE],
-  drop: fn(this: *mut ()),
+  drop: fn(this: NonNull<()>),
   _unpin: PhantomPinned,
 }
 
 impl<const MAX_SIZE: usize> TypeErased<MAX_SIZE> {
+  #[inline(always)]
   pub unsafe fn take<R: 'static>(self) -> R {
     assert!(
       std::mem::size_of::<R>() <= std::mem::size_of_val(&self.memory),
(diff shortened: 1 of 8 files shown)


Cargo.lock
bartlomieju, line 1: Where is this diff coming from if there are no changes in `Cargo.toml`? Can we skip these updates?

core/runtime/op_driver/erased_future.rs
```diff
@@ -1,20 +1,24 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+#![allow(dead_code)]
```
bartlomieju, line 2: Yikes, I don't like these because `clippy` doesn't catch that the code is actually used and they rot forever.

core/runtime/op_driver/futures_unordered_driver.rs
```diff
@@ -241,6 +229,18 @@ impl<C: OpMappingContext> OpDriver<C> for FuturesUnorderedDriver<C> {
   fn len(&self) -> usize {
     self.len.get()
   }
+
+  fn stats(&self) -> OpInflightStats {
+    let q = self.queue.queue.queue.borrow();
```
bartlomieju, line 234: `queue.queue.queue.queue.queue.queue.queue.queue.queue.queue.` 😅

core/runtime/op_driver/joinset_driver.rs
```diff
@@ -157,4 +154,8 @@ impl<C: OpMappingContext> OpDriver<C> for JoinSetDriver<C> {
   fn len(&self) -> usize {
     self.pending_ops.borrow().len()
   }
+
+  fn stats(&self) -> OpInflightStats {
+    OpInflightStats::default()
+  }
```
bartlomieju, line 160: This means that stats are never collected for joinset driver?

core/runtime/op_driver/mod.rs
```diff
@@ -19,9 +19,19 @@ pub use joinset_driver::JoinSetDriver;
...
 pub use self::op_results::V8RetValMapper;
 
+#[derive(Default)]
+#[allow(dead_code)] // follow-up PR
+/// Returns a set of stats on inflight ops.
+pub struct OpInflightStats {
+  /// The [`PromiseId`] of any inflight ops by each [`OpId`].
+  ops: Box<[PendingOpInfo]>,
+  supported: bool,
+}
```
bartlomieju, line 33: Could we avoid boxing here if we stored a slice and added a lifetime to `OpInflightStats`?

bartlomieju: Probably not a big deal though.

bartlomieju (approved):

Cargo.lock
mmastrac: I'll back this out. I think I ran a `cargo update` along the way.

core/runtime/op_driver/joinset_driver.rs
mmastrac: Correct -- that driver will hopefully go away soon once we're confident (maybe after 1.40?)

core/runtime/op_driver/mod.rs
mmastrac: I figured that these are not collected in a hot loop (and the docstring explicitly warns against that), so we should be OK to allocate.

core/runtime/op_driver/erased_future.rs
mmastrac: Let me see what I can trim here. I might be able to clean it up.

core/runtime/op_driver/futures_unordered_driver.rs
mmastrac: It's starting to get a little nested 😓

PR #466, opened 2024-01-23
feat(ops): add `#[cppgc]` argument attribute

littledivy:
Builtin support for getting a reference to a `CppGcObject<T>`.

```rust
#[op2(fast)]
fn op_file(#[cppgc] file: &std::fs::File) {}
```

### DIFF
--- a/ops/op2/dispatch_fast.rs
+++ b/ops/op2/dispatch_fast.rs
@@ -713,6 +713,12 @@ fn map_v8_fastcall_arg_to_arg(
       )?;
       quote!(let #arg_ident = #fast_api_callback_options.wasm_memory as _; #convert; let #arg_ident = Some(#arg_ident);)
     }
+    Arg::Special(Special::CppGcResource(ty)) => {
+      quote! {
+        let #arg_ident = #arg_ident.try_into().unwrap();
+        let #arg_ident = deno_core::cppgc::unwrap_cppgc_object::<#ty>(#arg_ident).unwrap();
+      }
+    }
     _ => quote!(let #arg_ident = #arg_ident as _;),
   };
   Ok(res)
@@ -820,6 +826,7 @@ fn map_arg_to_v8_fastcall_type(
     // Cow byte strings can be fast and don't require copying
     Arg::String(Strings::CowByte) => V8FastCallType::SeqOneByteString,
     Arg::External(..) => V8FastCallType::Pointer,
+    Arg::Special(Special::CppGcResource(_)) => V8FastCallType::V8Value,
     _ => return Err("a fast argument"),
   };
   Ok(Some(rv))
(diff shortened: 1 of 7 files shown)


ops/op2/signature.rs
```diff
@@ -138,13 +138,14 @@ impl ToTokens for V8Arg {
...
 
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub enum Special {
   HandleScope,
   OpState,
   JsRuntimeState,
   FastApiCallbackOptions,
   Isolate,
+  CppGcResource(syn::Path),
```
mmastrac, line 148: Use String here -- that's the current pattern for embedded types like this

mmastrac (commented):
Looks good, but before we land this let's add some integration tests to the testing folder to make sure it doesn't break.

ops/op2/signature.rs
littledivy: I don't this its good to use Strings. Eg: `quote!`-ing a `String` here becomes:

```rust
unwrap_cpp_object::<"std::fs::File">();
```

`parse_ident` wouldn't work because `std::fs::File` is not a ident. I'd have to reparse the String to a syn::Path during codegen.

ops/op2/dispatch_fast.rs
```diff
@@ -713,6 +713,17 @@ fn map_v8_fastcall_arg_to_arg(
       )?;
       quote!(let #arg_ident = #fast_api_callback_options.wasm_memory as _; #convert; let #arg_ident = Some(#arg_ident);)
     }
+    Arg::Special(Special::CppGcResource(ty)) => {
+      *needs_fast_api_callback_options = true;
+      quote! {
+        let #arg_ident = #arg_ident.try_into().unwrap();
+        let Some(#arg_ident) = deno_core::cppgc::unwrap_cppgc_object::<#ty>(#arg_ident) else {
```
mmastrac, line 720: `unwrap_cppgc_object` should be `try_unwrap_cppgc_object`

I think we may also want to change that method to take either `&v8::Local<...>` or `&v8::Object` so that the caller must hold the reference to the v8 object while the Rust object is alive.

ops/op2/dispatch_slow.rs
```diff
@@ -506,6 +507,19 @@ pub fn from_arg(
         };
       }
     }
+    Arg::Special(Special::CppGcResource(ty)) => {
+      let throw_exception = throw_type_error(
+        generator_state,
+        format!("expected {}", stringify_token(ty)),
```
mmastrac, line 513: Note that we should add some tests for this.

mmastrac (commented):
Added a second round of closer review

ops/op2/signature.rs
mmastrac: Yeah -- we currently re-parse those all through codegen in the other places as well. It's a tradeoff between keeping the Arg interface somewhat uncoupled from syn types.

It keeps the layers a little less tightly coupled. Let's use String for now -- if we decide to use syn types throughout the stack we should fix it for all the types.

littledivy: Done

ops/op2/dispatch_fast.rs
littledivy: Renamed to `try_unwrap`. 

Yeah I think we want to add lifetime restrictions but shouldn't be a problem for sync ops.

mmastrac (approved):
LGTM. We'll implement a cppgc integration test as a follow-up

PR #468, opened 2024-01-23
feat: expose op diffs to sanitizer interface

mmastrac:


core/00_primordials.js
```diff
@@ -36,6 +36,15 @@
 (() => {
   const primordials = {};
 
+  if (!Symbol.dispose) {
```
mmastrac, line 39: I think we should have these in deno_core -- wasn't sure exactly where the right place to install them would be.

bartlomieju: Is there a particular reason for it since V8 doesn't support `using` syntax right now anyway?

If we are to do it, I'd prefer we did this:
https://github.com/denoland/deno/blob/cc05dc142d1c57afd10cb08d84f99a25bd49678c/runtime/js/99_main.js#L78-L79

core/runtime/stats.rs
```diff
@@ -0,0 +1,152 @@
...
+
+#[derive(Clone)]
+pub struct RuntimeActivityStatsFactory {
+  pub(super) context_state: Rc<ContextState>,
+  pub(super) op_state: Rc<RefCell<OpState>>,
+}
+
+impl RuntimeActivityStatsFactory {
+  pub fn capture(self) -> RuntimeActivityStats {
+    let mut resources = ResourceOpenStats { resources: vec![] };
```
bartlomieju, line 21: Pre-allocate the vector to proper size

core/runtime/stats.rs
```diff
@@ -0,0 +1,152 @@
...
+    for resource in self.op_state.borrow().resource_table.names() {
+      resources
+        .resources
+        .push((resource.0, resource.1.to_string()))
+    }
+    let timers = TimerStats {
+      intervals: vec![],
+      timers: vec![],
```
bartlomieju, line 29: Pre-allocate to 64-128?

core/runtime/stats.rs
```diff
@@ -0,0 +1,152 @@
...
+  Resource(ResourceId, String),
+  Timer(usize, usize),
+  Interval(usize, usize),
+}
+
+impl RuntimeActivityStats {
+  /// Capture the data within this [`RuntimeActivityStats`] as a [`RuntimeActivitySnapshot`]
+  /// with details of activity.
+  pub fn dump(&self) -> RuntimeActivitySnapshot {
+    let mut v = vec![];
```
bartlomieju, line 73: Pre-allocate the vector

core/00_primordials.js
bartlomieju: I see that you use it in `checking`. I'm somewhat wary of adding this symbol here, could you move it to `checkin` setup code?

bartlomieju (approved):

core/00_primordials.js
mmastrac: Yep, that works!

PR #470, opened 2024-01-23
feat: Add metadata to op2 macro

### LINKED ISSUE #467: `#[op2(async)]` should require to specify sanitizer diagnostics
We will add a `#[meta` tag for `op2` that allows embedders to provide custom metadata for any op. Deno will have a unit test confirming that all async ops added from extensions have these two meta values.

```
#[op2(async)]
#[meta async_verb = "read from a Blob or File"]
#[meta async_fix = "awaiting the result of a Blob or File read"]
```

bartlomieju:
Closes #467


ops/op2/config.rs
```diff
@@ -85,6 +90,22 @@ impl MacroConfig {
...
       } else if flag == "reentrant" {
         config.reentrant = true;
+      } else if flag == "meta" {
+        let tokens =
+          syn::parse_str::<TokenTree>(&flag[4..])?.into_token_stream();
+        config.meta = std::panic::catch_unwind(|| {
+          rules!(tokens => {
+            ( $( $name:literal = $value:literal ),* ) => {
+              name.into_iter().zip(value.into_iter()).map(|(name, val)| {
+                let name_static: &'static str = name.to_string().leak();
```
mmastrac, line 100: We definitely should not leak here.

ops/op2/config.rs
```diff
@@ -85,6 +90,22 @@ impl MacroConfig {
...
+                let val_static: &'static str = val.to_string().leak();
+                (name_static, val_static)
+              })
+            }
+          })
+        })
+        .map_err(|_| Op2Error::InvalidMetaAttributes)?
+        .collect::<HashMap<&'static str, &'static str>>();
```
mmastrac, line 108: Collect as `String, String`

ops/op2/mod.rs
```diff
@@ -185,6 +187,15 @@ fn generate_op2(
   };
   let is_async = signature.ret_val.is_async();
   let is_reentrant = config.reentrant;
+  let meta_token_stream: TokenStream = config
+    .meta
+    .iter()
+    .map(|(key, value)| {
+      quote! {
+          #key: #value,
```
mmastrac, line 195: We actually want the output to look something like this (a `const` `HashMap`). You should probably create something that looks like this (the `HashMap` bits will probably need to be in the `DECL` below.

```
::std::collections::HashMap<&str, &str> = [

#( ( #key, #value ), )*

].iter().cloned().collect();
```

PR #471, opened 2024-01-23
fix: Fix segmentation fault in tests on CPUs with PKU support

Digifox03:
Due to V8's PKU feature, only the thread that initialized the V8 platform or those spawned from it can access the isolates; other threads will get a segmentation fault. Since each test runs in its thread, they crash on newer CPUs that support PKU.

This issue has two possible solutions:
1. Initialize the platform once in the main thread.
2. Turn off the PKU feature during testing.

This issue was already encountered for the `deno test` command, which was resolved by initializing the platform in the main thread (denoland/deno#20495).

In the case of rust tests, the same solution cannot be used, as the default test runner does not allow running code on the main thread before any test runs.

My proposed solution is to add a feature flag to deno_core, which, when enabled, uses the unprotected platform. An alternative solution could be to call `deno_core::JsRuntime::init_platform` at the beginning of each test.

This PR will also be a step towards fixing denoland/deno#21439


core/Cargo.toml
```diff
@@ -19,6 +19,7 @@ include_icu_data = []
 v8_use_custom_libcxx = ["v8/use_custom_libcxx"]
 include_js_files_for_snapshotting = []
 unsafe_runtime_options = []
+testing = []
```
mmastrac, line 22: Let's rename this to unsafe_use_unprotected_platform

core/runtime/jsruntime.rs
```diff
@@ -413,8 +413,16 @@ fn v8_init(
   };
   v8::V8::set_flags_from_string(&flags);
 
-  let v8_platform = v8_platform
-    .unwrap_or_else(|| v8::new_default_platform(0, false).make_shared());
+  let v8_platform = v8_platform.unwrap_or_else(|| {
+    if cfg!(any(test, feature = "testing")) {
```
mmastrac, line 417: `unsafe_use_unprotected_platform` as well

serde_v8/magic/v8slice.rs
```diff
@@ -107,7 +107,8 @@ where
     static V8_ONCE: std::sync::Once = std::sync::Once::new();
 
     V8_ONCE.call_once(|| {
-      let platform = v8::new_default_platform(0, false).make_shared();
+      let platform =
```
mmastrac, line 110:
```suggestion
      // We want to use the unprotected platform for unit tests
      let platform =
```

serde_v8/utils.rs
```diff
@@ -11,7 +11,7 @@ pub fn js_exec<'s>(
 }
 
 pub fn v8_init() {
```
mmastrac, line 13: This file should be feature-flagged with a new `unsafe_v8_utils` feature as this function isn't really ideal for export if we're going to use the unprotected platform.

mmastrac (commented):
We'll probably need to tweak the features a bit but it looks like the right approach.

serde_v8/utils.rs
Digifox03: I'm not sure how to proceed with that. If we put the file behind a feature flag, the tests will need to have the feature flag enabled by default, which is currently not possible (rust-lang/cargo#2911)

bartlomieju:
Do we need to add a Cargo feature here? Maybe we could rely on `#[cfg(test)]` attribute instead?

serde_v8/utils.rs
mmastrac: Could we use `#[cfg(any(test, feature="..."))]` again for that? The problem is really that the benchmarks need this.

mmastrac:
> Do we need to add a Cargo feature here? Maybe we could rely on `#[cfg(test)]` attribute instead?

I think the problem is that some downstream crates need the PKU workaround.

serde_v8/utils.rs
Digifox03: `cfg(test)` won't work either, as it is not enabled for integration tests, nor benchmarks.

mmastrac: If we use (any(test, feature)) won't it work?

Digifox03: As far as I understand, benchmarks and integration tests behave as if they are a separate crate with the library as their dependency with only the default feature flags enabled. This means that if the code is hidden behind a `cfg(feature = ...)` it won't be accessible. This also prevents the usage of `cfg(test)` because the `serde_v8` is not compiled for testing.

The solution I have found for hiding the utilities is to move them to an internal crate, as suggested in this [Stack Overflow answer](https://stackoverflow.com/a/44545091).

mmastrac: Take a look here -- we currently have benches that require internal-only features: https://github.com/denoland/deno_core/blob/main/core/Cargo.toml#L70

mmastrac: Great, thanks for doing this work! We'll try to land this and then it should be available for deno upstream.

mmastrac (approved):
LGTM, thanks!

PR #472, opened 2024-01-24
refactor(core): split up resource code and add a resource test

mmastrac:


core/io/resource.rs
```diff
@@ -0,0 +1,286 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+
+// Think of Resources as File Descriptors. They are integers that are allocated
+// by the privileged side of Deno which refer to various rust objects that need
+// to be persisted between various ops. For example, network sockets are
+// resources. Resources may or may not correspond to a real operating system
+// file descriptor (hence the different name).
```
bartlomieju, line 7: Remove?

bartlomieju (approved):

core/io/resource.rs
mmastrac: I think we need to leave it until we get rid of `rid`s completely 😢

PR #480, opened 2024-01-24
chore: improve publish dry-run

mmastrac:
Changes to publishing/build workflows:

 - No longer build twice when running a version bump. The only branch that will trigger a build will be `main`.
 - Tag builds run an empty matrix, since they are being built from a main commit that should succeed, and the publish workflow runs a test compile on top of everything.
   - Even if the tag build failed with an error, we still just went ahead and published that tag!
 - All PRs run a `cargo publish` dry-run against a test server


bartlomieju (approved):
LGTM, exciting!

PR #482, opened 2024-01-24
feat(core): add resources to stats test

mmastrac:


testing/checkin/runtime/async.ts
```diff
@@ -89,6 +89,16 @@ export class StatsCollection {
     return count;
   }
 
+  countResources(): number {
+    let count = 0;
+    for (const item of this.data) {
+      if ("Resource" in item) {
```
bartlomieju, line 95: Where is thing magic string coming from? Could you add a comment describing it?

bartlomieju (approved):

testing/checkin/runtime/async.ts
mmastrac: These come from `ResourceActivity` which is `#[serde]` return type from these ops. Note that these are checkin ops so this won't be in Deno itself. That code will be pure Rust.

I'll add some notes pointing at what these types are.

PR #483, opened 2024-01-25
feat(core): sanitization for timers

mmastrac:


core/runtime/stats.rs
```diff
@@ -27,10 +27,17 @@ impl RuntimeActivityStatsFactory {
...
-      timers: vec![],
+    let timer_count = self.context_state.timers.len();
+    let mut timers = TimerStats {
+      timers: Vec::with_capacity(timer_count),
+      repeats: BitSet::with_capacity(timer_count),
     };
+    for (timer_id, repeats) in &self.context_state.timers.iter() {
+      if repeats {
+        timers.repeats.insert(timers.timers.len());
+      }
```
bartlomieju, line 38: Could you explain this line?

bartlomieju (approved):

core/runtime/stats.rs
mmastrac: `repeats` is a bitset that reports whether a given index in the ID array is an interval (if true) or a timer (if false). I'll put this into a comment.

core/runtime/stats.rs
```diff
@@ -45,10 +52,8 @@ pub struct ResourceOpenStats {
 }
 
 pub struct TimerStats {
-  #[allow(dead_code)] // coming soon
   pub(super) timers: Vec<usize>,
-  #[allow(dead_code)] // coming soon
-  pub(super) intervals: Vec<usize>,
+  pub(super) repeats: BitSet,
```
mmastrac, line 56:
```suggestion
    /// `repeats` is a bitset that reports whether a given index in the ID array
    /// is an interval (if true) or a timer (if false).
    pub(super) repeats: BitSet,
```

PR #484, opened 2024-01-25
chore: add docs to github actions

mmastrac:
This should help all of us when swapping the build knowledge back in.


bartlomieju (approved):
Thanks, this is super useful.

PR #491, opened 2024-01-26
feat: Add ModuleLoadResponse

bartlomieju:
Checking if having this enum is beneficial and can help optimize some of the code.

In many situation embedders will rely on `ModuleLoader::prepare_module_load` 
and assemble the whole graph themselves. Then when actually loading modules,
they will wrap the already ready value into an empty future that needs to be boxed.

### DIFF
--- a/core/examples/ts_module_loader.rs
+++ b/core/examples/ts_module_loader.rs
@@ -6,7 +6,6 @@
 
 use std::cell::RefCell;
 use std::collections::HashMap;
-use std::pin::Pin;
 use std::rc::Rc;
 
 use anyhow::anyhow;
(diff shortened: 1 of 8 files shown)


core/examples/ts_module_loader.rs
bartlomieju, line 1: This file is duplicated in `checkin`. Let's bite the bullet and add `transpile` Cargo feature that provides this loader in `core/modules/loaders` so it can be easily reused.

core/modules/recursive_load.rs
```diff
@@ -265,14 +266,17 @@ impl RecursiveModuleLoad {
...
+                &specifier,
+                Some(&referrer),
+                is_dynamic_import,
+                requested_module_type,
+              );
+
+              let load_result = match load_response {
+                ModuleLoadResponse::Sync(result) => result,
+                ModuleLoadResponse::Async(fut) => fut.await,
+              };
```
bartlomieju, line 279: This is the critical file; check if having sync response can cause some optimizations here.

mmastrac (approved):
Nice improvement.

PR #506, opened 2024-01-28
refactor: Untangle "OpCtx" and "ContextState"

bartlomieju:
This commit removes entanglement between "OpCtx" and "ContextState" structs.

"OpCtx" no longer requires "ContextState" to be passed as an argument. Instead
it accepts "op_driver" arg which is "OpDriverImpl" (renamed from "DefaultOpDriver").

This allows us to fully set up "OpCtx" for every op and only then create "ContextState".
This unlocks a lock of further clean ups that are done in https://github.com/denoland/deno_core/pull/499.


core/ops.rs
```diff
@@ -173,9 +175,10 @@ impl OpCtx {
     *opt_mut = Some(error);
   }
 
-  /// Get the [`ContextState`] for this op.
-  pub(crate) fn context_state(&self) -> &ContextState {
-    &self.context_state
+  // TODO(bartlomieju): this forces use of `DefaultOpDriver`, but making it
+  // generic is a bit cumbersome. Probably need to address before landing.
+  pub(crate) fn op_driver(&self) -> &DefaultOpDriver {
+    &self.op_driver
```
bartlomieju, line 181: This is the crux of the fix - we only needed to have `ContextState` here so we could dispatch ops. But we can just pass "op driver" instead. 

This needs to be fixed before landing, because it (incorrectly) defines that we're using `DefaultOpDriver`.

Requesting help.

core/ops_metrics.rs
bartlomieju, line 1: Changes to this file might not be necessary to land this PR - verify that.

Still it's a nice cleanup so maybe we can just include it?

core/runtime/jsrealm.rs
```diff
@@ -75,10 +75,16 @@ pub(crate) struct ContextState<OpDriverImpl: OpDriver = DefaultOpDriver> {
 
 impl<O: OpDriver> ContextState<O> {
   pub(crate) fn new(
+    op_driver: Rc<O>,
     isolate_ptr: *mut v8::OwnedIsolate,
     get_error_class_fn: GetErrorClassFn,
-    ops_with_metrics: Vec<bool>,
+    op_ctxs: Box<[OpCtx]>,
```
bartlomieju, line 81: Look - no more entanglement between `ContextState` and `OpCtx` - now we can pass fully initialized ops here and remove a lot of complexity and spaghetti code.

core/runtime/jsrealm.rs
```diff
@@ -62,10 +62,10 @@ pub(crate) struct ContextState<OpDriverImpl: OpDriver = DefaultOpDriver> {
     RefCell<Option<Rc<v8::Global<v8::Function>>>>,
   pub(crate) unrefed_ops:
     RefCell<HashSet<i32, BuildHasherDefault<IdentityHasher>>>,
-  pub(crate) pending_ops: OpDriverImpl,
+  pub(crate) pending_ops: Rc<OpDriverImpl>,
   // We don't explicitly re-read this prop but need the slice to live alongside
   // the context
-  pub(crate) op_ctxs: RefCell<Box<[OpCtx]>>,
+  pub(crate) op_ctxs: Box<[OpCtx]>,
```
bartlomieju, line 68: No more `RefCell` here!

core/runtime/jsrealm.rs
```diff
@@ -62,10 +62,10 @@ pub(crate) struct ContextState<OpDriverImpl: OpDriver = DefaultOpDriver> {
     RefCell<Option<Rc<v8::Global<v8::Function>>>>,
   pub(crate) unrefed_ops:
     RefCell<HashSet<i32, BuildHasherDefault<IdentityHasher>>>,
-  pub(crate) pending_ops: OpDriverImpl,
+  pub(crate) pending_ops: Rc<OpDriverImpl>,
```
bartlomieju, line 65: I think this is fine - instead of cloning `Rc<ContextState>` for each op, we clone `OpDriver` instead.

core/runtime/jsruntime.rs
```diff
@@ -724,27 +725,8 @@ impl JsRuntime {
...
-      .task_spawner_factory
-      .clone()
-      .new_same_thread_spawner();
-    op_state.borrow_mut().put(spawner);
-    let spawner = context_state
-      .task_spawner_factory
-      .clone()
-      .new_cross_thread_spawner();
-    op_state.borrow_mut().put(spawner);
+    let op_driver = Rc::new(DefaultOpDriver::default());
```
bartlomieju, line 729: Instead of passing `ContextState`, we now need to pass `OpDriver` to `OpCtx`.

core/runtime/jsrealm.rs
```diff
@@ -181,8 +187,6 @@ impl JsRealmInner {
     state.exception_state.prepare_to_destroy();
     std::mem::take(&mut *state.js_event_loop_tick_cb.borrow_mut());
     std::mem::take(&mut *state.js_wasm_streaming_cb.borrow_mut());
-    // The OpCtx slice may contain a circular reference
-    std::mem::take(&mut *state.op_ctxs.borrow_mut());
```
bartlomieju, line 185: This needs to be verified with care - I believe this is fine, but not 100% sure. Probably the best to run this against `deno` repo and see if it causes any failures.

core/ops.rs
bartlomieju: Actually maybe using `DefaultOpDriver` is fine here? Since we give no way to switch driver besides Cargo features - and switching a feature will just switch out the default driver.

core/runtime/jsrealm.rs
bartlomieju: Actually I think this is fine

core/ops.rs
```diff
@@ -72,30 +73,38 @@ pub struct OpCtx {
...
-  context_state: Rc<ContextState>,
+  op_driver: Rc<DefaultOpDriver>,
   runtime_state: Rc<JsRuntimeState>,
 }
 
 impl OpCtx {
   #[allow(clippy::too_many_arguments)]
   pub(crate) fn new(
+    // TODO(bartlomieju): remove
+    op_count: usize,
```
bartlomieju, line 92: I really don't like this, but `OpMetricsFactoryFn` requires it. It appears we only use the total count in `OpMetricsSummaryTracker` that is only used in testing code.

Could we maybe remove `OpMetricsFactoryFn` to accept the total count of ops?

bartlomieju: If that's okay, then maybe this factory function is not needed anymore?

core/ops.rs
bartlomieju: Maybe it should be named `RealOpDriver`?

core/ops.rs
```diff
@@ -72,13 +72,13 @@ pub struct OpCtx {
   #[doc(hidden)]
   pub get_error_class_fn: GetErrorClassFn,
 
-  pub(crate) decl: Rc<OpDecl>,
+  pub(crate) decl: OpDecl,
```
bartlomieju, line 75: Don't need to wrap `OpDecl` in `Rc` anymore

core/runtime/jsrealm.rs
```diff
@@ -62,33 +62,32 @@ pub(crate) struct ContextState<OpDriverImpl: OpDriver = DefaultOpDriver> {
...
+  pub(crate) pending_ops: Rc<OpDriverImpl>,
   // We don't explicitly re-read this prop but need the slice to live alongside
   // the context
-  pub(crate) op_ctxs: RefCell<Box<[OpCtx]>>,
+  pub(crate) op_ctxs: Box<[OpCtx]>,
   pub(crate) isolate: Option<*mut v8::OwnedIsolate>,
   pub(crate) exception_state: Rc<ExceptionState>,
   pub(crate) has_next_tick_scheduled: Cell<bool>,
   pub(crate) get_error_class_fn: GetErrorClassFn,
-  pub(crate) ops_with_metrics: Vec<bool>,
```
bartlomieju, line 73: Not required anymore, one less vec to allocate.

core/ops.rs
bartlomieju: Went with `RealOpDriver`.

core/ops.rs
```diff
@@ -72,13 +72,13 @@ pub struct OpCtx {
...
   pub(crate) fast_fn_c_info: Option<NonNull<v8::fast_api::CFunctionInfo>>,
   pub(crate) metrics_fn: Option<OpMetricsFn>,
   /// If the last fast op failed, stores the error to be picked up by the slow op.
   pub(crate) last_fast_error: UnsafeCell<Option<AnyError>>,
 
-  context_state: Rc<ContextState>,
+  op_driver: Rc<RealOpDriver>,
```
bartlomieju, line 81: This is the main change here.

core/runtime/jsrealm.rs
```diff
@@ -48,12 +48,12 @@ impl Hasher for IdentityHasher {
   feature = "op_driver_joinset",
   not(feature = "op_driver_futuresunordered")
 ))]
-type DefaultOpDriver = super::op_driver::JoinSetDriver;
+pub(crate) type RealOpDriver = super::op_driver::JoinSetDriver;
```
mmastrac, line 51: Let's actually rename this to `OpDriverImpl`. We can get rid of the generic on `ContextState`.

core/runtime/jsrealm.rs
```diff
@@ -48,12 +48,12 @@ impl Hasher for IdentityHasher {
...
 
 #[cfg(feature = "op_driver_futuresunordered")]
-type DefaultOpDriver = super::op_driver::FuturesUnorderedDriver;
+pub(crate) type RealOpDriver = super::op_driver::FuturesUnorderedDriver;
 
-pub(crate) struct ContextState<OpDriverImpl: OpDriver = DefaultOpDriver> {
+pub(crate) struct ContextState<OpDriverImpl: OpDriver = RealOpDriver> {
```
mmastrac, line 56: Remove this generic and just use `OpDriverImpl`. This particular generic was more important during the testing phase, but we never actually need to make `ContextState` generic.

core/runtime/jsruntime.rs
```diff
@@ -2272,17 +2259,14 @@ impl JsRuntime {
...
-            &context_state.op_ctxs.borrow()[op_id as usize],
-            OpMetricsEvent::CompletedAsync,
-          );
-        } else {
-          dispatch_metrics_async(
-            &context_state.op_ctxs.borrow()[op_id as usize],
-            OpMetricsEvent::ErrorAsync,
-          );
+      {
+        let op_ctx = &context_state.op_ctxs[op_id as usize];
```
mmastrac, line 2263: Nice -- no borrow.

core/ops.rs
```diff
@@ -173,9 +173,8 @@ impl OpCtx {
     *opt_mut = Some(error);
   }
 
-  /// Get the [`ContextState`] for this op.
-  pub(crate) fn context_state(&self) -> &ContextState {
-    &self.context_state
+  pub(crate) fn op_driver(&self) -> &RealOpDriver {
```
mmastrac, line 176: Much better.

mmastrac (approved):
LGTM, some small changes before landing.

core/runtime/jsrealm.rs
mmastrac: My only worry here is that you might end up failing to drop the OpDriver in time, which may leave some pending ops that tend to cause trouble.

We can probably add a shutdown method to the driver for this case.

PR #509, opened 2024-01-29
fix: Remove 'leak' call when creating `v8::ExternalReferences`

svix-aaron1011:
Calling `v8::ExternalReferences::new` does not take ownership of the Vec - it internally copies the values into a new Vec. As a result, the memory owned by the original Vec was never getting freed.


bartlomieju (approved):
Thanks!

PR #520, opened 2024-01-30
feat: Add property definition utilities to Deno.core

bartlomieju:
These are common utilities that are duplicated several times in `deno` and Deploy.

### DIFF
--- a/core/01_core.js
+++ b/core/01_core.js
@@ -405,6 +405,7 @@
     op_get_promise_details,
     op_get_proxy_details,
     op_has_tick_scheduled,
+    op_lazy_load_esm,
     op_memory_usage,
     op_op_names,
     op_panic,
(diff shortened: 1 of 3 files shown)


bartlomieju:
@crowlKats requested it

core/01_core.js
```diff
@@ -455,6 +455,65 @@
...
+  function propGetterOnly(getter) {
+    return {
+      get: getter,
+      set() {},
+      enumerable: true,
+      configurable: true,
+    };
+  }
+
+  function propNonEnumerableLazyLoaded(getter, loadFn) {
```
crowlKats, line 494: i wonder if these should rather be a function that returns a function, where the outer function takes the loadFn, since you'd pass a loadFn to every call, which for cases ie webgpu, is quite a lot. though we could also just create a function on the use-site if necessary enough.

On the topic of load functions: its annoying that we have to create an identical load fn just with a different holding variable for every lazy loadable module, it would be nice if we had a utility function for that. maybe something along the lines of:
```js
function createLazyLoader(specifier) {
  let value;
  
  return function lazyLoad() {
    if (!value) {
      value = ops.lazy_load_esm(specifier);
    }
    return value;
  }
}

File: core/01_core.js [hunk 2]
```diff
@@ -455,6 +455,65 @@
...
+      set(v) {
+        loadFn();
+        valueIsSet = true;
+        value = v;
+      },
+      enumerable: false,
+      configurable: true,
+    };
+  }
+
```
crowlKats, line 516: can also https://github.com/denoland/deno/blob/main/runtime/js/98_global_scope_shared.js#L65-L88 be added? propWritable lazy load counterpart

bartlomieju:
@crowlKats reminder to update `lib.deno_core.d.ts`

crowlKats (approved):

PR #523, opened 2024-01-31
refactor: internal ops are under Deno.core.coreOps

bartlomieju:
This commit changes how ops defined in "deno_core"
crate are exposed to JavaScript code.

They are no longer added to "Deno.core.ops", but to "Deno.core.coreOps",
to make the distinction between them more obvious.

All ops that are required for embedders are now available under "Deno.core"
namespace.

This refactor uncovered a lot of cruft and technical debt accumulated over the
years - eg. there are several ops that are added unconditionally that are only
ever used in testing code. This poses a question, if "deno_core" should expose
some of the ops only when `cfg(test)` passes.


core/02_error.js
bartlomieju, line 1: This file should probably be rolled into `01_core.js` to avoid having to expose these APIs to `Deno.core`

core/runtime/bindings.rs
```diff
@@ -241,10 +256,13 @@ pub(crate) fn initialize_primordials_and_infra(
 }
 
 /// Set up JavaScript bindings for ops.
-pub(crate) fn initialize_deno_core_ops_bindings<'s>(
```
bartlomieju, line 244: This function needs to be split into two separate functions - one for "core" ops and one for "ext" ops.

core/runtime/tests/misc.rs
```diff
@@ -826,7 +826,7 @@ async fn test_promise_rejection_handler_generic(
     function throwError() {
       throw new Error("boom");
     }
-    const { op_void_async, op_void_async_deferred } = Deno.core.ensureFastOps();
+    const { opVoidAsync, opVoidAsyncDeferred } = Deno.core;
```
bartlomieju, line 829: Need to verify it, but I believe `deno` relies on these ops as well

core/runtime/bindings.rs
```diff
@@ -127,7 +127,7 @@ where
     .get(scope, key.into())
     .unwrap_or_else(|| panic!("{path} exists"))
     .try_into()
-    .unwrap_or_else(|_| panic!("unable to convert"))
+    .unwrap_or_else(|_| panic!("Unable to convert {path} to desired type"))
```
bartlomieju, line 130: This funciton should probably be a macro and accept a token tree to print out the desired type nicely

core/runtime/bindings.rs
```diff
@@ -272,7 +310,41 @@ pub(crate) fn initialize_deno_core_ops_bindings<'s>(
   );
 
   let undefined = v8::undefined(scope);
-  for op_ctx in op_ctxs {
+
+  let deno_core_op_ctxs = &op_ctxs[..last_deno_core_op_id];
+  let ext_op_ctxs = &op_ctxs[last_deno_core_op_id..];
```
littledivy, line 315: I'm a bit afraid that we might be complicating it even more.

Pardon my ignorance, why do we need to seperate `core.coreOps` from `core.ops`?

bartlomieju: This is just first step in a bigger refactor - ops defined in `deno_core` should not be accessible outside of `deno_core` itself - if they need to be, they should be exposed as methods on `Deno.core` namespace. This temporary `Deno.core.coreOps` is only available for the short duration of `01_core.js` execution. Then it's removed immediately. These ops are necessary for the runtime to function properly.

That way, `Deno.core.ops` will only contain ops defined in extensions - then again, we want to move away from using `Deno.core.ops` altogether and instead import them directly from `ext:core/ops` (though I'm still facing some issues that I haven't figured out how to tackle). 

After some back and forth I decided that this is the right way and it requires some intermediate refactors - I want to split `op_ctxs` into two separate vectors for "core" and "ext" ops.

core/runtime/bindings.rs
```diff
@@ -113,6 +113,24 @@ pub fn script_origin<'a>(
...
+    )
+    .unwrap_or_else(|_| {
+      panic!(
+        "Unable to convert {} to desired {}",
+        $path,
+        stringify!($type)
+      )
+    })
+  }};
+}
```
bartlomieju, line 132: @mmastrac PTAL if this makes sense

core/runtime/bindings.rs
littledivy: Thanks for explaining, sounds good.

bartlomieju:
Not important anymore

PR #525, opened 2024-01-31
feat: Add `dcore` bin target

bartlomieju:
This commit adds minimal binary target to "deno_core" crate.

It is now possible to "cargo run ./path/to/file.ts" to execute a file.
This feature is being added for development and testing purposes only
and is not meant for end-users. The API is minimal and the binary can't
do anything interesting.

"fs_module_loader" example has been removed in favor of binary target.

Closes https://github.com/denoland/deno_core/issues/412
Once #415 lands we can extend functionality further.


mmastrac (approved):
LGTM!

PR #526, opened 2024-01-31
refactor: make static strings use easier

bartlomieju:
Just a bit of a cleanup.

### DIFF
--- a/core/runtime/bindings.rs
+++ b/core/runtime/bindings.rs
@@ -153,6 +153,7 @@ pub mod v8_static_strings {
   pub static CALL_CONSOLE: &[u8] = b"callConsole";
   pub static FILENAME: &[u8] = b"filename";
   pub static DIRNAME: &[u8] = b"dirname";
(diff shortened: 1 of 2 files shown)


core/runtime/bindings.rs
```diff
@@ -131,6 +131,13 @@ where
...
 
 pub mod v8_static_strings {
+  pub fn new<'s>(
+    scope: &mut v8::HandleScope<'s>,
+    str_: &'static [u8],
+  ) -> v8::Local<'s, v8::String> {
+    v8::String::new_external_onebyte_static(scope, str_).unwrap()
+  }
+
   pub static DENO: &[u8] = b"Deno";
```
mmastrac, line 141: Something we should consider (in the future): Aapo added OneByteConst strings that can be const, and do a bit of the v8 string setup at compile time. These strings would be ideal for that.

mmastrac (approved):
LGTM, nice cleanup

core/runtime/bindings.rs
bartlomieju: Opened https://github.com/denoland/deno_core/issues/527

PR #528, opened 2024-01-31
feat: Add bindings to analyse WASM modules

bartlomieju:
This commit adds "CustomModuleEvaluationCtx" struct that is passed to
"CustomModuleEvaluationCb" and allows to analyze contents of WASM modules.

Specifically we add the ability to analyze imports and exports of WASM modules
that can be used to build up support for WASM imports.


bartlomieju:
Final PR before https://github.com/denoland/deno_core/pull/402 lands

mmastrac (approved):

PR #530, opened 2024-02-01
feat: Source map support in extensions

Digifox03:
Implement support for source maps for extensions right into deno_core.

- [x] Add tests
- [x] Support source map application
- [x] Refactor `core/source_maps.rs`
- [x] Add test for snapshots
- [x] Support snapshots


bartlomieju (commented):
Very cool, thanks for working on that. I will give it a thorough review.

>  Support snapshots

For me this is the requirement to land this PR. We need to nail down snapshot support as this is a pain point in `ext/node` in `deno` that uses a lot of TypeScript.

> Maybe refactor core/source_maps.rs

Is there anything in particular that stands out?

core/extensions.rs
```diff
@@ -31,6 +31,7 @@ pub enum ExtensionFileSourceCode {
 pub struct ExtensionFileSource {
   pub specifier: &'static str,
   pub code: ExtensionFileSourceCode,
+  pub source_map: Option<Vec<u8>>,
```
bartlomieju, line 34: We should use `ExtensionFileSourceCode` here instead - otherwise source maps will always end up in compiled binary even if they are snapshotted.

bartlomieju:
@Digifox03 it looks like there are still some git conflicts. Can you please resolve them?

testing/checkin/runtime/__init.js
```diff
@@ -16,6 +17,7 @@ globalThis.setInterval = timers.setInterval;
 globalThis.clearTimeout = timers.clearTimeout;
 globalThis.clearInterval = timers.clearInterval;
 globalThis.Worker = worker.Worker;
+globalThis.throwInExt = throwInExt.throwInExt;
```
mmastrac, line 20: Note that we already have a number of "throwing" functions -- you can probably just use `throwErrorWithContextSync` and `throwErrorWithContextAsync` from `checkin:error`

Digifox03: I could reuse the existing throwing functions, but since the line numbers are hardcoded into the test, the test would break whenever the functions in `checkin:error` are moved around.

core/runtime/snapshot_util.rs
```diff
@@ -232,11 +258,22 @@ pub(crate) fn set_snapshotted_data(
     scope.add_context_data(local_context, local_handled_promise_rejection_cb);
   assert_eq!(offset, JS_HANDLED_PROMISE_REJECTION_CB_DATA_INDEX);
 
+  let mut ext_source_maps =
+    Vec::with_capacity(snapshotted_data.ext_source_maps.len() * 2);
+  for (key, value) in snapshotted_data.ext_source_maps.iter() {
+    let value = std::str::from_utf8(value).unwrap();
```
Digifox03, line 264: Is there a better way to do this? Maybe using ArrayBuffers instead of converting to a string?

mmastrac: I think that we should switch to using explicit `serde_v8` serialization for the snapshot. I've added some benchmarks so we can run some experiments here, just haven't had time to experiment with it.

For now, we can probably land with strings and then re-address all of the snapshot sidecar data.

mmastrac:
This is looking good -- could you run `cargo bench --bench snapshot` before and after your changes to see what the snapshot startup impact is? I have some ideas if we see additional startup time here.

Digifox03:
Here is it the result in the `main` branch
```
take snapshot (empty)   time:   [12.632 ms 12.681 ms 12.734 ms]
Found 8 outliers among 50 measurements (16.00%)
  6 (12.00%) high mild
  2 (4.00%) high severe

take snapshot           time:   [14.567 ms 14.607 ms 14.648 ms]

load snapshot           time:   [1.4476 ms 1.4539 ms 1.4607 ms]
Found 1 outliers among 50 measurements (2.00%)
  1 (2.00%) high severe
```

Then in the `digifox03:source_maps` branch
```
take snapshot (empty)   time:   [12.644 ms 12.709 ms 12.778 ms]
                        change: [-0.4654% +0.2216% +0.9818%] (p = 0.52 > 0.05)
                        No change in performance detected.
Found 1 outliers among 50 measurements (2.00%)
  1 (2.00%) high mild

take snapshot           time:   [14.716 ms 14.767 ms 14.820 ms]
                        change: [+0.6381% +1.0942% +1.5670%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 1 outliers among 50 measurements (2.00%)
  1 (2.00%) high mild

load snapshot           time:   [1.4547 ms 1.4605 ms 1.4671 ms]
                        change: [-1.2285% +0.2846% +1.4645%] (p = 0.74 > 0.05)
                        No change in performance detected.
Found 4 outliers among 50 measurements (8.00%)
  3 (6.00%) high mild
  1 (2.00%) high severe
```

The benchmarks do not involve source maps, so at least in the case the feature isn't used, there is no performance penality.

mmastrac:
Looks like a reasonable performance impact -- we can certainly take a look at this in the future as we use this feature in Deno.

Digifox03:
I've noticed that `get_source_line` is only used in `error.rs` behind an `if !file_name.trim_start_matches('[').starts_with("ext:")`.
So I guess it is not necessary to support that for extensions, right?

mmastrac:
> So I guess it is not necessary to support that for extensions, right?

I am not 100% familiar with this code but I imagine this was an optimization to avoid doing sourcemap lookups for extension code. We can probably remove this.

mmastrac:
@Digifox03 I just landed #569 which will require some small changes here, but hopefully nothing too difficult. The snapshot sidecar data is now serde-serialized, and there's a fairly easy option to just toss any `v8::Data` subclass into the snapshot.

Let me know if you need some guidance to migrate over to this serialization style.

Digifox03:
Apparently, the check that skipped reading the source lines of extension is intended, as that is used to show the line of user code which caused the exception; it wouldn't be much usefull if instead it showed extension code.

testing/checkin/runner/mod.rs
```diff
@@ -75,13 +77,59 @@ deno_core::extension!(
...
     "checkin:worker" = "worker.ts",
+    "checkin:throwInExt" = "throwInExt.ts",
   ],
   state = |state| {
     state.put(TestFunctions::default());
     state.put(TestData::default());
   }
 );
 
+fn create_runtime_with_snapshot(
```
mmastrac, line 88: No need to split this -- we should run all the tests with a snapshot. Let's split this change out from this PR, however and land it seperately.

core/benches/snapshot/snapshot.rs
```diff
@@ -17,8 +19,8 @@ macro_rules! fake_extensions {
           deno_core::extension!(
             $name,
             ops = [ ops::$name ],
-            esm_entry_point = concat!("ext:", stringify!($name), "/file.js"),
-            esm = [ dir "benches/snapshot", "file.js", "file2.js" ]
+            esm_entry_point = concat!("ext:", stringify!($name), "/file.ts"),
+            esm = [ dir "benches/snapshot", "file.ts", "file2.ts" ]
```
mmastrac, line 23: I think we should split this into a typescript and non-typescript bench so we can see the relative impact of transpilation.

core/source_map.rs
```diff
@@ -92,7 +94,9 @@ impl<G: SourceMapGetter> SourceMapper<G> {
     let getter = self.getter.as_ref();
     let maybe_source_map =
       self.maps.entry(file_name.to_owned()).or_insert_with(|| {
-        let raw_source_map = getter.and_then(|it| it.get_source_map(file_name));
+        let raw_source_map = None
+          .or_else(|| self.ext_source_maps.get(file_name).cloned())
```
mmastrac, line 98: I worry this clone here could be expensive. Can we get the SourceMap from the slice here?

Digifox03: I don't think that is necessary for a few reasons:
1. `SourceMapGetter::get_source_map` already returns a `Vec<v8>` object, which most likely already performs a clone internally
2. The resulting `SourceMap` gets cached, therefore this code runs only once per source map.
3. The code is (as far as I know) already on a cold path that is only reached while displaying an error.

mmastrac: Regardless, I'm pretty sure we're doing an extra clone that doesn't need to happen.

Let's see if we can clean this logic up a bit -- can you try to use `if let Some(...)` instead of `or_else` chaining?

I'll take a look but I think we may be able to change the `SourceMapGetter` to avoid the current required cloning, that we can land in a future PR.

testing/integration/error_ext_stack/error_ext_stack.out
```diff
@@ -0,0 +1,4 @@
+[ERR] Error: Failed
+[ERR]     at innertTrowInExt (checkin:throwInExt:11:9)
+[ERR]     at throwInExt (checkin:throwInExt:7:3)
+[ERR]     at test:///integration/error_ext_stack/error_ext_stack.ts:3:1
```
mmastrac, line 4: I believe this line number is wrong -- is this an existing bug?

Digifox03: I don't know. `error_with_stack` presents the same issue, the line number is off by one.

mmastrac: I suspect we have something wrong in the checkin transpiler. We can fix this one later.

core/source_map.rs
```diff
@@ -94,10 +94,13 @@ impl<G: SourceMapGetter> SourceMapper<G> {
     let getter = self.getter.as_ref();
     let maybe_source_map =
       self.maps.entry(file_name.to_owned()).or_insert_with(|| {
-        let raw_source_map = None
-          .or_else(|| self.ext_source_maps.get(file_name).cloned())
-          .or_else(|| getter.and_then(|g| g.get_source_map(file_name)));
-        SourceMap::from_slice(&raw_source_map?).ok()
+        None
+          .or_else(|| {
+            SourceMap::from_slice(self.ext_source_maps.get(file_name)?).ok()
```
mmastrac, line 99: Great! 👍

testing/checkin/runtime/throwInExt.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * This is needed to test that stack traces in extensions are correct.
+ */
+export function throwInExt() {
+  innertTrowInExt();
+}
+
+function innertTrowInExt() {
```
mmastrac, line 10:
```suggestion
function innertThrowInExt() {
```

testing/checkin/runner/mod.rs
```diff
@@ -77,6 +77,7 @@ deno_core::extension!(
     "checkin:testing" = "testing.ts",
     "checkin:timers" = "timers.ts",
     "checkin:worker" = "worker.ts",
+    "checkin:throwInExt" = "throwInExt.ts",
```
mmastrac, line 80: Rename to `checkin:throw` for consistency

mmastrac (commented):
I don't see any major problems here stopping us from landing this.

Can you tweak the name of the `checkin:throwInExt` module and re-run benchmarks on main vs this branch? After that, we can land it!

Thanks for all the work on this PR, it looks great.

Digifox03:
On main
```
take snapshot (empty)   time:   [12.521 ms 12.573 ms 12.639 ms]
take snapshot           time:   [14.153 ms 14.217 ms 14.297 ms]
load snapshot           time:   [1.4354 ms 1.4411 ms 1.4513 ms]
```

On PR
```
take snapshot (empty)   time:   [12.479 ms 12.506 ms 12.544 ms]
                        change: [-1.0854% -0.5286% -0.0012%] (p = 0.05 > 0.05)
                        No change in performance detected.
take snapshot/plain     time:   [14.110 ms 14.177 ms 14.282 ms]
take snapshot/transpiled
                        time:   [14.558 ms 14.582 ms 14.609 ms]
load snapshot/plain     time:   [1.4205 ms 1.4271 ms 1.4365 ms]
load snapshot/transpiled
                        time:   [1.6588 ms 1.6615 ms 1.6648 ms]
```

mmastrac (approved):
LGTM. Thanks for this great feature.

Issue #527, opened 2024-01-31
v8_static_strings module should use v8::OneByteConst and v8::String::new_from_onebyte_const


bartlomieju:
These appear to be even cheaper than `v8::String::new_external_onebyte_static`

PR #531, opened 2024-02-01
refactor: use v8::OneByteConst

bartlomieju:
Closes https://github.com/denoland/deno_core/issues/527

Not sure I'm happy with having `v8_static_strings::new_from_static_str`. Maybe all op declarations
should provide their names as `v8::OneByteConst`. Also `v8_static_strings` should be renamed to
`v8_const_strings`?


mmastrac (approved):
LGTM

mmastrac:
> Maybe all op declarations should provide their names as v8::OneByteConst

I think that's a good idea. It's a const function, so it's effectively free for us to just generate `name_v8` next to `name`.

bartlomieju:
> > Maybe all op declarations should provide their names as v8::OneByteConst
> 
> I think that's a good idea. It's a const function, so it's effectively free for us to just generate `name_v8` next to `name`.

Argh.... upon further inspection it seems to be impossible right now as `v8::OneByteConst` doesn't implement `Clone` nor `Copy`. Maybe it's something that could be implemented in `rusty_v8`?

mmastrac:
> Not sure I'm happy with having v8_static_strings::new_from_static_str

Actually, we can probably just add a `deref` to `&'static str` for `&'static OneByteConst`, which would allow us to store `&'static OneByteConst` on ops and ExtensionFileSource...

mmastrac:
Looks like 

> Argh.... upon further inspection it seems to be impossible right now as `v8::OneByteConst` doesn't implement `Clone` nor `Copy`. Maybe it's something that could be implemented in `rusty_v8`?

That should be OK for now as `&'static OneByteConst` is `Copy + Clone`.

bartlomieju:
We'll do the necessary changes in a follow up PR.

PR #532, opened 2024-02-01
Refactor: encapsulate source map logic in SourceMapper

Digifox03:
This is necessary for #530


core/runtime/jsruntime.rs
```diff
@@ -375,8 +375,10 @@ pub type CompiledWasmModuleStore = CrossIsolateStore<v8::CompiledWasmModule>;
 /// Internal state for JsRuntime which is stored in one of v8::Isolate's
 /// embedder slots.
 pub struct JsRuntimeState {
-  pub(crate) source_map_getter: Option<Rc<Box<dyn SourceMapGetter>>>,
-  pub(crate) source_map_cache: Rc<RefCell<SourceMapCache>>,
+  pub(crate) source_mapper: RefCell<SourceMapper<Box<dyn SourceMapGetter>>>,
+  // This is not the right place for this, but it's the easiest way to make
+  // op_apply_source_map a fast op. This stashing should happen in #[op2].
+  pub(crate) stashed_file_name: RefCell<Option<String>>,
```
bartlomieju, line 381: Can't we move it to `SourceMapper`?

Digifox03: I think it makes more sense in the runtime state rather than the source mapper. That is because the source mapper's job is to provide and cache source maps. A more appropriate place for it could be in OpState or as an element of the GothamState.

bartlomieju: I'm not sure I agree - previously this field lived on the `SourceMapCache` and it's limited in use for source mapping. It feels quite natural to "hide it" in `SourceMapper`

core/runtime/jsruntime.rs
```diff
@@ -608,8 +607,7 @@ impl JsRuntime {
     let waker = op_state.waker.clone();
     let op_state = Rc::new(RefCell::new(op_state));
     let state_rc = Rc::new(JsRuntimeState {
-      source_map_getter: options.source_map_getter.map(Rc::new),
-      source_map_cache: Default::default(),
+      source_mapper: RefCell::new(SourceMapper::new(options.source_map_getter)),
```
bartlomieju, line 610: I'm a bit confused here - now we always create `SourceMapper` even if `options.source_map_getter` is not provided. Shouldn't we make `SourceMapper` an `Option` here?

Digifox03: There are two reasons I think it is better to not have it as an `Option`
1. When support for extension's source map is present, it will be always present, unless there are no extensions with typescript.
2. It encapsulates the behaviour of when no source map is present, which allows to avoid using `if Some(source_mapper) = ... { ... } else { Unchanged }` at all use sites.

bartlomieju: Okay, that's reasonable.

core/error.rs
```diff
@@ -331,28 +316,15 @@ impl JsError {
...
-              source_line_frame_index = Some(i);
-              break;
-            }
+      let mut source_mapper = state.source_mapper.borrow_mut();
+      for (i, frame) in frames.iter().enumerate() {
+        if let (Some(file_name), Some(line_number)) =
+          (&frame.file_name, frame.line_number)
+        {
+          if !file_name.trim_start_matches('[').starts_with("ext:") {
+            source_line = source_mapper.get_source_line(file_name, line_number);
```
bartlomieju, line 325: Heads up for the follow up PR that this line will need to be adjusted to work for extension code

bartlomieju (approved):
LGTM, nice cleanup. Thank you @Digifox03

Issue #533, opened 2024-02-01
Move "opcall_test.ts" from "deno" repo to here


bartlomieju:
This file: 
https://github.com/denoland/deno/blob/main/cli/tests/unit/opcall_test.ts
Is testing stuff in `deno` which is actually a `deno_core` interface. We should move these tests to `deno_core` repo.

PR #536, opened 2024-02-01
chore(core): port Deno resource tests and re-organize error test code

mmastrac:
- Moves the resource test from Deno to cover #533 
 - Moves some of the error ops around to centralize them

### DIFF
--- a/testing/checkin/runtime/__init.js
+++ b/testing/checkin/runtime/__init.js
@@ -3,10 +3,12 @@ import "ext:checkin_runtime/__bootstrap.js";
 import * as async from "checkin:async";
 import * as testing from "checkin:testing";
 import * as console from "checkin:console";
+import * as error from "checkin:error";
 import * as timers from "checkin:timers";
 import * as worker from "checkin:worker";
 testing;
 async;
+error;
 
 globalThis.console = console.console;
 globalThis.setTimeout = timers.setTimeout;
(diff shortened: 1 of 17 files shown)


testing/checkin/runtime/__init.js
```diff
@@ -3,10 +3,12 @@ import "ext:checkin_runtime/__bootstrap.js";
 import * as async from "checkin:async";
 import * as testing from "checkin:testing";
 import * as console from "checkin:console";
+import * as error from "checkin:error";
 import * as timers from "checkin:timers";
 import * as worker from "checkin:worker";
 testing;
 async;
+error;
```
bartlomieju, line 11: I must have missed that. Why do we need these here?

testing/checkin/runtime/__init.js
```diff
@@ -47,3 +49,6 @@ Reflect.defineProperty(globalThis, "onrejectionhandled", {
 });
 Deno.unrefTimer = timers.unrefTimer;
 Deno.refTimer = timers.refTimer;
+
+// TODO(mmastrac): arguably this could be registered in 00_infra.js since it's a deno_core error
+Deno.core.registerErrorClass("BadResource", Deno.core.BadResource);
```
bartlomieju, line 54: Let's register it properly in a follow up PR

bartlomieju (approved):

testing/checkin/runtime/__init.js
mmastrac: These need to be imported and then ignored to avoid the "unused import" error from the ExtModuleLoader. I suppose I can add a lint ignore for them being unused.

I think that the `node:` modules don't require this because we manually export them in `01_require.js`.

PR #538, opened 2024-02-01
fix(core): Ensure op driver is shut down unconditionally

mmastrac:
We needed to revert the resource sanitizer in https://github.com/denoland/deno/pull/22125 because of a panic in third-party test code.

The problem was that in certain cases, if the OpDriver was kept alive either by Rc<ContextState> (in older code) or Rc<OpDriverImpl> (in newer code), the Tokio task that polls the underlying ops would potentially poll an op that expected something to be in OpState, but we had already removed all the items in the OpState by that point.


core/runtime/jsruntime.rs
```diff
@@ -1219,9 +1222,8 @@ impl JsRuntime {
 
   /// Returns the runtime's op names, ordered by OpId.
   pub fn op_names(&self) -> Vec<&'static str> {
-    let main_realm = self.inner.main_realm.clone();
-    let state_rc = main_realm.0.state();
-    state_rc.op_ctxs.iter().map(|o| o.decl.name).collect()
+    let state = &self.inner.main_realm.0.context_state;
```
mmastrac, line 1225: This was unintentionally causing a leak of the realm. `ManuallyDrop::clone` is bad. I wish we could forbid it somehow.

bartlomieju: I think you could use `clippy.toml` and `disallowed-methods` there?

mmastrac: Oh, let me try that out...

core/runtime/op_driver/joinset_driver.rs
bartlomieju, line 1: There's a reference to this driver in `core/runtime/jsrealm.rs` and accompanying Cargo features

mmastrac: Ah thanks. Will remove.

bartlomieju (approved):

PR #541, opened 2024-02-02
Revert "feat: Add bindings to analyse WASM modules (#528)"

bartlomieju:
This reverts commit 012a28039104cddf69c67b87ef4d9829a7418ec8.

This functionality is not longer necessary, because we now have
https://crates.io/crates/wasm_dep_analyzer


mmastrac (approved):
LGTM!

PR #543, opened 2024-02-02
chore(core): remove ExtensionBuilder

mmastrac:
Removes `ExtensionBuilder` which has been unsupported and untested for almost a year. Also removes `init_js_only` which has been deprecated for nearly as long.


bartlomieju (approved):
LGTM.

Before landing, could you do a quick grep with `cs.github.com` and see if there are any repos actively using it?

PR #544, opened 2024-02-02
 refactor(core): clean up include_js_files! macros 

mmastrac:
This looks pretty big, but it's mostly a cleanup of `include_js_files!` that makes it easier to understand and easier to maintain moving forward.

`include_js_files!` supports four styles of includes:

 - `"file"` <- old
 - `"file" with_specifier "specifier"` <- old, triggers a deprecation warning with a hint
 - `specifier = "file"` <- new
 - `specifier = { source = "source" }` <- new, see below

The `specifier = "file"` syntax will be preferred to replace the current `"file" with_specifier "specifier"` syntax, but won't be removed yet. If we want to add any other possible options to this macro, it's just too hard to expand the old syntax (which is why I introduced this new syntax!).

We now support an inline, embeddable string script for all places where `include_js_files!` is used:

```
    deno_core::extension!(
      module_snapshot,
      esm_entry_point = "mod:test",
      esm = ["mod:test" =
        { source = "globalThis.TEST = 'foo'; export const TEST = 'bar';" },]
    );
```

Changes:
 - `ExtensionFileSource` cannot be constructed using the bare struct syntax anymore (a new field named `_unconstructable_use_new` helps suggest the right course of action)
 - `ExtensionFileSource` has new `const` constructors 
 - `include_js_files!` macro and related macros have been rewritten

### DIFF
--- a/core/extensions.rs
+++ b/core/extensions.rs
@@ -5,16 +5,18 @@ use crate::OpState;
 use anyhow::Context as _;
 use anyhow::Error;
 use std::borrow::Cow;
+use std::marker::PhantomData;
 use std::sync::Arc;
 use v8::fast_api::FastFunction;
 use v8::MapFnTo;
 
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub enum ExtensionFileSourceCode {
   /// Source code is included in the binary produced. Either by being defined
   /// inline, or included using `include_str!()`. If you are snapshotting, this
   /// will result in two copies of the source code being included - one in the
   /// snapshot, the other the static string in the `Extension`.
+  #[deprecated = "Use ExtensionFileSource::new"]
   IncludedInBinary(&'static str),
 
   // Source code is loaded from a file on disk. It's meant to be used if the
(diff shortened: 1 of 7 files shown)


core/runtime/tests/misc.rs
```diff
@@ -1058,41 +1050,34 @@ async fn tla_in_esm_extensions_panics() {
...
+  deno_core::extension!(
+    test_ext,
+    ops = [op_wait],
+    esm_entry_point = "mod:test",
+    esm = [
+      "mod:test" = { source = "import 'mod:error';" },
+      "mod:error" = {
+        source = r#"
+        function a() { throw new Error("This fails") };
+        a();
```
bartlomieju, line 1071: These will have leading spaces, right?

core/extensions.rs
```diff
@@ -5,41 +5,109 @@ use crate::OpState;
...
 use v8::MapFnTo;
 
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub enum ExtensionFileSourceCode {
   /// Source code is included in the binary produced. Either by being defined
   /// inline, or included using `include_str!()`. If you are snapshotting, this
   /// will result in two copies of the source code being included - one in the
   /// snapshot, the other the static string in the `Extension`.
+  #[deprecated = "Use ExtensionFileSource::new"]
```
bartlomieju, line 19: Is it only this variant that gets deprecated?

bartlomieju (approved):
To be extra sure, could you open a PR to `deno` that uses this branch and see how hard will the migration be?

core/runtime/tests/misc.rs
mmastrac: Correct. All of these embedded raw strings have an extra indent.

bartlomieju: Guess we're gonna use them mostly for testing, so that seems fine.

PR #545, opened 2024-02-02
feat(core): refactor ExtModuleLoader to log only when no execution errors

mmastrac:
`ExtModuleLoader` was panicking on Drop if modules were not evaulated, but it's better to make this an explicit operation instead. This way we can get the extension errors as they happen.


bartlomieju (approved):
Seems reasonable

PR #547, opened 2024-02-02
refactor: Prepare for Wasm imports support

bartlomieju:
This commit prepares for support of WASM modules.

A new variant of "ModuleType::Wasm" was added.

Actually trying to load a WASM module will throw an error.


mmastrac (approved):
LGTM

PR #555, opened 2024-02-05
refactor: destructure ops in 02_error.js

bartlomieju:
Working on https://github.com/denoland/deno/pull/22273 I found that we can't remove
`op_format_file_name`, `op_apply_source_map` and `op_apply_source_map_filename`
from `Deno.core.ops`, because they are used in `02_error.js`.


mmastrac (approved):
LGTM

PR #557, opened 2024-02-05
feat: Add `Deno.core.createCancelHandle()`

bartlomieju:
This is a common utility used in `ext/fs` and `ext/net` in `deno` repo.

It's weirdly created in `ext/web` so it can be shared between the two extensions.

### DIFF
--- a/core/01_core.js
+++ b/core/01_core.js
@@ -429,6 +429,7 @@
     op_timer_ref,
     op_timer_unref,
     op_unref_op,
+    op_cancel_handle,
 
     op_is_any_array_buffer,
     op_is_arguments_object,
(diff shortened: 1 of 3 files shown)


mmastrac (approved):
LGTM, but I am not sure we want to use this API forever -- it would actually be a lot better as a cppgc object. Happy to pull this code into deno_core for now at least.

bartlomieju:
> LGTM, but I am not sure we want to use this API forever -- it would actually be a lot better as a cppgc object. Happy to pull this code into deno_core for now at least.

That's a good idea. But let's pull the code to limit the blast radius in `deno`.

PR #559, opened 2024-02-06
chore(core): Add benchmark for snapshots

mmastrac:


bartlomieju (approved):
LGTM!

Let's figure out a way to collect this data in CI and create a simple page to track these numbers

PR #563, opened 2024-02-07
fix: disabling an op, preserves no of external references

bartlomieju:


core/extensions.rs
```diff
@@ -205,8 +207,26 @@ impl OpDecl {
...
+        })
+      } else {
+        None
+      },
+      fast_fn_with_metrics: if self.fast_fn_with_metrics.is_some() {
+        Some(FastFunction {
+          args: &[],
+          function: noop as _,
+          repr: v8::fast_api::Int64Representation::Number,
+          return_type: v8::fast_api::CType::Bool,
```
mmastrac, line 225:
```suggestion
          return_type: v8::fast_api::CType::Void,
```

core/extensions.rs
```diff
@@ -156,6 +156,8 @@ pub type GlobalTemplateMiddlewareFn =
 pub type GlobalObjectMiddlewareFn =
   for<'s> fn(&mut v8::HandleScope<'s>, v8::Local<'s, v8::Object>);
 
+fn noop() {}
```
mmastrac, line 159:
```suggestion
extern "C" fn noop() {}
```

mmastrac (approved):
LGTM. Note that the fast function for disabled functions won't throw. Maybe we should add a TODO?

bartlomieju:
> LGTM. Note that the fast function for disabled functions won't throw. Maybe we should add a TODO?

Is it even possible this will happen? If the disabled op is called an it throws it won't get optimized to become a fast fn? I'll add a TODO anyway.

PR #569, opened 2024-02-10
feat: Use serde_json and Snapshot*DataStore for managing snapshot sidecar data

mmastrac:
This code will be used for the future opcall tracing work, as we need to get a reference to a number of primordials on the Rust side and persist those references in the snapshot.

This drastically simplifies our management of snapshot sidecar data. 

As a positive side-effect, the management of module map data becomes trivially easy. In the future, we may decide to serialize that serde_v8 data using a more efficient protocol, but that's left as a future task.

The sidecar data is serialized in the snapshot by using a single, nested serde_v8 struct and a collection of `v8::Data` objects that are referenced by ID. Because we cannot use serde_v8 for `v8::Data` objects like `v8::Module`, these are simply aggregated and stored in successive indexes of the snapshot per-context data.

It's a shame we don't have the ability to create a `v8::FixedArray` because that would be ideal for this.

`cargo benchmark --bench snapshot` shows that the snapshot load time difference is within the noise margin.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,6 +118,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "bincode"
+version = "1.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1f45e9417d87227c7a56d22e471c6206462cba514c7590c09aff4cf6d1ddcad"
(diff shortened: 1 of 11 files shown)


littledivy (approved):
Looks good to me

PR #573, opened 2024-02-12
refactor(core): opcall tracing in Rust

mmastrac:
Moves the core of the opcall tracing functionality into Rust. Tested against deno (which will need small source changes).

Re-organizes the op dispatch code a bit as well, but performance is unimpacted per benchmarks (if not slightly faster).

### DIFF
--- a/core/00_infra.js
+++ b/core/00_infra.js
@@ -123,7 +123,7 @@
     const promise = new Promise((resolve) => {
       promiseRing[idx] = resolve;
     });
-    let wrappedPromise = PromisePrototypeThen(
+    const wrappedPromise = PromisePrototypeThen(
       promise,
       unwrapOpError(),
     );
@@ -180,7 +180,7 @@
       if (res.code) {
         err.code = res.code;
       }
-      // Strip unwrapOpResult() and errorBuilder() calls from stack trace
+      // Strip eventLoopTick() calls from stack trace
       ErrorCaptureStackTrace(err, eventLoopTick);
       throw err;
     };
--- a/core/00_infra.js
+++ b/core/00_infra.js
@@ -21,23 +21,12 @@
     RangeError,
     ReferenceError,
     SafeMap,
-    SafePromisePrototypeFinally,
-    StringPrototypeSlice,
     StringPrototypeSplit,
     SymbolFor,
     SyntaxError,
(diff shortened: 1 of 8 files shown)


core/runtime/jsrealm.rs
```diff
@@ -56,6 +58,8 @@ pub(crate) struct ContextState {
   pub(crate) wasm_instantiate_fn: RefCell<Option<Rc<v8::Global<v8::Function>>>>,
   pub(crate) unrefed_ops:
     RefCell<HashSet<i32, BuildHasherDefault<IdentityHasher>>>,
+  pub(crate) opcall_tracing_enabled: Cell<bool>,
+  pub(crate) opcall_traces: RefCell<HashMap<i32, String>>,
```
dsherret, line 62: In the future for perf, maybe it should be extracted out into a `struct FastValueCloneMapCell<T, U: Clone>(UnsafeCell<HashMap<T, U>>)` (or some other name) that clones the value on get in order to reduce the RefCell overhead.

dsherret (approved):
Nice, LGTM!

core/00_infra.js
```diff
@@ -166,27 +140,17 @@
...
+        throw "Missing promise in ring @ " + promiseId;
+      }
+      promiseRing[idx] = NO_PROMISE;
+      promise(res);
     }
-    promiseRing[idx] = NO_PROMISE;
-    return promise;
-  }
-
-  function newPromise() {
```
bartlomieju, line 181: Why is this method gone?

mmastrac: It got absorbed into the `setPromise` method to avoid method call overhead.

mmastrac: We skip the intermediate value for the promise resolver and just place the resolve method directly into the promise map now:

https://github.com/denoland/deno_core/pull/573/files#diff-c65a34b915562b309be0123993a975d6f83222c66390db083f403409d9251430R123

bartlomieju: Huh, did we never use `reject` method?

Side note: we should bench and see if using `Promise.withResolvers` is faster than doing it manually

mmastrac: I did that a while back and it ended up being slower 😢

PR #576, opened 2024-02-13
refactor(core): reduce cloning in opcall tracing

mmastrac:
Address @dsherret's comments a bit.

### DIFF
--- a/core/runtime/stats.rs
+++ b/core/runtime/stats.rs
@@ -8,8 +8,51 @@ use crate::PromiseId;
 use crate::ResourceId;
 use bit_set::BitSet;
 use serde::Serialize;
+use std::cell::Cell;
 use std::cell::RefCell;
+use std::collections::HashMap;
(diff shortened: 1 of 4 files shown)


core/runtime/stats.rs
```diff
@@ -8,9 +8,62 @@ use crate::PromiseId;
...
+use std::cell::Cell;
 use std::cell::RefCell;
+use std::collections::HashMap;
+use std::ops::Deref;
 use std::rc::Rc;
 
+#[derive(Default)]
+pub struct OpCallTraces {
+  enabled: Cell<bool>,
+  traces: RefCell<HashMap<PromiseId, Rc<str>>>,
```
mmastrac, line 21: We need to keep this RefCell because the more efficient get/get_all might allow the user to call back into this struct. It's unlikely but possible.

dsherret (approved):
LGTM!

PR #583, opened 2024-02-16
fix(core): don't leak sidecar data memory

mmastrac:
Split the sidecar data into a separate part of the output snapshot data so we don't need to put that data into the snapshot, which potentially leaks it. 

This also opens up the ability for us to have statically-referenced data in the sidecar snapshot which will make for some zero-copy features.


core/runtime/jsruntime.rs
```diff
@@ -735,9 +731,10 @@ impl JsRuntime {
       );
 
       // Get module map data from the snapshot
-      if init_mode.is_from_snapshot() {
-        snapshotted_data =
-          Some(snapshot_util::get_snapshotted_data(scope, context));
+      if let Some(raw_data) = sidecar_data {
+        snapshotted_data = Some(snapshot::load_snapshotted_data_from_snapshot(
+          scope, context, raw_data,
+        ));
```
bartlomieju, line 737: Yeah, this is cleaner

core/runtime/jsruntime.rs
```diff
@@ -1856,7 +1853,8 @@ impl JsRuntimeForSnapshot {
     }
 
     // Serialize the module map and store its data in the snapshot.
-    {
+    // TODO(mmastrac): This should deconstruct the realm completely.
```
bartlomieju, line 1856: I don't really understand this comment

core/runtime/snapshot.rs
```diff
@@ -0,0 +1,577 @@
...
+  fn process_chunk(&mut self, chunk: &[u8]) -> std::io::Result<()> {
+    self.file.write_all(chunk)
+  }
+
+  fn finalize(self) -> std::io::Result<Self::Output> {
+    Ok(self.file)
+  }
+}
+
+/// Serializes the snapshot to another serializer, with a bulk compression operation performed on top.
```
bartlomieju, line 245: What is bulk compression in this context?

core/runtime/snapshot.rs
```diff
@@ -0,0 +1,577 @@
...
+}
+
+pub(crate) struct SnapshottedData {
+  pub module_map_data: ModuleMapSnapshotData,
+  pub js_handled_promise_rejection_cb: Option<v8::Global<v8::Function>>,
+  pub ext_source_maps: HashMap<String, Vec<u8>>,
+}
+
+#[derive(Serialize, Deserialize)]
+pub(crate) struct RawSnapshottedData {
```
bartlomieju, line 456: Not a fan of this name - I don't see what's "raw" about this data - maybe `AuxSnapshotData`?

bartlomieju (approved):

core/runtime/snapshot.rs
mmastrac: I stole the term from `zstd` which we use in the LSP snapshot: `&zstd::bulk::compress(snapshot_slice, 22)`

core/runtime/snapshot.rs
mmastrac: I can rename this

core/runtime/jsruntime.rs
mmastrac: Expanded a bit

core/runtime/snapshot.rs
mmastrac: This is now `SerializableSnapshotSidecarData` to indicate that its intended destination is serde.

PR #586, opened 2024-02-20
fix(snapshot): execute `with_runtime_callback` on warmup runtime

littledivy:
Fixes `node:vm` missing context panic in https://github.com/denoland/deno/pull/22386


bartlomieju (approved):
Is there a regression test that you could add? Just so we don't scratch our heads next time around this bit.

Issue #589, opened 2024-02-21
CONTEXT_SETUP_SOURCES should use OneByteConst and add external refs


bartlomieju:
First step to lowering startup time.

PR #591, opened 2024-02-21
refactor: Use v8::OneByteConst for internals scripts

bartlomieju:
This commit changes internal JS source code of "deno_core"
to be included in the binary. We now use "v8::OneByteConst" and
pass external references to the sources which allows V8 to take
an optimized path when deserializing the snapshot. Local tests
with flamegraphs confirm that it's now cheaper to deserialize the
snapshot.

Closes https://github.com/denoland/deno_core/issues/589

### DIFF
--- a/core/main.rs
+++ b/core/main.rs
@@ -3,10 +3,13 @@
 use anyhow::anyhow;
 use anyhow::Context;
 use deno_core::anyhow::Error;
+use deno_core::snapshot::CreateSnapshotOptions;
+use deno_core::snapshot::SnapshotFileSerializer;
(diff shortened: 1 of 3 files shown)


littledivy (approved):
LGTM!

One thing to note is that `new_from_onebyte_const` runs a `const`-time 7-bit ascii test. This is build time overhead in MIRI. I noticed during my PoC that build time increased over a few seconds when adding 01_crypto.js, 01_webgpu.js and tsc main compiler.

mmastrac:
We could make that one-byte check debug only, wdyt?

PR #593, opened 2024-02-21
[WIP] refactor: add v8::OneByteConst to FastString::StaticAscii

bartlomieju:
This might be not needed, but I'm trying to figure out how to pass external references to these strings.

### DIFF
--- a/core/fast_string.rs
+++ b/core/fast_string.rs
@@ -27,7 +27,9 @@ pub enum FastString {
   Static(&'static str),
 
   /// Created from static data, known to contain only ASCII chars.
-  StaticAscii(&'static str),
+  /// For extra optimization when snapshotting, external references should
+  /// be provided for these strings.
+  StaticAscii(&'static str, v8::OneByteConst),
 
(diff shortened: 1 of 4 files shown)


core/fast_string.rs
```diff
@@ -27,7 +27,9 @@ pub enum FastString {
   Static(&'static str),
 
   /// Created from static data, known to contain only ASCII chars.
-  StaticAscii(&'static str),
+  /// For extra optimization when snapshotting, external references should
+  /// be provided for these strings.
+  StaticAscii(&'static str, v8::OneByteConst),
```
mmastrac, line 32: It was a mistake for me to make this enum accessible -- we should make it an inner private member of the FastString struct (named FastStringInner) and then we can initialize it in a method (ie: FastString::static_ascii(...))

bartlomieju: I'm not sure I understand what you mean. What we would initialize in a method?

bartlomieju:
Not needed anymore.

PR #598, opened 2024-02-23
fix(core): ensure `this` receiver in setTimeout and handle errors

mmastrac:


.dprint.json
bartlomieju, line 1: What happened here?

bartlomieju (approved):

.dprint.json
mmastrac: Accident -- will put this in a follow-up PR

PR #600, opened 2024-02-23
chore: checkin type-checks w/tsc

mmastrac:
The lint step will actually type-check checkin and all of the extension code in checkin.

### DIFF
--- a/.github/workflows/ci-lint/action.yml
+++ b/.github/workflows/ci-lint/action.yml
@@ -7,8 +7,14 @@ runs:
       run: |
         cargo fmt -- --check
         deno fmt --check ./core/ ./ops/ ./serde_v8/
-        npm install -g typescript
-        tsc --noEmit -p testing/tsconfig.json
+    - name: tsc
+      shell: bash
+      run: |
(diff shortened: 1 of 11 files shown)


.github/workflows/ci-lint/action.yml
```diff
@@ -7,6 +7,14 @@ runs:
       run: |
         cargo fmt -- --check
         deno fmt --check ./core/ ./ops/ ./serde_v8/
+    - name: tsc
+      shell: bash
+      run: |
+        # If npm is installed, run tsc
+        if command -v npm &> /dev/null; then
+          npm install -g typescript
```
bartlomieju, line 15: I'm not super happy about it, but that improves our lint checks then fine :)

bartlomieju (approved):

.github/workflows/ci-lint/action.yml
bartlomieju: Oh well, bootstrapping is hard

mmastrac: Switched this out for deno :D

PR #602, opened 2024-02-24
refactor: Remove deno_core::Snapshot

bartlomieju:
This commit removes `deno_core::Snapshot`.

Embedders need to pass a `&'static [u8]` for the startup
snapshot now.

It was an unnecessary abstraction that complicated things.

To satisfy test requirements, all snapshots are leaked. It's not
a big deal, because they are really small and this setup better
corresponds with the real world usage (and doesn't require
expensive integration tests that would need to compile a full
fledged binary).


mmastrac (approved):
LGTM, much simpler.

PR #604, opened 2024-02-24
feat: Allow to serialize source files into a snapshot

bartlomieju:
This commit changes snapshotting methods to allow to serde
source files.

The pointers to these source files are passed to V8, but we currently
don't use this functionality. I'll try to implement that before landing this PR.


bartlomieju:
Not needed anymore.

PR #607, opened 2024-02-26
feat(dcore): Allow to load snapshot from file

bartlomieju:


mmastrac (commented):
Before we go too far on this I think we should try to merge it into checkin. I've already done a bunch of work to snapshot in there and it's better if we can avoid having a second bit of testing infra to maintain

Issue #612, opened 2024-02-27
Investigate `web_timeout::tests::test_chaos::seed_2_99` flake


mmastrac:
Probably a subtle logic error, has only happened once.

Condition that failed:

`self.timers.borrow().len() != self.data_map.borrow().len() + self.tombstone_count.get()`

```
---- web_timeout::tests::test_chaos::seed_2_99 stdout ----
count=6527 ref_count=7213
5818
6424
6501
6526
thread 'web_timeout::tests::test_chaos::seed_2_99' panicked at core/web_timeout.rs:448:7:
assertion `left == right` failed
  left: 4
 right: 5


failures:
    web_timeout::tests::test_chaos::seed_2_99
```

```

  #[cfg(test)]
  pub fn assert_consistent(&self) {
    if self.data_map.borrow().is_empty() {
      // If the data map is empty, we should have no timers, no tombstones and no unref'd count
      assert_eq!(self.timers.borrow().len(), 0);
      assert_eq!(self.tombstone_count.get(), 0);
      assert_eq!(self.unrefd_count.get(), 0);
    } else {
      assert_eq!(
        self.timers.borrow().len(),
        self.data_map.borrow().len() + self.tombstone_count.get()
      );
      assert!(self.unrefd_count.get() <= self.data_map.borrow().len());
    }
  }
```

PR #613, opened 2024-02-27
fix(core): Fix subtle timer bug

mmastrac:
Fixes https://github.com/denoland/deno_core/issues/612

When we have an ineffective poll, we were failing to decrement the tombstone count. We also didn't check to see if the timers were consistent if polling returning `Pending`, which is why this bug slipped through the net.


core/web_timeout.rs
```diff
@@ -365,15 +370,15 @@ impl<T: Clone> WebTimers<T> {
             timer_type,
           ));
         } else {
-          self.tombstone_count.set(self.tombstone_count.get() - 1);
+          tombstones -= 1;
```
dsherret, line 373: I like this name.

dsherret (approved):
LGTM

PR #618, opened 2024-02-28
chore: add a tools/check.ts for lint and format

mmastrac:
Usage:
 - `tools/check.ts lint`: runs a lint and reports errors
 - `tools/check.ts lint --fix`: attempts to fix lint problems
 - `tools/check.ts format`: formats the codebase
 - `tools/check.ts format --check`: checks if the codebase is formatted, fails if not

### DIFF
--- a/tools/format.ts
+++ b/tools/format.ts
@@ -1,0 +1,5 @@
+#!/usr/bin/env -S deno run --quiet --allow-read --allow-write --allow-run --allow-env
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+import { main } from "./check.ts";
+
+await main("format", Deno.args[0]);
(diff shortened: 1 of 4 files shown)


bartlomieju (approved):

tools/format.ts
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env -S deno run --quiet --allow-read --allow-write --allow-run --allow-env
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+import { main } from "./check.ts";
+
+await main("format", Deno.args[0]);
```
bartlomieju, line 5: Maybe these could be `format` and `lint` and passing proper strings could be done internally in `tools/check.ts`?. Not a big deal though

mmastrac: I'll punt on this for now but I probably have a few random fixes for these tools and I can do it then.

PR #620, opened 2024-02-29
refactor(core): fast string overhaul

mmastrac:
With this change, `FastString`s created from `ascii_str!` are one-byte const capable.

Because `FastString` may hold a non-static value like a `Box` or `Arc<str>`, we introduce `FastStaticString` as a way to guarantee that a string is compatible with v8 `OneByteConst`s.

To simplify use of the API, the embedder-facing script load/execute functions now take an `impl IntoModuleSourceCode`, allowing us to remove the static string overloads.


core/examples/op2.rs
```diff
@@ -44,10 +44,7 @@ fn main() -> Result<(), Error> {
 
   let future = async move {
     let mod_id = js_runtime
-      .load_main_module(
-        &main_module,
-        Some(deno_core::FastString::from(module_code)),
-      )
+      .load_main_module(&main_module, module_code)
```
bartlomieju, line 47: Nitpick - I'd prefer to make `load_main_module` work on URL and `load_main_module_from_code` to accept a second param. TIOLI

core/runtime/bindings.rs
```diff
@@ -801,7 +801,7 @@ pub fn create_exports_for_ops_virtual_module<'s>(
   op_ctxs: &[OpCtx],
   scope: &mut v8::HandleScope<'s>,
   global: v8::Local<v8::Object>,
-) -> Vec<(FastString, v8::Local<'s, v8::Value>)> {
+) -> Vec<(FastStaticString, v8::Local<'s, v8::Value>)> {
```
bartlomieju, line 804: Do you think it would make sense to add `FastConstString`? Eg for virtual modules we always now upfront the names and can make all the names const strings.

core/runtime/jsruntime.rs
```diff
@@ -1265,10 +1267,9 @@ impl JsRuntime {
     state.op_ctxs.iter().map(|o| o.decl.name).collect()
   }
 
-  /// Executes traditional JavaScript code (traditional = not ES modules).
-  ///
-  /// The execution takes place on the current main realm, so it is possible
-  /// to maintain local JS state and invoke this method multiple times.
+  /// Executes traditional, non-ECMAScript-module JavaScript code, This code executes in
```
bartlomieju, line 1270:
```suggestion
  /// Executes traditional, non-ECMAScript-module JavaScript code, this code executes in
```

core/runtime/jsruntime.rs
```diff
@@ -258,7 +260,7 @@ impl Future for RcPromiseFuture {
   }
 }
 
-const VIRTUAL_OPS_MODULE_NAME: &str = "ext:core/ops";
+static VIRTUAL_OPS_MODULE_NAME: FastStaticString = ascii_str!("ext:core/ops");
```
bartlomieju, line 263: Ditto - I think we could make this const string to make it essentially free?

core/fast_string.rs
```diff
@@ -4,31 +4,159 @@ use serde::Deserializer;
...
 use v8::NewStringType;
 
+static EMPTY_STRING: v8::OneByteConst =
+  v8::String::create_external_onebyte_const("".as_bytes());
+
+/// A static string that is compile-time checked to be ASCII and is stored in the
+/// most efficient possible way to create V8 strings.
+#[derive(Clone, Copy)]
+#[repr(transparent)]
+pub struct FastStaticString {
```
bartlomieju, line 22: Hm, I guess disregard my previous comments. Naming is hard 😓

bartlomieju (approved):

PR #628, opened 2024-03-01
feat: Inspector integration

bartlomieju:
This commit adds Chrome DevTools integration for the "dcore"
binary target. The inspector server can be activated using
"DCORE_INSPECTOR" env var.

The implementation is a copy-paste of existing inspector impl
from "deno" and is meant as a forcing function to refactor and
streamline the implementation. Once we make it agnostic of the
server we might make it as part of "deno_core" library.


foo.js
```diff
@@ -0,0 +1,16 @@
+function setInterval(callback, delay = 0) {
```
mmastrac, line 1: Should remove or move this one before landing

mmastrac (approved):
LGTM. I think we can land this and clean it up as we go. Ideally the bits of this patch that don't deal w/hyper or fastwebsockets would live in deno_core.

core/dcore/inspector_server.rs
```diff
@@ -0,0 +1,498 @@
...
+    _ = deregister_inspector_handler => unreachable!(),
+    _ = server_handler => {},
+  }
+}
+
+/// The pump future takes care of forwarding messages between the websocket
+/// and channels. It resolves when either side disconnects, ignoring any
+/// errors.
+///
+/// The future proxies messages sent and received on a warp WebSocket
```
mmastrac, line 386: warp?

core/dcore/inspector_server.rs
```diff
@@ -0,0 +1,498 @@
...
+/// This works because UnboundedSender/UnboundedReceiver are implemented in the
+/// 'futures' crate, therefore they can't participate in Tokio's cooperative
+/// task yielding.
+async fn pump_websocket_messages(
+  mut websocket: WebSocket<TokioIo<hyper::upgrade::Upgraded>>,
+  inbound_tx: UnboundedSender<String>,
+  mut outbound_rx: UnboundedReceiver<InspectorMsg>,
+) {
+  'pump: loop {
+    tokio::select! {
```
mmastrac, line 400: Maybe make this a set of tokio tasks to simplify this code

bartlomieju: Let's land it first and then I will rip this module apart and rewrite it to something more sane.

PR #630, opened 2024-03-02
refactor: move "dcore" binary target to a separate directory

bartlomieju:
Just to not complicate dependencies within `deno_core` library.

This crate is not publishable and is meant only for development
and testing.


mmastrac (approved):
LGTM

PR #647, opened 2024-03-08
fix(core): don't error out with unref'd ops in TLA

mmastrac:


testing/lib.rs
```diff
@@ -58,6 +58,7 @@ integration_test!(
   error_ext_stack,
   error_with_stack,
   error_without_stack,
+  pending_unref_op_tla,
```
bartlomieju, line 61: Could you paste the current panic out that's happening right now? I don't see a scenario that would panic, but rather deadlock.

bartlomieju (approved):

testing/lib.rs
mmastrac: It's actually "Top-level await promise never resolved" -- the test I added was panicking at one point. :)

mmastrac:
Fixes https://github.com/denoland/deno/issues/20814

PR #650, opened 2024-03-08
Replace usage of derive_more derives with manual implementations

### LINKED ISSUE #640: Consider removing derive_more dependency for serde_v8
This dependency is only used to derive trivial `Deref`, `DerefMut`, `AsRef` and `AsMut` implementations. However, it requires 3.3s to compile on my machine and much more if you count all its dependencies.

Would you accept a PR that replaces these derives with hand-written (or `macro_rules!`-produced) implementations?

Related: https://github.com/denoland/rust-urlpattern/issues/47

svix-jplatte:
I found some of the generated impls like [this one](https://docs.rs/serde_v8/latest/serde_v8/struct.BigInt.html#impl-AsMut%3C__AsMutT%3E-for-BigInt) quite weird so I removed or replaced them with different ones, as seemed reasonable in the specific cases.

Let me know what you think. The individual commit messages have more details on why I chose certain replacements.

Closes #640.


mmastrac (approved):
LGTM, thanks!

PR #655, opened 2024-03-13
feat: support for v8 code cache

igorzi:
This PR adds V8 code cache support for ES modules loaded through `ModuleLoader`, and for scripts evaluated through `evalContext` (used by `require` in Deno). Embedders have full control over whether code caching is enabled, and how code cache is retrieved/stored.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ deno_core = { version = "0.270.0", path = "./core" }
 deno_ops = { version = "0.146.0", path = "./ops" }
 serde_v8 = { version = "0.179.0", path = "./serde_v8" }
 
-v8 = { version = "0.85.0", default-features = false }
+v8 = { version = "0.86.0", default-features = false }
 deno_ast = { version = "=0.32.0", features = ["transpiling"] }
 deno_unsync = "0.3.2"
 deno_core_icudata = "0.0.73"
--- a/core/modules/loaders.rs
+++ b/core/modules/loaders.rs
@@ -87,12 +87,12 @@ pub trait ModuleLoader {
   /// can store the provided code cache for future executions of the same module.
   ///
   /// It's not required to implement this method.
-  fn store_code_cache(
+  fn code_cache_ready(
     &self,
     _module_specifier: &ModuleSpecifier,
     _code_cache: &[u8],
-  ) -> Result<(), Error> {
-    Ok(())
+  ) -> Pin<Box<dyn Future<Output = ()>>> {
+    async {}.boxed_local()
   }
 }
 
--- a/core/modules/loaders.rs
+++ b/core/modules/loaders.rs
@@ -82,6 +82,18 @@ pub trait ModuleLoader {
   ) -> Pin<Box<dyn Future<Output = Result<(), Error>>>> {
     async { Ok(()) }.boxed_local()
   }
+
+  /// Called when new v8 code cache is available for this module. Implementors
+  /// can store the provided code cache for future executions of the same module.
+  ///
+  /// It's not required to implement this method.
+  fn store_code_cache(
+    &self,
+    _module_specifier: &ModuleSpecifier,
+    _code_cache: &[u8],
+  ) -> Result<(), Error> {
+    Ok(())
+  }
 }
 
 /// Placeholder structure used when creating
@@ -186,6 +198,7 @@ impl ModuleLoader for ExtModuleLoader {
       ModuleType::JavaScript,
       ModuleSourceCode::String(source),
       specifier,
+      None,
     )))
   }
 
@@ -240,6 +253,7 @@ impl ModuleLoader for LazyEsmModuleLoader {
       ModuleType::JavaScript,
       ModuleSourceCode::String(source),
       specifier,
+      None,
     )))
   }
 
@@ -316,6 +330,7 @@ impl ModuleLoader for FsModuleLoader {
         module_type,
         ModuleSourceCode::Bytes(code.into_boxed_slice().into()),
         &module_specifier,
+        None,
       );
       Ok(module)
     }
@@ -377,6 +392,7 @@ impl ModuleLoader for StaticModuleLoader {
         ModuleType::JavaScript,
         ModuleSourceCode::String(code.try_clone().unwrap()),
         module_specifier,
+        None,
       ))
     } else {
       Err(generic_error("Module not found"))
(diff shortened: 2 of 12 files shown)


Cargo.toml
```diff
@@ -23,7 +23,7 @@ deno_core = { version = "0.270.0", path = "./core" }
 deno_ops = { version = "0.146.0", path = "./ops" }
 serde_v8 = { version = "0.179.0", path = "./serde_v8" }
 
-v8 = { version = "0.85.0", default-features = false }
+v8 = { version = "0.86.0", default-features = false }
```
igorzi, line 26: FYI - 0.86.0 doesn't exist yet (https://github.com/denoland/rusty_v8/pull/1430)

core/modules/loaders.rs
```diff
@@ -82,6 +82,18 @@ pub trait ModuleLoader {
   ) -> Pin<Box<dyn Future<Output = Result<(), Error>>>> {
     async { Ok(()) }.boxed_local()
   }
+
+  /// Called when new v8 code cache is available for this module. Implementors
+  /// can store the provided code cache for future executions of the same module.
+  ///
+  /// It's not required to implement this method.
+  fn store_code_cache(
```
bartlomieju, line 90: Is there a reason this method is sync? Given that `prepare_module_load` and `load` are async maybe it would make sense to make it async as well? That way we wouldn't have to make blocking FS calls to store the code cache.

core/modules/mod.rs
```diff
@@ -353,6 +359,7 @@ impl ModuleType {
 pub struct ModuleSource {
   pub code: ModuleSourceCode,
   pub module_type: ModuleType,
+  pub code_cache: Option<Cow<'static, [u8]>>,
```
bartlomieju, line 362: Is using `'static` here required by V8 API? Seems very restrictive 😬

core/runtime/jsruntime.rs
```diff
@@ -526,6 +531,16 @@ pub struct RuntimeOptions {
   /// A callback that allows to evaluate a custom type of a module - eg.
   /// embedders might implement loading WASM or test modules.
   pub custom_module_evaluation_cb: Option<CustomModuleEvaluationCb>,
+
+  // Controls whether V8 code cache is enabled. Code cache can be applied
+  // to ES modules (loaded through ModuleLoader) and to scripts evaluated
```
bartlomieju, line 536:
```suggestion
  // to ES modules (loaded through `ModuleLoader`) and to scripts evaluated
```

core/runtime/jsruntime.rs
```diff
@@ -526,6 +531,16 @@ pub struct RuntimeOptions {
...
+  // through evalContext.
```
bartlomieju, line 537:
```suggestion
  // through `Deno.core.evalContext`.
```

core/modules/loaders.rs
igorzi: Good point, will make it async.

How about the callback inside `op_eval_context`? `op_eval_context` is current sync; any issues with making it async?

core/modules/mod.rs
igorzi: No, it's not required by V8 API. Will make it less restrictive.

core/modules/loaders.rs
igorzi: Actually, never mind. Given that `require` is sync I'd assume that we want to keep `op_eval_context` sync.

core/modules/mod.rs
igorzi: After thinking about this some more, I'm not sure this would be useful. If the embedder loads the cache from some external source (e.g. file on disk or a database) - they will just load the cache and hand over the owned `Vec<u8>` when creating `ModuleSource`. There is no reason for the implementor to keep the cache around after creating `ModuleSource`. It would just complicate the cache lifetime management since `ModuleSource` has to outlive the cache. It would make sense for `ModuleSource` to just own the cache with `Vec<u8>`.

Providing code cache as `&[u8]` would be useful if the implementor had the cache as part of their binary (in which case the lifetime is `'static`).

Do you have some specific use case in mind?

core/modules/loaders.rs
bartlomieju: Yeah, I think we should keep it sync - I think we can make API similar to Node's vm where you need to provide already loaded cache there.

core/modules/mod.rs
bartlomieju: So my thinking is that if we were to provide it for some user modules (reading form DENO_DIR) we'd have to leak the cache data to transmute it to `'static` lifetime. Using owned vector seems much more approachable and would prevent potential leaks if you spawn workers and provide code cache for them but then the worker is terminated but the memory has already been leaked.

igorzi: With `Cow<'static, [u8]>` we would be able to do both. In the case of user modules that I'm targeting - it will be owned vector read from DENO_DIR. I think the `'static` reference can be used in the future for providing code cache for builtin modules (or maybe for user modules from `deno compile` binary). We could just change it to `Vec<u8>` for now since that's what we need for user modules.

Or are you specifically thinking about the use case where multiple workers are loading the same modules, so we'd want to share the same code cache across all workers?

bartlomieju: I'm taking about the latter, but for the first pass it's fine if we just copy the code cache data. 

So it's fine really for now.

bartlomieju (approved):

core/modules/loaders.rs
igorzi: My current approach is to handle code cache outside of js, by using callbacks specified in `RuntimeOptions` by the embedder (see `core/ops_builtin_v8.rs`). Are you suggesting that we pass code cache objects through `evalContext` in javascript? That's going to make integration much more complex, and will also add more js/native transitions to read/write cache.

bartlomieju: Yes, that was my thinking - do you see immediate need to do it in `evalContext`? I think we should focus on validating the approach for proper ES modules and then we can expand handling of code cache to other places like `require()` and `Deno.core.evalContext`

igorzi: Yes, I'd like `require` to also be covered in this initial pass because it's making a big difference for cold starts for apps that use npm modules (which internally `require` many other npm modules).

bartlomieju: Hmmmm, then I guess there's no way around it. Then maybe we should have both sync and async version of providing code cache data? `require()` is already fully synchronous and block event loop, but it would be good if loading ES modules (either on startup or via dynamic import() wouldn't block it). WDYT?

igorzi: Yes, that's what the latest revision of this PR is doing. For ES modules, there is the async `ModuleLoader:: code_cache_ready` method.

For `require` there are 2 sync callbacks: to get and to save the code cache from/to the embedder.

bartlomieju: Sounds good, ship it!

PR #665, opened 2024-03-21
Allow relative imports in extension JS/TS code

nathanwhit:
Before

```ts
// In `ext/node/polyfills/foo.ts`
import * as errors from "ext:deno_node/internal/errors.ts";
```

Now

```ts
// In `ext/node/polyfills/foo.ts`
import * as errors from "./internal/errors.ts";
```

It's pretty ugly because the `url` crate restricts you from directly modifying "bases" or "hosts" of URLs that aren't valid bases, but this avoids having to reimplement a worse version of `Url::join`


bartlomieju (approved):
Nice work

Issue #521, opened 2024-01-30
Ops are added to "Deno.core.ops" on startup with snapshot


bartlomieju:
I just found out that we are adding JS bindings to ops on each startup - even if we are running with a snapshot that has all the ops already saved.

This can be illustrated like so:
1. Create a runtime for snapshotting
2. All built-in `core` ops are added to `Deno.core.ops`
3. Create a snapshot
4. Create a runtime from a snapshot
5. All built-in `core` ops are added to `Deno.core.ops` again

This effectively renders `ensureFastOps()` useless as ops are always bound to `Deno.core.ops` on startup.

bartlomieju:
I think this is actually invalid. This behavior can be disabled using `skip_op_registration` setting when starting up the `JsRuntime`.

PR #667, opened 2024-03-21
refactor: Remove Deno.core.ensureFastOps

bartlomieju:
Instead embedders should tear off appropriate ops
directly from `Deno.core.ops` object.

This API just added another layer of indirection for no obvious
reason.

Ref https://github.com/denoland/deno_core/issues/521


mmastrac (approved):
LGTM

PR #668, opened 2024-03-21
feat(op2): Allow attaching static metadata to ops in #[op2]

nathanwhit:
Currently, only `sanitizer_details` and `sanitizer_fix` (to replace [this map](https://github.com/denoland/deno/blob/2ce645137a6539411dad8d9fe359bc651b215ba5/cli/tools/test/fmt.rs#L293-L294)), but adding more metadata fields shouldn't require any updates to the macro code.

Now you can write

```rust
#[op2(async)]
#[serde]
#[meta(
  sanitizer_details = "receive a message from a BroadcastChannel",
  sanitizer_fix = "closing the BroadcastChannel"
)]
pub async fn op_broadcast_recv(..) {
 ..
}
```

This expands to setting values in the `OpMetadata` struct, which is attached to `OpDecl` and publically accessible via the `metadata` method on `OpCtx`

Closes https://github.com/denoland/deno_core/issues/467

### DIFF
--- a/core/extensions.rs
+++ b/core/extensions.rs
@@ -1,6 +1,7 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 use crate::modules::IntoModuleCodeString;
 use crate::modules::ModuleCodeString;
+use crate::ops::OpMetadata;
 use crate::runtime::bindings;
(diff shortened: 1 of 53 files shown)


mmastrac (approved):
LGTM

PR #673, opened 2024-03-22
feat(op2): support cppgc in async ops

littledivy:
`#[cppgc]` return values were added in https://github.com/denoland/deno_core/commit/a736d868d091406ee20c3fda44bde56def2e73f3 which made 
creation of cppgc backed objects possible without direct access to V8 APIs. 

This patch builds on top of that and adds `#[cppgc]` support for async ops.

### DIFF
--- a/core/runtime/ops.rs
+++ b/core/runtime/ops.rs
@@ -597,9 +597,7 @@ mod tests {
       op_buffer_any_length,
       op_arraybuffer_slice,
       op_test_get_cppgc_resource,
-      op_test_get_cppgc_resource_async,
       op_test_make_cppgc_resource,
-      op_test_make_cppgc_resource_async,
       op_external_make,
       op_external_process,
       op_external_make_ptr,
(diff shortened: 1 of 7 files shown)


core/runtime/ops.rs
```diff
@@ -1865,6 +1867,20 @@ mod tests {
...
+    TestResource { value: 42 }
+  }
+
+  #[op2(async)]
+  #[smi]
+  pub async fn op_test_get_cppgc_resource_async(
+    #[cppgc] resource: &TestResource,
+  ) -> u32 {
+    resource.value
+  }
```
bartlomieju, line 1882: Can you do these in `checkin` instead. We don't want them added for all programs

core/runtime/ops.rs
```diff
@@ -1877,6 +1893,20 @@ mod tests {
     Ok(())
   }
 
+  #[tokio::test]
+  pub async fn test_op_cppgc_object_async(
+  ) -> Result<(), Box<dyn std::error::Error>> {
```
bartlomieju, line 1898: Ditto

core/runtime/ops.rs
littledivy: this is in `mod test` i.e. test runtime

bartlomieju (approved):

core/runtime/ops.rs
mmastrac: I think it's probably best to do as a checkin test tbh.

PR #676, opened 2024-03-22
fix(op2): Mark "fast" op function pointers `extern "C"`

nathanwhit:
before `#[op2(fast)]` would generate

```rust
fn v8_fn_ptr_fast(..) {
  //
}
```

now

```rust
extern "C" fn v8_fn_ptr_fast(..) {
  //
}
```


aapoalas (approved):
Ooof. Nice catch.

littledivy (approved):
LGTM

PR #677, opened 2024-03-22
refactor: Remove usages of serde_v8::Value

bartlomieju:
Removes usages of `serde_v8::Value` in favor of manually handling `v8::Value` in:
- `op_leak_tracing_get_all`
- `op_get_promise_details`
- `op_get_proxy_details`

PR #678, opened 2024-03-24
refactor: remove more usages of serde_v8::Value

bartlomieju:
Follow up to #677


core/async_cancel.rs
```diff
@@ -795,7 +795,7 @@ mod tests {
       let cancel_handle = Rc::new(CancelHandle::new());
       let result = loop {
         select! {
-          r = TcpStream::connect("1.2.3.4:12345")
+          r = TcpStream::connect("127.0.0.1:12345")
```
bartlomieju, line 798: This test didn't work on a plane because it was hitting a remote server

ops/op2/generator_state.rs
```diff
@@ -4,8 +4,6 @@ use proc_macro2::Ident;
 pub struct GeneratorState {
   /// Identifiers for each of the arguments of the original function
   pub args: Vec<Ident>,
-  /// The new identifier for the original function's contents.
-  pub call: Ident,
```
bartlomieju, line 8: Not exactly sure why, but now clippy complains that this field is never used. @mmastrac can you take a look and confirm?

bartlomieju: Done in https://github.com/denoland/deno_core/pull/680

littledivy (approved):
LGTM

PR #679, opened 2024-03-24
refactor: don't require global WebAssembly namespace on startup

bartlomieju:


littledivy (approved):
LGTM, maybe add a test for --jitless?

mmastrac (approved):
LGTM

bartlomieju:
> LGTM, maybe add a test for --jitless?

It's quite hard to do as we don't currently have a way to test binaries.

PR #681, opened 2024-03-25
refactor: Make serde_v8::Value private

bartlomieju:
This commit makes serde_v8::Value variant private - meaning that users
can no longer accept `serde_v8::Value` in the ops as arguments. Instead
`v8::Value` should be used directly. It is also disallowed to use `serde_v8::Value`
in the structure arguments - instead the `v8::Value` should be passed as an
explicit argument. It lowers flexibility, but makes it so that there's only a 
single way of passing V8 values.

Unfortunately the implementation itself can't be removed because it's used internally
to provide serde for `#[serde]` arguments.


littledivy (approved):
LGTM

PR #682, opened 2024-03-30
feat: op2 codegen for object methods

littledivy:
Introducing op2++ (a.k.a `op2` for object methods). 

op2++ is to op2 as C++ is to C.

This patch introduces a small extension on top of the existing op2 codegen infrastructure to enable writing cppgc object methods in Rust:

```rust
pub struct City {
  name: String,
}

impl City {
  #[op2(method(City))]
  #[string]
  fn get_name(&self) -> String {
    self.name.clone()
  }

  #[op2(fast, method(City))]
  fn print_name(&self) {
    println!("{}", self.name);
  }
}

#[op2]
pub fn op_city_new<'a>(
  ctx: &OpCtx,
  #[string] name: String,
) -> v8::Local<'a, v8::Object> {
  let obj = deno_core::cppgc::make_cppgc_object(scope, City { name });
  City::register_get_name(scope, ctx, obj);
  City::register_print_name(scope, ctx, obj);
  obj
}
```

```ts
import { op_city_new } from "ext:core/ops";

const city = op_city_new("Pune");
city.print_name();

const name = city.get_name();
```

Notable changes:
- Added `&OpCtx` as argument, required for registration of methods.
- New attribute `method(Type)` where Type is the `type` of the reciever `self`. Due to the independent nature of op2 codegen we need to know the concrete type. 
- Generate registration methods `Type::register_op` that can be called to register the method onto a Cppgc object.

Future work could make these constructors transparent by processing `impl`s.

### DIFF
--- a/core/runtime/bindings.rs
+++ b/core/runtime/bindings.rs
@@ -323,6 +323,17 @@ pub(crate) fn initialize_deno_core_ops_bindings<'s>(
   }
 }
 
+pub fn register_op_method(
+  scope: &mut v8::HandleScope,
+  op_ctx: OpCtx,
+  obj: v8::Local<v8::Object>,
+) {
+  let key = op_ctx.decl.name_fast.v8_string(scope);
+  let op_fn = op_ctx_function(scope, Box::leak(Box::new(op_ctx)));
+
+  obj.set(scope, key.into(), op_fn.into()).unwrap();
+}
+
 fn op_ctx_function<'s>(
   scope: &mut v8::HandleScope<'s>,
   op_ctx: &OpCtx,
--- a/core/runtime/jsruntime.rs
+++ b/core/runtime/jsruntime.rs
@@ -183,8 +183,6 @@ impl InnerIsolateState {
     unsafe {
       ManuallyDrop::take(&mut self.main_realm).0.destroy();
     }
-
-    debug_assert_eq!(Rc::strong_count(&self.state), 1);
   }
 
   pub fn prepare_for_snapshot(mut self) -> v8::OwnedIsolate {
--- a/ops/op2/dispatch_fast.rs
+++ b/ops/op2/dispatch_fast.rs
@@ -641,6 +641,10 @@ fn map_v8_fastcall_arg_to_arg(
       *needs_js_runtime_state = true;
       quote!(let #arg_ident = &#js_runtime_state;)
     }
+    Arg::Ref(RefType::Ref, Special::OpCtx) => {
+      *needs_opctx = true;
+      quote!(let #arg_ident = #opctx;)
+    }
     Arg::State(RefType::Ref, state) => {
       *needs_opctx = true;
       let state =
@@ -800,6 +804,7 @@ fn map_arg_to_v8_fastcall_type(
     | Arg::Ref(_, Special::OpState)
     | Arg::Rc(Special::JsRuntimeState)
     | Arg::Ref(RefType::Ref, Special::JsRuntimeState)
+    | Arg::Ref(RefType::Ref, Special::OpCtx)
     | Arg::State(..)
     | Arg::Special(Special::Isolate)
     | Arg::OptionState(..)
--- a/ops/op2/dispatch_fast.rs
+++ b/ops/op2/dispatch_fast.rs
@@ -439,6 +439,21 @@ pub(crate) fn generate_dispatch_fast(
     quote!()
   };
 
+  let with_self = if generator_state.needs_self {
+    gs_quote!(generator_state(self_ty) => {
+      let self_: &#self_ty = deno_core::cppgc::try_unwrap_cppgc_object(this.into()).unwrap();
+    })
+  } else {
+    quote!()
+  };
+
+  let name = &generator_state.name;
+  let call = if generator_state.needs_self {
+    quote!(self_. #name)
+  } else {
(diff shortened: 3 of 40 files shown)


ops/op2/dispatch_fast.rs
```diff
@@ -503,7 +518,7 @@ pub(crate) fn generate_dispatch_fast(
 
     #[allow(clippy::too_many_arguments)]
     extern "C" fn #fast_function(
-      _: deno_core::v8::Local<deno_core::v8::Object>,
```
mmastrac, line 506: Let's split this one line change into its own PR + the test changes to reduce the PR size.

mmastrac (commented):
Let's do a small change to break the `this` change out and then work through this.

ops/op2/dispatch_fast.rs
littledivy: it will be a warning (unused argument) if landed seperately

mmastrac: No worries -- just add `allow_unused` for it. We can remove that at a later date.

core/runtime/bindings.rs
```diff
@@ -323,6 +323,17 @@ pub(crate) fn initialize_deno_core_ops_bindings<'s>(
   }
 }
 
+pub fn register_op_method(
```
mmastrac, line 326: We should try to create these once when the JsRuntime is created and then look them up.

core/runtime/jsruntime.rs
```diff
@@ -183,8 +183,6 @@ impl InnerIsolateState {
     unsafe {
       ManuallyDrop::take(&mut self.main_realm).0.destroy();
     }
-
-    debug_assert_eq!(Rc::strong_count(&self.state), 1);
```
mmastrac, line 187: This is definitely a smell here -- we're leaking OpCtxs.

ops/op2/dispatch_fast.rs
```diff
@@ -439,6 +439,21 @@ pub(crate) fn generate_dispatch_fast(
     quote!()
   };
 
+  let with_self = if generator_state.needs_self {
+    gs_quote!(generator_state(self_ty) => {
+      let self_: &#self_ty = deno_core::cppgc::try_unwrap_cppgc_object(this.into()).unwrap();
```
mmastrac, line 444: 👍  awesome

ops/op2/mod.rs
```diff
@@ -281,11 +303,37 @@ fn generate_op2(
...
-      #[inline(always)]
-      #(#attrs)*
-      #op_fn
+      #op_fn_impl
     }
-  })
+  };
+
+  if *needs_self {
+    let register = format_ident!("register_{name}");
```
mmastrac, line 311: This shouldn't be necessary -- this code is effectively the same as if it were on the `OpDecl` itself.

ops/op2/mod.rs
```diff
@@ -122,13 +124,14 @@ fn generate_op2(
     zip(signature.args.iter(), &func.sig.inputs).collect::<Vec<_>>();
 
   let mut args = vec![];
-  let mut needs_args = false;
+  let mut needs_args = config.method.is_some();
```
mmastrac, line 127: Aren't we using the receiver for self? This shouldn't change, right?

ops/op2/signature.rs
```diff
@@ -1345,6 +1349,7 @@ fn parse_type_path(
       }
       ( OpState ) => Ok(CBare(TSpecial(Special::OpState))),
       ( JsRuntimeState ) => Ok(CBare(TSpecial(Special::JsRuntimeState))),
+      ( OpCtx ) => Ok(CBare(TSpecial(Special::OpCtx))),
```
mmastrac, line 1352: I don't think any of this OpCtx stuff will be necessary in the end

mmastrac:
Ok, looked through this PR a few times and I think I know the right plan here. Ignore my previous comment about a `this` param change being separate.Let's do it in this order:

1. Split out just enough of this PR to support `&self` params in `op2`. Ignore the registration part entirely and just manage the cppgc unwrapping functionality. For methods that don't use a receiver, ensure that you keep that parameter named `_` to avoid a large PR. Don't generate any registration methods at all, and let's add tests that bare, `&self`-receiver methods correctly work w/cppgc objects. We should not require any sort of special annotation for these -- just specifying a `self` receiver should be enough to detect that this is an impl method, and it will fail to compile if there's no surrounding impl.
2. Add a registry of op2 methods to JsRuntime. Track the generated function for each extension op2 `DECL` in a map, and ensure that it survives snapshotting. If extension middleware modifies an op, we should be able to retrieve the final op function, even if it's been disabled or nop'd.
3. Create a macro to attach a set of instance methods to an impl. Maybe something like `members!(get_name, print_name, async_method)` . This will generate a slice of static OpDecls that can be used for auto-registration.
4. Add a `&[&'static OpDecl]` parameter to `make_cppgc_object` that defines the members to attach to the created object. In `make_cppgc_object` you should be able to access that registry of op2 methods, retrieve `v8::Local<v8::Function>` for each of those decls, and then perform the set operation. At a later date, we can pre-generate `v8::ObjectTemplate`s for each cppgc object to make this cheaper.

gearonix:
Awesome

PR #685, opened 2024-04-06
Use `const fn` for op declaration

littledivy:
Switch op declaration from a `impl #name` to a `const fn #name`. 
This form for declaration can be used inside `impl` blocks.

```rust
struct Foo {}

impl Foo {
    #[op2(fast)]
    pub fn bar(#[state] state: &mut OpState) {}
}

// const DECL: OpDecl = Foo::bar();
```

Based on Matt's suggestion in https://github.com/denoland/deno_core/pull/682

[Playground link](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=2d7152d548b4a466779c401551b21b05)

This patch wraps `const` around the existing
`impl` codegen to unblock work on op2++

### DIFF
--- a/ops/op2/mod.rs
+++ b/ops/op2/mod.rs
@@ -244,46 +244,51 @@ fn generate_op2(
 
   Ok(quote! {
     #[allow(non_camel_case_types)]
-    #(#attrs)*
-    #vis struct #name <#(#generic),*> {
-      // We need to mark these type parameters as used, so we use a PhantomData
-      _unconstructable: ::std::marker::PhantomData<(#(#generic),*)>
-    }
-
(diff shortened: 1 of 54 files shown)


ops/op2/mod.rs
```diff
@@ -244,46 +244,51 @@ fn generate_op2(
...
-      #slow_fn
+      impl <#(#generic : #bound),*> #name <#(#generic),*> {
+        pub const fn name() -> &'static str {
+          stringify!(#name)
+        }
 
-      #[inline(always)]
-      #(#attrs)*
-      #op_fn
+       #[deprecated(note = "Use the const op::DECL instead")]
```
mmastrac, line 278: I'd be OK to remove this as part of this work

mmastrac (approved):
LGTM, optional if you'd like to remove the deprecated stuff

PR #686, opened 2024-04-07
op2++: codegen for `&self` parameter

littledivy:
Refactor out codegen for `&self` from #682 

```rust
struct State {
  name: &'static str,
}

impl State {
  #[op2(fast, method(State)]
  fn print(&self) {
    println!("{}", self.name);
  }
}

const STATE_DECL: [OpDecl; 1] = [
  State::print()
];
```

### DIFF
--- a/ops/op2/mod.rs
+++ b/ops/op2/mod.rs
@@ -122,13 +122,14 @@ fn generate_op2(
     zip(signature.args.iter(), &func.sig.inputs).collect::<Vec<_>>();
 
   let mut args = vec![];
-  let mut needs_args = false;
+  let mut needs_args = config.method.is_some();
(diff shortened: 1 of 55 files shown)


ops/op2/mod.rs
```diff
@@ -122,13 +122,14 @@ fn generate_op2(
     zip(signature.args.iter(), &func.sig.inputs).collect::<Vec<_>>();
 
   let mut args = vec![];
-  let mut needs_args = false;
+  let mut needs_args = config.method.is_some();
```
littledivy, line 125: Function callback arguments are needed for: `args.this()`

mmastrac (approved):
LGTM!

PR #692, opened 2024-04-09
fix(core): miri tweaks

mmastrac:
- Fix some sketchiness in FastString that we were waiting on rusty_v8 to fix.
 - Run more tests under miri, comments for those that can't

### DIFF
--- a/.github/workflows/ci-coverage/action.yml
+++ b/.github/workflows/ci-coverage/action.yml
@@ -12,4 +12,4 @@ runs:
       uses: codecov/codecov-action@v3
       with:
         files: lcov.info
-        fail_ci_if_error: true
+        fail_ci_if_error: false
--- a/core/fast_string.rs
+++ b/core/fast_string.rs
@@ -29,39 +29,19 @@ impl FastStaticString {
     FastStaticString { s }
   }
 
-  pub const fn as_str(&self) -> &'static str {
-    // SAFETY: We know this was constructed from a valid one-byte string.
-    unsafe { std::str::from_utf8_unchecked(self.as_bytes()) }
+  pub fn as_str(&self) -> &'static str {
+    self.s.as_ref()
   }
 
-  pub const fn as_bytes(&self) -> &'static [u8] {
-    unsafe {
-      let ptrs: [usize; 3] = std::ptr::read(self.s as *const _ as *const _);
-      std::slice::from_raw_parts(ptrs[1] as *const u8, ptrs[2])
-    }
+  pub fn as_bytes(&self) -> &'static [u8] {
+    self.s.as_ref()
   }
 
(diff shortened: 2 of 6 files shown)


.github/workflows/ci-coverage/action.yml
```diff
@@ -12,4 +12,4 @@ runs:
       uses: codecov/codecov-action@v3
       with:
         files: lcov.info
-        fail_ci_if_error: true
```
mmastrac, line 15: Codecov is not always succeeding :/

core/fast_string.rs
```diff
@@ -471,9 +451,6 @@ macro_rules! __op_name_fast {
 mod tests {
   use super::*;
 
-  // TODO(mmastrac): we need to upstream as_str() on OneByteConst to rusty_v8 to get
-  // rid of the hacks that break miri.
-  #[cfg(not(miri))]
```
bartlomieju, line 476: Have we actually upstreamed this change and just waiting on `rusty_v8` release?

core/module_specifier.rs
```diff
@@ -335,7 +335,11 @@ mod tests {
     ];
 
     // The local path tests assume that the cwd is the deno repo root.
-    let cwd = current_dir().unwrap();
+    let cwd = if cfg!(miri) {
+      PathBuf::from("/miri")
+    } else {
+      current_dir().unwrap()
+    };
```
bartlomieju, line 342: Add a comment here?

core/fast_string.rs
mmastrac: It's been out for a bit -- I just forgot about this cleanup.

core/module_specifier.rs
```diff
@@ -335,7 +335,11 @@ mod tests {
     ];
 
     // The local path tests assume that the cwd is the deno repo root.
```
mmastrac, line 337:
```suggestion
    // The local path tests assume that the cwd is the deno repo root. Note
    // that we can't use `cwd` in miri tests, so we just use `/miri` instead.
```

core/fast_string.rs
```diff
@@ -29,39 +29,19 @@ impl FastStaticString {
...
-    // SAFETY: Workaround compiler error in create_external_onebyte_const for long ascii strings
-    unsafe {
-      debug_assert!(s.is_ascii());
-      let c = v8::String::create_external_onebyte_const(&[]);
-      let mut ptrs: OneByteConst = std::mem::transmute(c);
-      ptrs.cached_data = std::mem::transmute(s.as_ptr());
-      ptrs.length = s.len();
-      std::mem::transmute(ptrs)
-    }
+    v8::String::create_external_onebyte_const(s)
```
littledivy, line 44: nice

littledivy (approved):
lgtm

PR #693, opened 2024-04-10
Allow cloning `Rc<ContextState>` into new contexts

littledivy:
Required for  https://github.com/denoland/deno/pull/23305/

Allowing cloning `Rc<ContextState>` into new contexts. 

During cleanup we assert that main context does not have more than one strong count to `JsRuntimeState` but this may not be true when there are more than one contexts that may not have been GC'ed yet. Hence, JsRuntimeState is now stored as a pointer to OpCtx as it is guarnateed to live longer than ops.


core/ops.rs
```diff
@@ -245,7 +245,7 @@ impl OpCtx {
 
   /// Get the [`JsRuntimeState`] for this op.
   pub(crate) fn runtime_state(&self) -> &JsRuntimeState {
-    &self.runtime_state
+    unsafe { &*self.runtime_state }
```
mmastrac, line 248: SAFETY comment

mmastrac (commented):
LGTM w/the comment. Does node bubble up unhandled rejections in a vm.Context to the main context?

littledivy:
> Does node bubble up unhandled rejections in a vm.Context to the main context?

Nope

mmastrac (approved):

PR #696, opened 2024-04-12
Fix extension! parameters with `const fn` decl

littledivy:
Added a test which compiles


testing/checkin/runner/mod.rs
```diff
@@ -105,7 +111,7 @@ fn create_runtime(
 
   let snapshot = runtime_for_snapshot.snapshot();
   let snapshot = Box::leak(snapshot);
-  let extensions = vec![checkin_runtime::init_ops()];
+  let extensions = vec![checkin_runtime::init_ops::<()>()];
```
bartlomieju, line 114: Does that mean that all extensions will need to specify `::<()>`?

littledivy: No, this is just a test for generics, we didn't have any

PR #709, opened 2024-04-22
fix(core): don't lose waker for tasks

mmastrac:
Ref https://github.com/denoland/deno/issues/21686

If we had additional tasks generated while servicing these tasks we'd eventually fail to wake the waker (which no longer exists in the task spawner).

### DIFF
--- a/core/runtime/jsruntime.rs
+++ b/core/runtime/jsruntime.rs
@@ -2278,12 +2278,24 @@ impl JsRuntime {
 
     // Poll any pending task spawner tasks. Note that we need to poll separately because otherwise
     // Rust will extend the lifetime of the borrow longer than we expect.
-    let tasks = context_state.task_spawner_factory.poll_inner(cx);
-    if let Poll::Ready(tasks) = tasks {
+    let mut retries = 3;
+    while let Poll::Ready(tasks) =
+      context_state.task_spawner_factory.poll_inner(cx)
+    {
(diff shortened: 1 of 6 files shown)


core/runtime/jsruntime.rs
```diff
@@ -2278,13 +2278,25 @@ impl JsRuntime {
...
+    {
       // TODO(mmastrac): we are using this flag
       dispatched_ops = true;
       for task in tasks {
         task(scope);
       }
+      // We may need to perform a microtask checkpoint here
+      scope.perform_microtask_checkpoint();
+
+      // We don't want tasks that spawn other tasks the starve the event loop, so break
```
aapoalas, line 2293: **nitpick:** "the starve"

testing/unit/task_test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+test(async function testTaskSubmit3() {
+  for (let i = 0; i < 3; i++) {
+    const { promise, resolve } = Promise.withResolvers();
+    op_task_submit(() => {
+      resolve(undefined);
+    });
+    await promise;
+  }
+});
```
bartlomieju, line 32: Does it still work fine if the loop has 100 or 1000 iterations?

mmastrac: It does -- I added a 100 iteration test "just in case"

bartlomieju (approved):

PR #727, opened 2024-05-02
feat: Add `ToV8`/`FromV8` impls for bool and numeric types

nathanwhit:
Adds two wrapper structs which implement two different modes of serializing/deserializing numeric types:

```rust
pub struct Smi<T: SmallInt>(pub T);

// serializes like #[smi] (but without fast call support)
#[op2]
#[to_v8]
fn op_returns_smi() -> Smi<u32> {
  Smi(1)
}

pub struct Number<T: Numeric>(pub T);

// serializes like #[number]
#[op2]
#[to_v8]
fn op_returns_number() -> Number<u64> {
  Number(2)
}
```

Also adds an impl for `bool` itself, which is fairly uncontroversial I think.

TODO:
- [x] add tests

### DIFF
--- a/core/error.rs
+++ b/core/error.rs
@@ -96,6 +96,33 @@ pub fn get_custom_error_class(error: &Error) -> Option<&'static str> {
   error.downcast_ref::<CustomError>().map(|e| e.class)
 }
 
+/// A wrapper around `anyhow::Error` that implements `std::error::Error`
+#[repr(transparent)]
+pub struct StdAnyError(pub Error);
+impl std::fmt::Debug for StdAnyError {
+  fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    write!(f, "{:?}", self.0)
+  }
+}
+
+impl std::fmt::Display for StdAnyError {
+  fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    write!(f, "{}", self.0)
+  }
+}
+
+impl std::error::Error for StdAnyError {
+  fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+    self.0.source()
+  }
+}
+
+impl From<Error> for StdAnyError {
+  fn from(err: Error) -> Self {
+    Self(err)
+  }
+}
+
 pub fn to_v8_error<'a>(
   scope: &mut v8::HandleScope<'a>,
   get_class: GetErrorClassFn,
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -2,6 +2,7 @@
 pub mod arena;
 mod async_cancel;
 mod async_cell;
+pub mod convert;
 pub mod cppgc;
 pub mod error;
 mod error_codes;
@@ -26,7 +27,6 @@ mod path;
 mod runtime;
 mod source_map;
 mod tasks;
-pub mod to_from_v8;
 mod web_timeout;
 
 // Re-exports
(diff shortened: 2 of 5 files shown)


core/error.rs
```diff
@@ -96,6 +96,33 @@ pub fn get_custom_error_class(error: &Error) -> Option<&'static str> {
   error.downcast_ref::<CustomError>().map(|e| e.class)
 }
 
+/// A wrapper around `anyhow::Error` that implements `std::error::Error`
+#[repr(transparent)]
+pub struct StdAnyError(pub Error);
```
nathanwhit, line 101: Annoyingly, `anyhow::Error` doesn't impl the `Error` trait. Not sure if we want to go down the error enum route (like `serde_v8::Error`) or continue using `anyhow::Error` as our primary error type.

core/lib.rs
```diff
@@ -26,7 +26,7 @@ mod path;
 mod runtime;
 mod source_map;
 mod tasks;
-mod to_from_v8;
+pub mod to_from_v8;
```
nathanwhit, line 29: Make this public (and maybe rename it), or re-export the markers?

core/to_from_v8.rs
```diff
@@ -17,3 +20,152 @@ pub trait FromV8<'a>: Sized {
     value: v8::Local<'a, v8::Value>,
   ) -> Result<Self, Self::Error>;
 }
+
+// impls
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
mmastrac, line 26: We're going to need to ensure the docs for this are decent. We can fill in more details later:

```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
/// Marks a numeric type as being serialized as a v8 `smi` in a `v8::Integer`.
```

core/to_from_v8.rs
```diff
@@ -17,3 +20,152 @@ pub trait FromV8<'a>: Sized {
...
+    value: v8::Local<'a, v8::Value>,
+  ) -> Result<Self, Self::Error> {
+    let v = crate::runtime::ops::to_i32_option(&value).ok_or_else(|| {
+      crate::error::type_error(format!("Expected {}", T::NAME))
+    })?;
+    Ok(Smi(T::from_i32(v)))
+  }
+}
+
+pub struct Number<T: Numeric>(pub T);
```
mmastrac, line 87: Mark as transparent/w the same derives, plus a doc.

core/to_from_v8.rs
```diff
@@ -17,3 +20,152 @@ pub trait FromV8<'a>: Sized {
...
+  #[allow(clippy::wrong_self_convention)]
+  fn as_f64(self) -> f64;
+  fn from_value(value: &v8::Value) -> Option<Self>;
+}
+
+macro_rules! impl_numeric {
+  ($($t:ty : $from: path ),*) => {
+    $(
+      impl Numeric for $t {
+        const NAME: &'static str = stringify!($t);
```
mmastrac, line 100: It's a shame https://doc.rust-lang.org/std/any/fn.type_name.html isn't const :/

core/lib.rs
mmastrac: I think we can make it public. Bikeshedding some names:

`v8_convert`, `convert`, `to_v8`

core/to_from_v8.rs
```diff
@@ -1,5 +1,8 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 
+use crate::{error::StdAnyError, runtime::ops};
+use std::convert::Infallible;
+
 pub trait ToV8<'a> {
```
mmastrac, line 6: While we're here, let's add some basic docs to ToV8 and FromV8.

core/lib.rs
nathanwhit: Went with `convert`, I feel like `deno_core::convert::ToV8` reads nicely

core/to_from_v8.rs
nathanwhit: Added some docs, feel free to review

mmastrac (approved):
LGTM!

PR #729, opened 2024-05-05
fix UB in bytestring::from_v8

robtfm:
reliably on rust 1.78 (and i think maybe occasionally previously) `ByteString::from_v8`'s strategy of `set_len`ing a smallvec seems to be UB and results in bad memory accesses in `op_fetch` (and probably `op_node_http_request` as well).

i don't have a minimal repro but i do have a project with deno embedded which reliably fails on different architectures, just from calling fetch() - the `method` param ends up holding garbage.

not sure if the approach i took is the fastest safe strategy - please let me know if there's a better way. but actually this type is rarely used in v8->deno direction anyway so maybe it doesn't matter much.

### DIFF
--- a/serde_v8/magic/bytestring.rs
+++ b/serde_v8/magic/bytestring.rs
@@ -73,20 +73,16 @@ impl FromV8 for ByteString {
       return Err(Error::ExpectedLatin1);
     }
     let len = v8str.length();
-    let mut buffer = SmallVec::with_capacity(len);
-    #[allow(clippy::uninit_vec)]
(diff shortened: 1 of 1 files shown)


robtfm:
There’s another similar block in u16string which could be changed too if you like. I don’t have an issue with it but it’s the same vec_uninit -> ffi pattern

robtfm:
sorry spoke too soon. i was checking if this was the issue by using a string in op_fetch directly instead, then made the patch and forgot to revert my change to string.

i am still getting heap corruption with this change, so the actual problem must be elsewhere.

PR #735, opened 2024-05-10
chore: bump Rust and cargo update

mmastrac:
We bump Rust to 1.78, but for the tests to pass we need to remove the UB that we've been relying on:

1. Opaque pointers no longer rely on values on the stack (!)
2. We no longer take a `&mut` from a u64 that comes from a `&` (UB)
3. ToV8Buffer gets a RefCell for proper interior mutability

Lots of sketchiness left in serde_v8 but hopefully this fixes the major fire.

serde_v8 is considered deprecated so we're likely not going to touch this code much more than this band-aid.

### DIFF
--- a/serde_v8/ser.rs
+++ b/serde_v8/ser.rs
@@ -260,8 +260,8 @@ impl<'a, 'b, 'c, T: MagicType + ToV8> ser::SerializeStruct
 
   fn end(self) -> JsResult<'a> {
     // SAFETY: transerialization assumptions imply `T` is still alive.
-    let x: &mut T =
-      unsafe { (self.opaque as *mut T).as_mut().unwrap_unchecked() };
+    let x: &T =
+      unsafe { (self.opaque as *const T).as_ref().unwrap_unchecked() };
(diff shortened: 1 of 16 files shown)


lucacasonato (approved):
RSLGTM

serde_v8/ser.rs
```diff
@@ -254,16 +252,20 @@ impl<'a, 'b, 'c, T: MagicType + ToV8> ser::SerializeStruct
...
+    self.opaque = unsafe { std::ptr::read(value as *const _ as *const u64) };
     Ok(())
   }
 
   fn end(self) -> JsResult<'a> {
+    // SERIALIZATION CRIMES
+
     // SAFETY: transerialization assumptions imply `T` is still alive.
-    let x: &mut T = unsafe { opaque_deref_mut(self.opaque) };
+    let x: &T =
```
mmastrac, line 267: This was not only unsafe, but UB -- we turned a ref into a pointer into a u64 into a `&mut`.

nathanwhit (approved):
LGTM, looks a lot better (and is now not clearly unsound)

PR #750, opened 2024-05-22
fix(core): remove assert flag

mmastrac:


core/modules/tests.rs
```diff
@@ -504,35 +504,9 @@ fn test_json_module() {
...
-    let specifier_a = ascii_str!("file:///a.js");
     let specifier_b = ascii_str!("file:///b.js");
-    let mod_a = module_map
-      .new_es_module(
-        scope,
-        true,
-        specifier_a,
-        ascii_str!(
-          r#"
-          import jsonData from './c.json' assert {type: "json"};
```
bartlomieju, line 518: I think you can only replace `assert` with `with` and it will still work. Unless we already have a test that cover import attributes

bartlomieju (approved):

core/modules/tests.rs
mmastrac: Yup. This test covered both styles previously -- removed the legacy syntax one only, leaving the new syntax one.

PR #752, opened 2024-05-23
refactor: pass hash into code cache

dsherret:
This avoids the cli having to re-lookup the hash again.

### DIFF
--- a/core/modules/map.rs
+++ b/core/modules/map.rs
@@ -93,7 +93,7 @@ impl ModEvaluate {
 type CodeCacheReadyCallback =
   Box<dyn Fn(&[u8]) -> Pin<Box<dyn Future<Output = ()>>>>;
 pub(crate) struct CodeCacheInfo {
-  code_cache: Option<Cow<'static, [u8]>>,
+  code_cache_data: Option<Cow<'static, [u8]>>,
   ready_callback: CodeCacheReadyCallback,
 }
 
@@ -139,7 +139,6 @@ pub(crate) struct ModuleMap {
   pending_code_cache_ready: Cell<bool>,
   module_waker: AtomicWaker,
   data: RefCell<ModuleMapData>,
-  enable_code_cache: bool,
 
   /// A counter used to delay our dynamic import deadlock detection by one spin
   /// of the event loop.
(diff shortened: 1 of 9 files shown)


core/modules/tests.rs
```diff
@@ -275,7 +288,8 @@ impl ModuleLoader for MockLoader {
 
   fn code_cache_ready(
     &self,
-    module_specifier: &ModuleSpecifier,
+    module_specifier: ModuleSpecifier,
```
dsherret, line 291: Changed this to an owned value because deno_core has it available without cloning and this can help prevent cloning if someone wants to send this to another thread when saving the code cache.

core/modules/tests.rs
```diff
@@ -275,7 +288,8 @@ impl ModuleLoader for MockLoader {
...
+    _hash: u64,
     code_cache: &[u8],
   ) -> Pin<Box<dyn Future<Output = ()>>> {
```
dsherret, line 294: We should look into doing something similar for the script callbacks. I'm just not sure how to convert a `Local<String>` to a `&str`... maybe deno_core should just create the hash.

bartlomieju: You can convert it like so: `&x.to_rust_string_lossy(scope)`

bartlomieju (approved):

core/modules/map.rs
```diff
@@ -346,24 +343,25 @@ impl ModuleMap {
           ModuleSource::get_string_source(module_url_found.as_str(), code)
             .map_err(ModuleError::Other)?;
 
-        let (code_cache_info, module_url_found) = if self.enable_code_cache {
```
igorzi, line 349: This changed to no longer use `enable_code_cache` when deciding if to fire the `code_cache_ready` callback. I think `enable_code_cache` may still be needed, because on the first run there will be no `code_cache` yet, and we do want the callback to fire.

dsherret: As discussed in DMs, this is now enabled based on whether someone provides a `ModuleSourceCodeCache` or not:

```rs
#[derive(Debug)]
pub struct ModuleSourceCodeCache {
  pub hash: u64,
  pub data: Option<Cow<'static, [u8]>>,
}

// ....
pub struct ModuleSource {
  // ...
  pub code_cache: Option<ModuleSourceCodeCache>,
```

(Probably will rename `ModuleSourceCodeCache` because it's a bad name)

PR #754, opened 2024-05-23
fix: make `__callSiteEvals` private

0f-0b:


ry (commented):
Thanks for the patch. It seems reasonable - have you run any benchmarks? Is there any performance impact?

ref https://github.com/denoland/deno_core/pull/739#pullrequestreview-2074199556

0f-0b:
Looks like it's faster with this change.

```shellsession
$ ./bench_main

running 1 test
test bench ... bench:   2,317,572 ns/iter (+/- 36,373)

test result: ok. 0 passed; 0 failed; 0 ignored; 1 measured

$ ./bench_private_call_sites

running 1 test
test bench ... bench:   2,031,424 ns/iter (+/- 46,885)

test result: ok. 0 passed; 0 failed; 0 ignored; 1 measured

```

Benchmark code:

```rs
use bencher::{benchmark_group, benchmark_main, Bencher};
use deno_core::{ascii_str, v8, JsRuntime, RuntimeOptions};

fn bench(b: &mut Bencher) {
  let mut runtime = JsRuntime::new(RuntimeOptions {
    unsafe_expose_natives_and_gc: true,
    ..Default::default()
  });
  runtime
    .execute_script(
      "",
      r#"
        function bench() {
          for (let i = 0; i < 1000; i++) {
            new Error().stack;
          }
        }

        %PrepareFunctionForOptimization(bench);
        bench();
        %OptimizeFunctionOnNextCall(bench);
        bench();
      "#,
    )
    .unwrap();
  let context = runtime.main_context();
  let scope = &mut runtime.handle_scope();
  let bench = {
    let key = ascii_str!("bench").v8_string(scope).into();
    let value = context.open(scope).global(scope).get(scope, key).unwrap();
    v8::Local::<v8::Function>::try_from(value).unwrap()
  };
  let recv = v8::undefined(scope).into();
  b.iter(|| bench.call(scope, recv, &[]));
}

benchmark_group!(benches, bench);
benchmark_main!(benches);
```

bartlomieju (approved):
LGTM, seems like a much better solution that doesn't expose this funky object.

Issue #757, opened 2024-05-28
decoding percent-encoding source string in `Error.stack` to follow browser behaviour


Hajime-san:
# context

context in https://github.com/denoland/deno/pull/23200#issue-2222469787 the section of `handle runtime behaviour`
and https://github.com/denoland/deno/pull/23200#issuecomment-2134029241

# overview

## spec

It seems that there is no specification for `Error.stack` today, and it is implementation-defined behaviour by browser engines.

- https://github.com/tc39/proposal-error-stacks

## example to reproduce

Here is an example code to show stack trace, and runs it on modern browsers.

- index.html

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <script type="module" src="./🍋.js"></script>
</head>
<body>
</body>
</html>
```

- 🍋.js

```js
throw new Error('🍋')
```

#### Google Chrome(122.0.6261.112)

```log
🍋.js:2 Uncaught Error: 🍋
    at 🍋.js:2:7
(anonymous) @ 🍋.js:2
```

I think that the stack trace implementation is located as follows.

- https://source.chromium.org/chromium/chromium/src/+/refs/tags/122.0.6261.112:v8/src/inspector/v8-stack-trace-impl.cc;l=351-367

#### Firefox(117.0)

```log
Uncaught Error: 🍋
    <anonymous> http://localhost:8080/🍋.js:1
```

#### Safari(17.3.1)

```log
Error: 🍋
module code - 🍋.js:2
```

#### Node.js(20.12.0)

```log
% node 🍋.mjs         
file:///path/to/%F0%9F%8D%8B.mjs:1
throw new Error('🍋')
      ^

Error: 🍋
    at file:///path/to/%F0%9F%8D%8B.mjs:1:7
    at ModuleJob.run (node:internal/modules/esm/module_job:218:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:329:24)
    at async loadESM (node:internal/process/esm_loader:28:7)
    at async handleMainPromise (node:internal/modules/run_main:113:12)
```

## related issues

https://github.com/denoland/deno/issues/18983

PR #759, opened 2024-05-28
feat: decoding percent-encoding source string in `Error.stack`

Hajime-san:
resolves https://github.com/denoland/deno_core/issues/757

## Note

- [ ] when deno_core crate bumped to deno, the test will fail without fixing the output
https://github.com/denoland/deno/blob/2024c974b613f94f31559a1b32e2d747c2083e91/tests/integration/test_tests.rs#L96
- [ ] when deno_core crate bumped to deno, the issue will resolve
https://github.com/denoland/deno/issues/18983


Cargo.toml
```diff
@@ -43,6 +43,7 @@ libc = "0.2.126"
 memoffset = ">=0.9"
 num-bigint = { version = "0.4", features = ["rand"] }
 parking_lot = "0.12.0"
+percent-encoding = "2.3.0"
```
Hajime-san, line 46: I picked `percent-encoding` crate due to that was used in deno main repo.

bartlomieju:
Sorry for a slow turn-around @Hajime-san. This looks good to me, but it would be great if a test case could be added. It can be as simple as a unit test for `format_file_name` function.

core/error.rs
```diff
@@ -692,17 +692,24 @@ pub(crate) fn has_call_site(
...
-  if file_name.len() <= DATA_URL_ABBREV_THRESHOLD {
-    return None;
-  }
   let url = Url::parse(file_name).ok()?;
   if url.scheme() != "data" {
     return None;
   }
+  if file_name.len() <= DATA_URL_ABBREV_THRESHOLD {
+    return Some(file_name.to_string());
+  }
```
Hajime-san, line 712: This may slightly degrade performance with checking url scheme at first, but I think it's generally fine and improves readability.

bartlomieju: Is this necessary? If not I would suggest to revert this, parsing URL for every stack frame will have measureable impact

Hajime-san: My understanding of the role of the `abbrev_file_name` function is returning `Some` when the argument of the `file_name` string  starts with `data:...` due to avoid decoding Data URLs in the `format_file_name` function.
So, I think that it can rewrite as follows.

```rust
fn abbrev_file_name(file_name: &str) -> Option<String> {
  if !file_name.starts_with("data:") {
    return None;
  }
  if file_name.len() <= DATA_URL_ABBREV_THRESHOLD {
    return Some(file_name.to_string());
  }
  let url = Url::parse(file_name).ok()?;
 ...
```

bartlomieju: Yeah, that sounds good!

Hajime-san: I've fix it as above.
Thank you for point out the performance!

bartlomieju (approved):
LGTM, thanks @Hajime-san!

PR #761, opened 2024-06-03
perf: Batch calls to `ModuleLoader::prepare_load` for dynamic imports

nathanwhit:
Collects dynamic imports between event loop ticks and then calls `prepare_load` with the collected imports. This reduces the number of calls to `prepare_load`.
Performance impact on deno is noticeable but relatively modest, around 6-7%:

```
❯ hyperfine --warmup 10 "deno run -A repro.ts" "devdeno run -A repro.ts"
Benchmark 1: deno run -A repro.ts
  Time (mean ± σ):     211.8 ms ±   2.0 ms    [User: 212.7 ms, System: 85.8 ms]
  Range (min … max):   209.3 ms … 217.4 ms    14 runs

Benchmark 2: devdeno run -A repro.ts
  Time (mean ± σ):     197.3 ms ±   1.2 ms    [User: 189.9 ms, System: 84.5 ms]
  Range (min … max):   195.9 ms … 200.7 ms    14 runs

Summary
  devdeno run -A repro.ts ran
    1.07 ± 0.01 times faster than deno run -A repro.ts

❯ hyperfine --warmup 10 "deno run -A repro2.ts" "devdeno run -A repro2.ts"
Benchmark 1: deno run -A repro2.ts
  Time (mean ± σ):     146.3 ms ±   1.0 ms    [User: 180.9 ms, System: 51.6 ms]
  Range (min … max):   144.5 ms … 148.0 ms    20 runs

Benchmark 2: devdeno run -A repro2.ts
  Time (mean ± σ):     137.8 ms ±   0.8 ms    [User: 160.7 ms, System: 51.7 ms]
  Range (min … max):   136.5 ms … 139.9 ms    21 runs

Summary
  devdeno run -A repro2.ts ran
    1.06 ± 0.01 times faster than deno run -A repro2.ts
```

My main concern here is the loss of granularity in errors that occur during `prepare_load`. Since we now are calling `prepare_load` with multiple specifiers, if the loader returns an error we can't really tell which specifier the error applies to. That causes a behavior change.

### DIFF
--- a/core/modules/map.rs
+++ b/core/modules/map.rs
@@ -27,6 +27,7 @@ use crate::runtime::JsRealm;
 use crate::runtime::SnapshotLoadDataStore;
 use crate::runtime::SnapshotStoreDataStore;
 use crate::FastStaticString;
+use crate::FastString;
 use crate::JsRuntime;
 use crate::ModuleLoadResponse;
 use crate::ModuleSource;
@@ -41,6 +42,8 @@ use futures::task::noop_waker_ref;
 use futures::task::AtomicWaker;
 use futures::Future;
 use futures::StreamExt;
+use smallvec::smallvec;
+use smallvec::SmallVec;
 use v8::Function;
 use v8::PromiseState;
 
@@ -61,7 +64,7 @@ use super::LazyEsmModuleLoader;
 use super::RequestedModuleType;
 
 type PrepareLoadFuture =
-  dyn Future<Output = (ModuleLoadId, Result<RecursiveModuleLoad, Error>)>;
+  dyn Future<Output = (SmallVec<[RecursiveModuleLoad; 1]>, Result<(), Error>)>;
 
 type CodeCacheReadyFuture = dyn Future<Output = ()>;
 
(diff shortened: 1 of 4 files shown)


core/modules/map.rs
```diff
@@ -1405,27 +1487,34 @@ impl ModuleMap {
...
+                .borrow_mut()
+                .push(load.into_future());
+            }
             self.pending_dynamic_imports_pending.set(true);
           }
           Err(err) => {
             let exception = to_v8_type_error(scope, err);
-            self.dynamic_import_reject(scope, dyn_import_id, exception);
+            for load in dyn_imports {
+              self.dynamic_import_reject(scope, load.id, exception.clone());
```
nathanwhit, line 1516: This is the part I'm most concerned about, since we just reject all of the import promises in this batch even though the error might only be caused by a subset.

bartlomieju: Let's change the return type of this hook to `Vec<Result<(), AnyError>>` 👍

PR #762, opened 2024-06-04
fix: wakers of new `ModuleLoadFuture` are never registered when handling concurrent redirects

arnauorriols:


bartlomieju (approved):
Confirmed that the test hangs without this change. LGTM!

PR #767, opened 2024-06-06
Add non-panic constructor for runtimes

rscarson:
Added a new constructor variant that does not panic
Existing constructor can then simply take in the returned error and panic

### DIFF
--- a/core/runtime/jsruntime.rs
+++ b/core/runtime/jsruntime.rs
@@ -605,18 +605,25 @@ impl JsRuntime {
   }
 
   /// Only constructor, configuration is done through `options`.
-  pub fn new(mut options: RuntimeOptions) -> JsRuntime {
-    setup::init_v8(
-      options.v8_platform.take(),
-      cfg!(test),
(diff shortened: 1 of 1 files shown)


core/runtime/jsruntime.rs
```diff
@@ -1884,20 +1891,28 @@ fn create_context<'a>(
...
+    }
+  }
+
+  /// Try to create a new runtime, returning an error if the process fails.
+  pub fn try_new(
+    options: RuntimeOptions,
+  ) -> Result<JsRuntimeForSnapshot, Error> {
     setup::init_v8(
-      options.v8_platform.take(),
+      options.v8_platform.clone(),
```
bartlomieju, line 1909: Why `clone` here? Maybe `.take()` will be enough

rscarson: Woops! Oversight on my part - fixed

bartlomieju (approved):
LGTM, thanks!

PR #769, opened 2024-06-06
fix: Wake event loop when dynamic import promise resolves

nathanwhit:
Fixes a hang that could occur when a module contained a dynamic import that doesn't resolve immediately, and there was a pending op outstanding.

The original reproduction (ref  https://github.com/denoland/deno/issues/24098) effectively boiled down to this:

```ts
// main.ts
Deno.serve(async function (_req: Request): Promise<Response> {
    await import("./repro.ts");
    console.log("imported module");
    return new Response("done");
});

// repro.ts
await new Promise<void>((resolve) => setTimeout(resolve, 0));
console.log("module executed");
```

Upon getting a request, it would hang when trying to import `"repro.ts"`. What was happening here was that `Deno.serve` created an op (`op_http_serve`) that stayed pending indefinitely, so we [assumed](https://github.com/denoland/deno_core/blob/4beb54e862805be6b2c4d8fb2bfc49ea72705799/core/runtime/jsruntime.rs#L1816C6-L1822C9) that we would poll again, but that didn't actually hold. Unlike non-dynamic module evaluation, we weren't waking the event loop for polling when the dynamic import evaluation promise resolved.

The fix here is to attaching callbacks to the dynamic import promise to wake the event loop when it resolves, similar to what we do for normal module evaluation.

### DIFF
--- a/core/modules/map.rs
+++ b/core/modules/map.rs
@@ -1191,9 +1191,9 @@ impl ModuleMap {
     // of module evaluation is a promise.
     //
     // This promise is internal, and not the same one that gets returned to
-    // the user. We add an empty `.catch()` handler so that it does not result
-    // in an exception if it rejects. That will instead happen for the other
-    // promise if not handled by the user.
+    // the user. We add handlers to wake the event loop when the promise resolves
+    // (or rejects). The catch handler also serves to prevent an exception if the internal promise
+    // rejects. That will instead happen for the other if not handled by the user.
     //
     // For more details see:
     // https://github.com/denoland/deno/issues/4908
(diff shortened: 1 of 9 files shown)


core/modules/map.rs
```diff
@@ -1210,11 +1210,37 @@ impl ModuleMap {
...
+      }
+
       let promise = v8::Local::<v8::Promise>::try_from(value)
         .expect("Expected to get promise as module evaluation result");
-      let empty_fn =
-        crate::runtime::bindings::create_empty_fn(tc_scope).unwrap();
-      promise.catch(tc_scope, empty_fn);
+
+      let module_map =
+        v8::External::new(tc_scope, Box::into_raw(Box::new(self.clone())) as _);
```
bartlomieju, line 1231: I wonder if this is really necessary. `ModuleMap` is already available using slot, eg. `JsReadlm::module_map_from`. Do you think we could use it that way?

core/modules/map.rs
```diff
@@ -1210,11 +1210,37 @@ impl ModuleMap {
...
+      ) {
+        let module_map = get_module_map(args.data());
+        module_map.module_waker.wake();
+      }
+
       let promise = v8::Local::<v8::Promise>::try_from(value)
         .expect("Expected to get promise as module evaluation result");
-      let empty_fn =
-        crate::runtime::bindings::create_empty_fn(tc_scope).unwrap();
-      promise.catch(tc_scope, empty_fn);
```
bartlomieju, line 1217: Interesting! We've had this one for the longest time. Would be great if it's not longer needed

bartlomieju (commented):
Overall looks good, I have one question if we can skip juggling `Externals`

bartlomieju (approved):
Great investigation and fix! Let's give it a go, I will include it in https://github.com/denoland/deno/pull/24128.

PR #775, opened 2024-06-07
fix: `ModuleMap` was not dropped if it had pending futures

dsherret:
During `JsRuntime` drop there were situations when `ModuleMap` was actually not
dropped. This could happen if there were pending futures stored in the `ModuleMap`.

These futures contain references to `ModuleMap`. The situation lead to memory being
leaked and subtle bugs in logic because the futures were not dropped either.

`ModuleMap::destroy()` was added that clear all pending futures and
other relevant data from the map during destruction of `JsRealm`.

### DIFF
--- a/core/modules/map.rs
+++ b/core/modules/map.rs
@@ -146,6 +146,14 @@ pub(crate) struct ModuleMap {
 }
 
 impl ModuleMap {
+  /// There is a circular Rc reference between the module map and the futures,
+  /// so when destroying the module map we need to clear the pending futures.
+  pub fn clear_pending_futures(&self) {
+    self.preparing_dynamic_imports.borrow_mut().clear();
+    self.pending_dynamic_imports.borrow_mut().clear();
+    self.code_cache_ready_futs.borrow_mut().clear();
+  }
+
   pub(crate) fn next_load_id(&self) -> i32 {
     // TODO(mmastrac): move recursive module loading into here so we can avoid making this pub
     let mut data = self.data.borrow_mut();
--- a/core/modules/map.rs
+++ b/core/modules/map.rs
@@ -153,6 +153,7 @@ impl ModuleMap {
     self.preparing_dynamic_imports.borrow_mut().clear();
     self.pending_dynamic_imports.borrow_mut().clear();
     self.code_cache_ready_futs.borrow_mut().clear();
+    std::mem::take(&mut *self.data.borrow_mut());
   }
 
   pub(crate) fn next_load_id(&self) -> i32 {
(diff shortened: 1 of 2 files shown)


core/runtime/jsrealm.rs
```diff
@@ -197,7 +197,8 @@ impl JsRealmInner {
 
       let module_map =
         ctx.get_aligned_pointer_from_embedder_data(MODULE_MAP_SLOT_INDEX);
-      let _ = Box::from_raw(module_map as *mut Rc<ModuleMap>);
+      let module_map = Box::from_raw(module_map as *mut Rc<ModuleMap>);
```
dsherret, line 200: We are we putting an [Rc in a Box](https://www.youtube.com/watch?v=5Cx7jzq2Bx4)?

core/modules/map.rs
```diff
@@ -146,6 +146,14 @@ pub(crate) struct ModuleMap {
 }
 
 impl ModuleMap {
+  /// There is a circular Rc reference between the module map and the futures,
+  /// so when destroying the module map we need to clear the pending futures.
+  pub fn clear_pending_futures(&self) {
+    self.preparing_dynamic_imports.borrow_mut().clear();
+    self.pending_dynamic_imports.borrow_mut().clear();
+    self.code_cache_ready_futs.borrow_mut().clear();
```
bartlomieju, line 154: We should clear all the data before returning this. Just to make sure nothing else holds an `Rc<>` to module map.

core/runtime/jsrealm.rs
```diff
@@ -197,7 +197,8 @@ impl JsRealmInner {
...
+      module_map.clear_pending_futures();
```
bartlomieju, line 201: We should add a debug assert here to check the strong count of `Rc<ModuleMap>`

core/runtime/jsrealm.rs
bartlomieju: @littledivy I missed that in the review. Was there a specific reason we use `Box<Rc<T>>`?

nathanwhit: Yeah I wonder this too. I can't think of a reason it would be necessary - to V8 it's just an opaque pointer, and
 `Rc::into_raw` + `Rc::from_raw` should be fine to use.

bartlomieju (approved):

core/runtime/jsrealm.rs
littledivy: `Rc::into_raw` returns a `*const T` but we want a `*const Rc<T>` to clone it. Another workaround is:

```rust
fn clone_rc_raw(raw: *const T) -> Rc<T> {
  unsafe { Rc::increment_strong_count(raw) };
  unsafe { Rc::from_raw(ptr) }
}
```

littledivy: https://github.com/denoland/deno_core/pull/778

PR #772, opened 2024-06-07
store `ContextState` and `ModuleMap` in embedder slots

littledivy:
Remove usage of rusty_v8 annex slots for storing `ContextState` and `ModuleMap` and instead store them directly in a dedicated embedder field.

Needed for https://github.com/denoland/deno/pull/23976

PR #776, opened 2024-06-07
Revert "store `ContextState` and `ModuleMap` in embedder slots (#772)"

bartlomieju:
This reverts commit 98d52cf97c1574b3ae642b8b4745590d72da7d0c.

CC @littledivy I had trouble updating `deno_core` in Deno as it died on `node_unit_tests::vm_test`
with signal 11 during `vm context promise rejection` test. Also `Box<Rc<...>>` seems fishy as pointed
as by @dsherret in https://github.com/denoland/deno_core/pull/775#discussion_r1631733833

### DIFF
--- a/core/runtime/jsrealm.rs
+++ b/core/runtime/jsrealm.rs
@@ -28,9 +28,6 @@ use std::rc::Rc;
 use std::sync::Arc;
 use v8::Handle;
 
-pub const CONTEXT_STATE_SLOT_INDEX: i32 = 1;
-pub const MODULE_MAP_SLOT_INDEX: i32 = 2;
-
 // Hasher used for `unrefed_ops`. Since these are rolling i32, there's no
 // need to actually hash them.
 #[derive(Default)]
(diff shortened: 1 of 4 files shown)


core/runtime/jsrealm.rs
```diff
@@ -225,32 +203,25 @@ impl JsRealm {
     scope: &mut v8::HandleScope,
   ) -> Rc<ContextState> {
     let context = scope.get_current_context();
-    // SAFETY: slot is valid and set during realm creation
-    unsafe {
-      let rc = context
-        .get_aligned_pointer_from_embedder_data(CONTEXT_STATE_SLOT_INDEX);
-      let rc = &*(rc as *const Rc<ContextState>);
```
dsherret, line 232: It seemed to inconsistently not using a box here?

dsherret (approved):
LGTM

core/runtime/jsrealm.rs
bartlomieju: Yeah, that looks like it

lucacasonato (approved):

core/runtime/jsrealm.rs
littledivy: That is because its a pointer to the Rc we are referencing.

`Box<Rc<ContextState>>::into_raw == *const Rc<ContextState>`

PR #788, opened 2024-06-19
feat: add Option<GcResource> support in ops

lucacasonato:
Optionals are supported both as arguments, via
`#[cppgc] resource: Option<&T>`, and as return
values, via `Option<T>`.

This commit also adds the `GcResource` trait,
which all CppGc resources must implement. This is
done to ensure that one does not accidentally
create CppGc objects for resources that one did
not intend to be managed by CppGc.

### DIFF
--- a/core/lib.rs
+++ b/core/lib.rs
@@ -69,6 +69,7 @@ pub use crate::async_cell::RcLike;
 pub use crate::async_cell::RcRef;
 pub use crate::convert::FromV8;
 pub use crate::convert::ToV8;
+pub use crate::cppgc::GcResource;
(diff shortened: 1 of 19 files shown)


core/lib.rs
```diff
@@ -8,7 +8,7 @@ pub mod arena;
 mod async_cancel;
 mod async_cell;
 pub mod convert;
-pub mod cppgc;
+mod cppgc;
```
littledivy, line 11: Why remove `pub`? `make_cppgc_object` and `try_unwrap_cppgc_object` is used in node:vm

littledivy (approved):
LGTM, one comment and a test needs to be updated

PR #790, opened 2024-06-20
fix: root cppgc objects during async ops

lucacasonato:
Previously cppgc objects were not correctly rooted
during async ops. This commit fixes that.

It also improves lifetimes in the generated op
code. This results in the compiler being able to
reason about lifetimes of function arguments
correctly, preventing compilation of incorrect
code. Previously `#[op2] async fn foo(a: &mut HandleScope)`
would compile for example, even though it is not
safe.

### DIFF
--- a/core/cppgc.rs
+++ b/core/cppgc.rs
@@ -41,6 +41,7 @@ pub fn make_cppgc_object<'a, T: 'static>(
   obj
 }
 
+#[allow(clippy::needless_lifetimes)]
 pub fn try_unwrap_cppgc_object<'sc, T: 'static>(
   val: v8::Local<'sc, v8::Value>,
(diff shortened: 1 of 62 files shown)


devsnek (approved):

core/cppgc.rs
```diff
@@ -41,8 +41,9 @@ pub fn make_cppgc_object<'a, T: 'static>(
   obj
 }
 
+#[allow(clippy::needless_lifetimes)]
```
bartlomieju, line 44: Is this ignore actually needed here?

lucacasonato: No, I could remove the lifetimes. But they are very important to the correctness of the function, so it's better to keep them explicit

Issue #787, opened 2024-06-19
Allow `#[cppgc] &mut T` in sync ops


littledivy:
This will greatly simplify Zlib and net.Blocklist in Deno as we have to wrap them in RefCell for mutability. Should be pretty easy to implement

lucacasonato:
_Copying over my comments from #793:_

I am not sure we can guarantee any safety at all with this:

```rs
struct Wrap(usize);

#[op2(fast)]
async fn op_read(#[cppgc] wrap: &Wrap) {
  println!("foo start: {}", wrap.0);
  tokio::time::sleep(Duration::from_secs(1));
  println!("foo end:   {}", wrap.0);
}

#[op2(fast)]
fn op_write(#[cppgc] wrap: &mut Wrap) {
  wrap.0 += 1;
  println!("bar:       {}", wrap.0);
}
```

```js
const wrap = op_make_wrap();

const promise = op_read(wrap); // foo start: 0
op_write(wrap);                // bar:       1
await promise;                 // foo end:   1
```

Here we are explicitly violating one of the Rust memory safety axioms: a user must not be able to modify a value, while someone has an immutable reference (`&Wrap`), to that value.

I don't think this is fixable by making local changes to what are valid `&mut T` values in an op. You'd have to:

- disallow `&T` in async ops
- ensure the same value can not have both a `&mut T` and `&T` reference within a single sync op
- ensure the sync op is not reentrant

And further, `&mut T` may need to be `Pin<&mut T>` (this I am less sure about).

To then get access to cppgc values in async ops, one would not get a reference, but something like a `Gc<T>`. This `Gc` struct would be `?Send`. The struct would have a `borrow()` / `borrow_mut()`. The value returned from these would be `GcRef<T>` / `GcRefMut<T>` structs that are also `?Send`. They deref to `&T` / `&mut T`. When calling `borrow()` we'd have to ensure at runtime that there is currently no `borrow_mut()` ongoing (and vice versa). You can see, this looks an awful lot like a `RefCell` (with likely similar perf considerations).

That also does not seem viable, because we are now adding the overhead of `RefCell` to all cppgc objects.

[...]

PR #791, opened 2024-06-20
Allow `#[cppgc] &mut T` in sync ops

littledivy:
Allow `#[cppgc] _: &mut T` and `#[cppgc] _: Option<&mut T>`

Closes https://github.com/denoland/deno_core/issues/787

Trying to use it with async op gives a compiler error:
```
error[E0308]: mismatched types
    --> core/runtime/ops.rs:1911:3
     |
1911 |   #[op2(async)]
     |   ^^^^^^^^^^^^^
     |   |
     |   types differ in mutability
     |   arguments to this function are incorrect
     |
     = note: expected mutable reference `&mut TestResource`
                        found reference `&TestResource`
note: associated function defined here
    --> core/runtime/ops.rs:1911:3
     |
1911 |   #[op2(async)]
     |   ^^^^^^^^^^^^^
1912 |   pub async fn op_test_set_cppgc_resource(#[cppgc] resource: &mut TestResource, value: u32) {
```

### DIFF
--- a/core/cppgc.rs
+++ b/core/cppgc.rs
@@ -41,10 +41,16 @@ pub fn make_cppgc_object<'a, T: 'static>(
   obj
 }
 
+#[repr(C)]
+struct InnerMember {
+  inner: [usize; 2],
+  ptr: *mut (),
+}
+
 #[allow(clippy::needless_lifetimes)]
 pub fn try_unwrap_cppgc_object<'sc, T: 'static>(
   val: v8::Local<'sc, v8::Value>,
-) -> Option<&'sc T> {
+) -> Option<&'sc mut T> {
   let Ok(obj): Result<v8::Local<v8::Object>, _> = val.try_into() else {
     return None;
   };
@@ -60,13 +66,13 @@ pub fn try_unwrap_cppgc_object<'sc, T: 'static>(
   // The object can only be created by `make_cppgc_object`.
   let member = unsafe {
     let ptr = obj.get_aligned_pointer_from_internal_field(SLOT_OFFSET);
-    let obj = &*(ptr as *const v8::cppgc::InnerMember);
-    obj.get::<CppGcObject<T>>()
+    let obj = &*(ptr as *mut InnerMember);
+    obj.ptr.cast::<CppGcObject<T>>().as_mut().unwrap()
   };
 
   if member.tag != TypeId::of::<T>() {
     return None;
   }
 
-  Some(&member.member)
+  Some(&mut member.member)
 }
--- a/core/cppgc.rs
+++ b/core/cppgc.rs
@@ -43,6 +43,7 @@ pub fn make_cppgc_object<'a, T: GcResource + 'static>(
   obj
 }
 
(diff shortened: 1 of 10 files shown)


core/cppgc.rs
```diff
@@ -41,10 +41,16 @@ pub fn make_cppgc_object<'a, T: 'static>(
   obj
 }
 
+#[repr(C)]
+struct InnerMember {
+  inner: [usize; 2],
+  ptr: *mut (),
+}
```
littledivy, line 48: TODO: add `v8::cppgc::InnerMember::as_mut` in rusty_v8

bartlomieju (approved):
LGTM

core/cppgc.rs
```diff
@@ -43,10 +43,17 @@ pub fn make_cppgc_object<'a, T: GcResource + 'static>(
...
+
 #[allow(clippy::needless_lifetimes)]
 pub fn try_unwrap_cppgc_object<'sc, T: GcResource + 'static>(
   val: v8::Local<'sc, v8::Value>,
-) -> Option<&'sc T> {
```
lucacasonato, line 49: This is not safe anymore. Please now mark as unsafe. The caller has to guarantee that there are not more than one user currently called `try_unwrap_cppgc_object` on this object.

ops/op2/signature.rs
```diff
@@ -1473,19 +1473,17 @@ fn parse_type_state(ty: &Type) -> Result<Arg, ArgError> {
...
   match (position, ty) {
-    (Position::Arg, Type::Reference(of)) if of.mutability.is_none() => {
-      match &*of.elem {
-        Type::Path(of) => Ok(Arg::CppGcResource(stringify_token(&of.path))),
-        _ => Err(ArgError::InvalidCppGcType(stringify_token(&of.elem))),
-      }
-    }
+    (Position::Arg, Type::Reference(of)) => match &*of.elem {
+      Type::Path(of) => Ok(Arg::CppGcResource(stringify_token(&of.path))),
+      _ => Err(ArgError::InvalidCppGcType(stringify_token(ty))),
```
lucacasonato, line 1478:
```
      _ => Err(ArgError::InvalidCppGcType(stringify_token(*of.elem))),
```

lucacasonato (commented):
I am not clear as to the safety of this. For example, this is currently valid but very unsafe:

```
#[op2(async)]
async fn op_use_cppgc_object_mut_async(#[cppgc] _wrap: &mut Wrap, #[cppgc] _wrap2: &mut Wrap) {}
```

When called with the same object for first and second argument.

lucacasonato (commented):
This should also not be allowed on `#[op2(reentrant)]` (that would be unsafe).

I think we should revert this for now - it's not clear to me how to make this safe yet, and what restrictions we need. We should talk about this more.

Also, for such fundamental issues we need more than one review. These are incredibly delicate components of our stack, and they need to be very well audited, especially when involving unsafe code.

core/cppgc.rs
littledivy: I'll add an unsafe variant that returns `&mut`

PR #796, opened 2024-06-26
Add a v8::Global variant of magic::Value

rscarson:
Allows the direct deserialization of globals across serde boundaries, which will be exceptionally useful in downstream projects

As discussed with @bartlomieju


serde_v8/magic/global_value.rs
```diff
@@ -0,0 +1,52 @@
...
+use crate::magic::transl8::ToV8;
+
+/// serde_v8::Value is used internally to serialize/deserialize values in
+/// objects and arrays. This struct was exposed to user code in the past, but
+/// we don't want to do that anymore as it leads to inefficient usages - eg. wrapping
+/// a V8 object in `serde_v8::Value` and then immediately unwrapping it.
+//
+// SAFETY: caveat emptor, the rust-compiler can no longer link lifetimes to their
+// original scope, you must take special care in ensuring your handles don't
+// outlive their scope.
```
nathanwhit, line 13:
```suggestion
/// A wrapper around `v8::Global<v8::Value>` to allow for passing globals transparently through a serde boundary.
///
/// Serializing a `GlobalValue` creates a `v8::Local` from the global value, and deserializing creates a `v8::Global` from the local value.
```
Feel free to rewrite, just a quick suggestion

nathanwhit (commented):
LGTM aside from the doc comment.

It would also be good to add a simple test to `serde_v8/tests/magic.rs`, if you don't mind.

rscarson:
> LGTM aside from the doc comment.
> 
> It would also be good to add a simple test to `serde_v8/tests/magic.rs`, if you don't mind.

I would need to have the crate export `GlobalValue`
`Value` used to be exported which was changed recently

I would be happy to re-add Value and GlobalValue to serde_v8's exports
Would certainly make it easier to use

And then the existing `Value` can be tested too, which it currently isn't afaik

rscarson:
> LGTM aside from the doc comment.
> 
> It would also be good to add a simple test to `serde_v8/tests/magic.rs`, if you don't mind.

Done!

nathanwhit (approved):
LGTM, thanks!

PR #814, opened 2024-07-09
chore: clean up core.d.ts and other types in the repo

lucacasonato:


ops/op2/dispatch_slow.rs
```diff
@@ -377,6 +377,19 @@ pub fn from_arg(
         };
       }
     }
+    Arg::OptionString(Strings::RefStr) => {
```
bartlomieju, line 380: Is this on purpose?

lucacasonato: Whoops - nope. That's for a different PR!

marvinhagemeister (approved):
LGTM

PR #829, opened 2024-07-17
perf: Make SourceMaps cheaper by using static strs

bartlomieju:
This commits updates `SourceMapper` API to be a bit more performant
by using `ModuleName` (which in case of snapshots is a `&'static str`)
instead of owned `String`.

### DIFF
--- a/core/extension_set.rs
+++ b/core/extension_set.rs
@@ -14,7 +14,6 @@ use crate::ops::OpCtx;
 use crate::runtime::ExtensionTranspiler;
 use crate::runtime::JsRuntimeState;
 use crate::runtime::OpDriverImpl;
-use crate::source_map::SourceMapper;
 use crate::ExtensionFileSource;
 use crate::FastString;
 use crate::GetErrorClassFn;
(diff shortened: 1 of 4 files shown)


core/extension_set.rs
```diff
@@ -274,29 +272,41 @@ pub fn into_sources(
...
+      let (code, maybe_source_map) =
+        load(transpiler, file, &mut load_callback)?;
       sources.lazy_esm.push(LoadedSource {
         source_type: ExtensionSourceType::LazyEsm,
         specifier: ModuleName::from_static(file.specifier),
         code,
       });
+      if let Some(source_map) = maybe_source_map {
+        source_maps.push((ModuleName::from_static(file.specifier), source_map));
+      }
```
littledivy, line 284: nit: maybe this can be a field on `LoadedSource` instead? slightly cleaner and saves a vec alloc. 
```suggestion
      sources.lazy_esm.push(LoadedSource {
        source_type: ExtensionSourceType::LazyEsm,
        specifier: ModuleName::from_static(file.specifier),
        source_map: maybe_source_map
        code,
      });
```

littledivy (approved):
LGTM

PR #830, opened 2024-07-17
perf: Make exception handling faster by using `v8::OneByteConst`

bartlomieju:
This should make handling of exception slightly faster, by
using "v8::OneByteConst" strings from `core/runtime/v8_static_strings.rs`
module. This should save a bunch of mem copies and expensive
allocations when handling errors between JS and Rust.


nathanwhit (approved):
LGTM

Issue #806, opened 2024-07-04
Memory leak in jsruntime `deno_core::ops::OpCtx::new`


spolu:
In a setup where many js runtimes are created (one for each script execution served by a service) running the service in valgrind and dumping the leaks reveal some resources not being freed correctly when the jsruntime is dropped:

```
       Error leaked 2.5 kiB in 860 blocks
        Info at operator new(unsigned long) (vg_replace_malloc.c:483)
             at v8__CTypeInfo__New (binding.cc:2144)
             at new (fast_api.rs:64)
             at deno_core::ops::OpCtx::new (ops.rs:130)
             at deno_core::extension_set::create_op_ctxs (extension_set.rs:140)
             at deno_core::runtime::jsruntime::JsRuntime::new_inner (jsruntime.rs:722)
             at deno_core::runtime::jsruntime::JsRuntime::try_new (jsruntime.rs:624)
             at deno_core::runtime::jsruntime::JsRuntime::new (jsruntime.rs:608)
             at dust::deno::js_executor::JSExecutor::exec::{{closure}} (js_executor.rs:206)
             at <tokio::time::timeout::Timeout<T> as core::future::future::Future>::poll (timeout.rs:202)
             at dust::deno::js_executor::JSExecutor::run::{{closure}}::{{closure}} (js_executor.rs:258)
             at <core::pin::Pin<P> as core::future::future::Future>::poll (future.rs:123)
             at tokio::runtime::scheduler::current_thread::CoreGuard::block_on::{{closure}}::{{closure}}::{{closure}} (mod.rs:659)
       Error leaked 4.7 kiB in 15 blocks
        Info at calloc (vg_replace_malloc.c:1595)
             at calloc (rtld-malloc.h:44)
             at allocate_dtv (dl-tls.c:370)
             at _dl_allocate_tls (dl-tls.c:629)
             at allocate_stack (allocatestack.c:429)
             at pthread_create@@GLIBC_2.34 (pthread_create.c:655)
             at v8::base::Thread::Start() (platform-posix.cc:1223)
             at v8::platform::DefaultWorkerThreadsTaskRunner::DefaultWorkerThreadsTaskRunner(unsigned int, double (*)(), v8::base::Thread::Priority) (default-worker-threads-task-runner.cc:18)
             at v8::platform::DefaultPlatform::EnsureBackgroundTaskRunnerInitialized() (default-platform.cc:141)
[...]

Issue #649, opened 2024-03-08
`OpCtx` leaks `CFunctionInfo`


svix-aaron1011:
When creating a new `JsRuntime`, a `CFunctionInfo` gets allocated: https://github.com/denoland/deno_core/blob/ddbffc518114f862d94c99a4821c647f6aa0ed0d/core/ops.rs#L115-L123

However, it isn't freed when `OpCtx` is dropped, leading to a memory leak

bartlomieju:
@mmastrac can we store it in `IsolateAllocations`?

mmastrac:
I think we just need to free it in Drop

PR #832, opened 2024-07-18
Don't leak CFunctionInfo and CTypeInfo

littledivy:
Fixes https://github.com/denoland/deno_core/issues/649
Fixes https://github.com/denoland/deno_core/issues/806

### DIFF
--- a/core/ops.rs
+++ b/core/ops.rs
@@ -107,10 +107,10 @@ impl Drop for OpCtx {
     if let Some(ptr) = self.fast_fn_c_info {
       // SAFETY: Call drop manually because `fast_fn_c_info` is a `NonNull` and doesn't implement `Drop`.
       unsafe {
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM!

svix-jplatte:
Note that this was done before and reverted, there is an re-apply PR for this at #714, but AFAIU it's blocked on https://github.com/denoland/deno/pull/24169 passing CI.

nathanwhit:
It looks like this PR is passing CI in deno, but the diff looks essentially identical (the only difference being putting that args and ret into a tuple instead of separate fields).

Why does this work while the original PR failed?

littledivy:
I've been trying to figure that out but its likely that the failures were caused by something unrelated back then. This one is  passing so i'll added the original author as co-author and merge.

spolu:
Thanks!

PR #838, opened 2024-07-23
feat(core): Use native stack trace prepare callback instead of `Error.prepareStackTrace`

nathanwhit:
Fixes issues with the `bindings` npm package, as well as `callsites`. 

Moves all error stack trace formatting to rust, allowing us to remove some ops (`op_set_call_site_evals`, `op_apply_source_map`, and some others).

This also moves the stack trace formatting from the `deno` crate to `deno_core` (`format_location` and `format_frame`) this allows us to eliminate some duplication. Previously we had multiple places that were doing formatting that we needed to keep in sync.


core/error.rs
```diff
@@ -210,6 +211,93 @@ pub struct JsStackFrame {
...
+///
+/// # Example
+/// ```ignore
+/// assert_eq!(
+///   parse_eval_origin("eval at foo (bar at (file://a.ts:1:2))"),
+///   Some(("eval at foo (bar at (", ("file://a.ts", 1, 2), "))")),
+/// );
+/// ```
+///
+fn parse_eval_origin(
```
nathanwhit, line 255: Fixes an issue where we weren't actually source mapping the origin for errors from eval. So previously if you had

```ts
// foo.ts



eval("throw new Error()"); // this is line 5
```

You would get

```
Error: Error
    at eval (eval at <anonymous> (file:///foo.ts:2:1), <anonymous>:1:7)
    at file:///foo.ts:5:1
```

Note that the eval part is incorrect `(eval at <anonymous> (file:///foo.ts:2:1)`, it should be line 5 instead of line 2. With this change, you get the right answer.

---

It's pretty minor so I'm fine removing this if it's not worth the complexity

bartlomieju: I think it's fine, can we have a small unit test for it?

core/error.rs
```diff
@@ -773,6 +923,410 @@ pub fn throw_type_error(scope: &mut v8::HandleScope, message: impl AsRef<str>) {
...
+  // stash the callsites on the error object. this is used by `JsError::from_exception_inner`
+  // to get more details on an error, because the v8 StackFrame API is missing some info.
+  if let Ok(obj) = error.try_cast::<v8::Object>() {
+    let key = call_site_evals_key(scope);
+    obj.set_private(scope, key, callsites.into());
+  }
+
+  // `globalThis.Error.prepareStackTrace`
+  let global = scope.get_current_context().global(scope);
+  let global_error = get_property(scope, global, ERROR).unwrap().cast();
```
bartlomieju, line 1108: I wonder if we could get it only once on startup and store it. Maybe not a big deal

bartlomieju (approved):
LGTM; probably good to get a second pair of eyes too. @devsnek can you take a look?

bartlomieju:
This should probably be a `feat` - we're changing the output of stack traces and in a (potentially) breaking way.

devsnek (approved):

PR #853, opened 2024-07-30
feat: Add ImportAssertionsSupport

bartlomieju:
Needed for https://github.com/denoland/deno/pull/24743.

### DIFF
--- a/core/runtime/jsruntime.rs
+++ b/core/runtime/jsruntime.rs
@@ -556,18 +556,32 @@ pub struct RuntimeOptions {
   pub import_assertions_support: ImportAssertionsSupport,
 }
 
+pub struct ImportAssertionsSupportCustomCallbackArgs {
+  pub maybe_specifier: Option<String>,
+  pub maybe_line_number: Option<usize>,
+  pub column_number: usize,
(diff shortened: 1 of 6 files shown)


core/runtime/jsruntime.rs
```diff
@@ -550,6 +552,78 @@ pub struct RuntimeOptions {
...
+          eprintln!("{:0width$}^^^^^^", " ", width = start_column);
+        }
+        eprintln!(
+          "  at {}:{}:{}",
+          specifier.to_rust_string_lossy(scope),
+          maybe_line_number.unwrap(),
+          start_column
+        );
+      }
+    }
```
bartlomieju, line 621: This prints a message like:
```
⚠️  Import assertions are deprecated. Use `with` keyword, instead of 'assert' keyword.

import data from "./data.json" assert { type: "json" };

                               ^^^^^^
  at file:///Users/ib/dev/deno/foo.js:1:31
```

Kinda feel we might want to reuse it in the CLI, just print it in yellow (with a `Warning:` prefix). Need to figure out how to easily share it

lucacasonato (approved):
LGTM

PR #856, opened 2024-07-31
fix: Add missing `toString` to callsites, handle user provided eval origin

### LINKED ISSUE #854: `callSite.toString()` returns `"[object Object]"` since 0.299.0
```rs
use deno_core::{JsRuntime, RuntimeOptions};

fn main() {
  let code = r#"
    Error.prepareStackTrace = (error, structuredStackTrace) => {
      let stack = Error.prototype.toString.call(error);
      for (const callSite of structuredStackTrace) {
        stack += `\n    at ${callSite}`;
      }
      return stack;
    };

    Deno.core.ops.op_print(new Error().stack, false);
  "#;
  let mut runtime = JsRuntime::new(RuntimeOptions::default());
  runtime.execute_script("", code).unwrap();
}
```

Expected output:

```text
Error
    at <anonymous>:10:28
```

Actual output:

```text
Error
    at [object Object]
```

nathanwhit:
Fixes #854. Fixes https://github.com/denoland/deno/issues/24782 once this lands there.

The first issue is pretty straightforward, I didn't realize that the `Callsite` object had a `toString` method (though in retrospect it makes sense).

The second issue is that we try to do some parsing of the eval origin so we can source map it, but if the origin wasn't in the form we expected we would return undefined instead of leaving it alone. 

So in this case

```js
const AsyncFunction = Object.getPrototypeOf(async function () {
  // empty
}).constructor;

const func = new AsyncFunction(
  `return doesNotExist();
    //# sourceURL=empty.eval`,
);

func.call({});
```
We try to parse the `evalOrigin` (which is `"empty.eval"`) and since it's not in the expected format (the form that comes from v8 directly) we were returning `None` (which became `null` once it got to JS). Instead, we should just leave it alone and return it directly.

(It turns out that if users specify the `sourceURL` in evaled code it becomes the evalOrigin verbatim)

### DIFF
--- a/core/error.rs
+++ b/core/error.rs
@@ -950,6 +950,7 @@ v8_static_strings::v8_static_strings! {
   GET_PROMISE_INDEX = "getPromiseIndex",
   PREPARE_STACK_TRACE = "prepareStackTrace",
   ORIGINAL = "_orig",
+  TO_STRING = "toString",
 }
 
 #[inline(always)]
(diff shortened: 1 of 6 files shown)


bartlomieju (approved):
LGTM! Please update PR title before landing

PR #858, opened 2024-08-01
fix: Error.pST callsite methods should be on prototype

lucacasonato:
- **chore: report cause when JsStackFrame fails to deserialize**
- **fix: Error.pST callsite methods should be on prototype**

### DIFF
--- a/core/runtime/v8_static_strings.rs
+++ b/core/runtime/v8_static_strings.rs
@@ -12,7 +12,7 @@ pub(crate) use v8_static_strings;
 v8_static_strings!(
   BUILD_CUSTOM_ERROR = "buildCustomError",
   CALL_CONSOLE = "callConsole",
-  CALL_SITE_EVALS = "callSiteEvals",
+  CALL_SITE_EVALS = "deno_core::call_site_evals",
   CAUSE = "cause",
   CODE = "code",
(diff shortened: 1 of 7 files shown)


core/runtime/v8_static_strings.rs
```diff
@@ -12,7 +12,7 @@ pub(crate) use v8_static_strings;
 v8_static_strings!(
   BUILD_CUSTOM_ERROR = "buildCustomError",
   CALL_CONSOLE = "callConsole",
-  CALL_SITE_EVALS = "callSiteEvals",
+  CALL_SITE_EVALS = "deno_core::call_site_evals",
```
bartlomieju, line 15: This looks like a bug?

lucacasonato: No I changed this on purpose. V8's docs for `v8::Private::for_api`:

```
  /// To minimize the potential for clashes, use qualified names as keys,
  /// e.g., "Class#property".
```

The qualified name for the internal property `call_site_evals` is `deno_core::call_site_evals` (this string never makes it to user visible JS).

bartlomieju (approved):
LGTM, once CI passes

marvinhagemeister (approved):

PR #865, opened 2024-08-05
fix: add `getScriptNameOrSourceURL` to `CallSite` prototype

marvinhagemeister:
Both Chrome and Node have a `getScriptNameOrSourceUrl()` method on the `CallSite` prototype. We didn't have that which causes an error in `vitest`, see https://github.com/denoland/deno/issues/24898


bartlomieju (approved):

lucacasonato (approved):
This is not be error prototype. It's the v8 callsite object prototype. Please update the commit message 

Implementation looks good though

marvinhagemeister:
Oops good point. Updated the commit message and PR here to refer to the `CallSite` prototype instead.

devsnek (approved):

PR #869, opened 2024-08-12
fix: don't panic when eval_origin is None

lucacasonato:


bartlomieju:
Is there a test we could add?

lucacasonato:
Yeah, working on it. Unfortunately quite difficult because according to V8 source code this shouldn't be able to happen :sweat_smile:

bartlomieju:
Take a look at example here: https://github.com/denoland/deno/issues/24782 I think it had the same (or similar) problem

lucacasonato:
It's a different problem. It's denoland/deno#24990

PR #878, opened 2024-08-21
refactor: rework error handling

### LINKED ISSUE #810: Remove usages of `anyhow::Error` in library code
Instead any error type that needs to cross Rust->JS boundary should require to implement `ToV8Error` trait.

Example "footgun" usage:
https://github.com/denoland/deno_core/blob/0f7fced6e574393f2673461b6e9fc12a461eca23/core/error.rs#L126C8-L126C19

crowlKats:
Closes #810


core/runtime/op_driver/op_results.rs
```diff
@@ -263,21 +263,64 @@ impl<C: OpMappingContext> OpResult<C> {
...
-  #[serde(rename = "$err_class_name")]
-  class_name: &'static str,
-  message: String,
-  code: Option<&'static str>,
+#[derive(Debug)]
+pub struct OpError(Box<dyn JsErrorClass>);
+
+impl std::error::Error for OpError {
+  fn source(&self) -> Option<&(dyn Error + 'static)> {
+    todo!()
```
nathanwhit, line 271: could this `todo!` cause trouble? regardless, could probably just leave it the default impl for now instead of panicking

ops/op2/signature.rs
```diff
@@ -819,19 +819,25 @@ pub enum SignatureError {
   RetError(#[from] RetError),
   #[error("Only one lifetime is permitted")]
   TooManyLifetimes,
-  #[error("Generic '{0}' must have one and only bound (either <T> and 'where T: Trait', or <T: Trait>)")]
+  #[error("Generic '{0}' must have one and only bound (either <T> and 'where T: Trait', or <T: Trait>)"
```
nathanwhit, line 822: formatting change seems wonky

core/error.rs
```diff
@@ -1,143 +1,260 @@
...
+    scope: &mut v8::HandleScope,
+  ) -> v8::Global<v8::Value> {
+    let err_string = self.get_message().to_string();
+    let mut error_chain = vec![];
+    let mut intermediary_error: Option<&(dyn Error)> = Some(&self);
+
+    while let Some(err) = intermediary_error {
+      if let Some(source) = err.source() {
+        let source_str = source.to_string();
+        if source_str != err_string {
```
nathanwhit, line 116: Is this check necessary?

core/error.rs
```diff
@@ -1,143 +1,260 @@
...
+    SYNTAX_ERROR => v8::Exception::syntax_error(scope, message),
+    _ => v8::Exception::error(scope, message),
   }
 }
 
 pub fn to_v8_error<'a>(
   scope: &mut v8::HandleScope<'a>,
-  get_class: GetErrorClassFn,
-  error: &Error,
+  error: &impl JsErrorClass,
```
nathanwhit, line 257: If it doesn't hurt perf, maybe use `dyn` instead of `impl` here so this function isn't duplicated for each error type? May not be a big deal though

nathanwhit (approved):
LGTM! Just a handful of small comments that don't necessarily need to be addressed in this PR

Cargo.lock
```diff
@@ -567,7 +567,7 @@ dependencies = [
  "swc_visit",
  "swc_visit_macros",
  "text_lines",
- "thiserror",
+ "thiserror 1.0.60",
```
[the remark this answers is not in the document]
crowlKats: yep, we have been slowly doing this

core/runtime/op_driver/op_results.rs
crowlKats: removed the source impl as not necessary

core/benches/snapshot/snapshot.rs
```diff
@@ -50,10 +50,22 @@ fn make_extensions_ops() -> Vec<Extension> {
   fake_extensions!(init_ops, a, b, c, d, e, f, g, h, i, j, k, l, m, n)
 }
 
+deno_error::js_error_wrapper!(
+  deno_ast::ParseDiagnostic,
+  JsParseDiagnostic,
+  "TypeError"
```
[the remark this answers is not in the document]
crowlKats: surprisingly causes issues with clippy; this code can be removed in a follow-up anyways by updating deno_ast

core/error.rs
crowlKats: good point; done. found another spot where i could do this as well so done there too.

core/error.rs
crowlKats: i believe we were doing similar handling like this somewhere before, and when i tried removing it, some stuff broke, so didnt tihnk too much about it and just kept it as it is

nathanwhit: oh interesting, didn't know we were already doing something like this. seems fine then

crowlKats: or i might be going insane

PR #881, opened 2024-08-23
fix: improve `extension!` macro

magurotuna:
This commit makes the following refinements for `extension!` macro:

- Improve the doc comments to cover all the fields available now (i.e. `external_references` and `customizer`)
- Fix the wrong field name in the doc (`config` -> `options`)
- Fix the compile error regarding a wrong expansion of repetitive `external_references` elements
- Let it correctly handle the cases where `options` are provided but `state_fn` is not. For example, the previous implementation emits a compile error for the code below, which is fixed in this patch:

```rs
deno_core::extension!(
  my_ext,
  options = { value: String },
);
```

```
error: expected expression, found `;`
  --> src/bin/deno_core_extension.rs:15:1
   |
15 | / deno_core::extension!(
16 | |     my_ext,
17 | |     options = { value: String },
18 | | );
   | |_^ expected expression
   |
   = note: this error originates in the macro `$crate::extension` which comes from the expansion of the macro `deno_core::extension` (in Nightly builds, run with -Z macro-backtrace for more info)

```


magurotuna:
not sure if this is still revevant. closing it for now

PR #889, opened 2024-08-29
Fix aliasing bugs revealed by Tree Borrows

### LINKED ISSUE #884: Tree Borrows support
Hi all,

I'm one of the people working on [Tree Borrows](https://www.ralfj.de/blog/2023/06/02/tree-borrows.html), a candidate for a successor for Stacked Borrows. Stacked Borrows (and similarly Tree Borrows) of course aims to make Rust's references usable for optimizations. It's also included in Miri, and since you test `deno_core` under Miri, you already ensure that you comply with Stacked Borrow's rules.

While we designed Tree Borrows to be less strict in many cases, there are unfortunately some cases where it's more strict than SB (mostly because before, SB had some "dirty hacks"). It turns out that `deno_core` relies on such a hack. You can see it yourself by running miri with `MIRIFLAGS="-Zmiri-tree-borrows"`.

The precise issue is hiding [here](https://github.com/denoland/deno_core/blob/main/core/runtime/op_driver/future_arena.rs#L106):
```
  fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
    // SAFETY: We know the underlying futures are both pinned by their allocations
    unsafe {
      match self.get_mut() {
        Self::Arena(a) => {
          Pin::new_unchecked(a.ptr.assume_init().as_mut()).poll(cx) // the `a.ptr`
        }
        Self::Box(b) => b.as_mut().poll(cx),
      }
    }
  }
}
```
The code `a.ptr` in the `Self::Arena` case uses the `Deref` trait implicitly, and this creates a shared reference to the entire `DynFutureInfoErased`, including the `data` field which is pinned somewhere else., even though you only ever use this to access the `ptr` field (as far as I can see). Unfortunately, this invalidates the (pinned) mutable reference to the second field (which stores the future's internal data), causing futures to randomly have UB later (when the code `rustc` desugared them to does a write to the implicit struct storing locals).

One solution (that seems to work, according to preliminary testing) would be to first change the `DynFutureInfoErased` to the following:

```
struct DynFutureInfoErased<T, C> {
  ptr: MaybeUninit<NonNull<dyn ContextFuture<T, C>>>,
  data: UnsafeCell<TypeErased<MAX_ARENA_FUTURE_SIZE>>, // the unsafecell is new
}
```

And to then ensure that `ArenaBox` never gives out mutable references (since `UnsafeCell` only does its magic on shared ones).

I'm in the process of developing [such a change](https://github.com/JoJoDeveloping/deno_core/tree/tb-fix-approach-2-unsafecell), and it is not too invasive so far. But before I waste too much time on it, I would like to hear whether you'd appreciate it. I am also not sure I fully understand the internals of `deno_core`, so perhaps you have some more opinions/guidance/ideas on what, if any, you want to have done here.

JoJoDeveloping:
This PR fixes some aliasing bugs caused by improper usage of shared and mutable references intermixed with raw pointer accesses. These have been discovered by running the testsuite under [Tree Borrows](https://www.ralfj.de/blog/2023/06/02/tree-borrows.html), a new aliasing model for Rust and a potential replacement for Stacked Borrows. While TB is in general more lenient than SB, there are some cases where it is more strict due to SB relying on gross hacks, which have been removed from TB.

Turns out that `deno_core` relied on such hacks. The problem is with the `DynFutureInfoErased`, which has arbitrary data that is being modified by a future while other references to it are held in the runtime. To prevent this, that is put into an `UnsafeCell` which makes such modifications allowed.

But `UnsafeCell` only ensures that shared references can be modified without causing UB, it does _not_ permit mutable references to be aliased. Thus, one should take care not to create mutable references here. The easiest way of doing this is to remove the implementation of `DerefMut` and `AsMut` for `ArenaBox<T>`, and work around the few cases where that leads to breakage (by using raw pointers, which is also more robust in terms of the aliasing model).

This fixes #884. See that issue for more information.

I have not audited the entire crate, so it is possible that other TB bugs are lurking elsewhere; especially in the parts that Miri can currently not test properly. But this at least makes all the test that could work, do work.

### DIFF
--- a/core/arena/unique_arena.rs
+++ b/core/arena/unique_arena.rs
@@ -60,8 +60,8 @@ impl<T: 'static> ArenaBox<T> {
   /// This function returns a raw pointer without managing the memory, potentially leading to
   /// memory leaks if the pointer is not properly handled or deallocated.
   #[inline(always)]
-  pub fn into_raw(mut alloc: ArenaBox<T>) -> NonNull<T> {
-    let ptr = NonNull::from(alloc.data_mut());
+  pub fn into_raw(alloc: ArenaBox<T>) -> NonNull<T> {
+    let ptr: NonNull<ArenaBoxData<T>> = alloc.ptr;
     std::mem::forget(alloc);
     unsafe { Self::ptr_from_data(ptr) }
   }
@@ -94,8 +94,10 @@ impl<T> ArenaBox<T> {
   }
 
   #[inline(always)]
-  fn data_mut(&mut self) -> &mut ArenaBoxData<T> {
-    unsafe { self.ptr.as_mut() }
+  pub(crate) fn deref_data(&self) -> NonNull<T> {
+    unsafe {
+      NonNull::new_unchecked(std::ptr::addr_of_mut!((*self.ptr.as_ptr()).data))
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


devsnek (approved):
i don't have any context on this code specifically but the overall changes seem good for an arena-shaped api

JoJoDeveloping:
Hi, is there any update on this? I rebased this on top of `main`, the checks all pass, is there anything you still want me to do?

devsnek (approved):

JoJoDeveloping:
Thanks a lot for merging 🎉

Issue #885, opened 2024-08-28
Evaluate updated module


gearonix:
```rust
async fn run_js(file_path: &str) -> Result<(), AnyError> {
    let main_module = deno_core::resolve_path(file_path, env::current_dir()?.as_path())?;
    let mut js_runtime = deno_core::JsRuntime::new(deno_core::RuntimeOptions {
        module_loader: Some(Rc::new(TsModuleLoader)),
        startup_snapshot: Some(RUNTIME_SNAPSHOT),
        extensions: vec![runjs::init_ops()],
        ..Default::default()
    });

    fs::write(file_path, "console.log(\"1\")")?;

    // js_runtime.lazy_load_es_module_with_code()
    let mod_id = js_runtime.load_main_es_module(&main_module).await?;
    let _ = js_runtime.mod_evaluate(mod_id).await?;
    // [out]: "hello 1"
    js_runtime.run_event_loop(Default::default()).await?;

    fs::write(file_path, "console.log(\"2\")")?;
    
    // should log [out]: "hello 2"
    let _ = js_runtime.mod_evaluate(mod_id).await?;
    // got error
    // assertion `left == right` failed: Module already evaluated.
    // Perhaps you've re-provided a module or extension that was already included in the snapshot

    Ok(())
}
```
Should I create a new JsRuntime instance and load the updated file?

marvinhagemeister:
`.mod_evaluate` evaluates modules that have already been loaded into memory. It doesn't reload files from disk.

gearonix:
```rust
    let main_module = deno_core::resolve_path(file_path, env::current_dir()?.as_path())?;
 
    fs::write(file_path, "console.log(\"1\")")?;

    // js_runtime.lazy_load_es_module_with_code()
    let mod_id = js_runtime.load_side_es_module(&main_module).await?;
    let _ = js_runtime.mod_evaluate(mod_id).await?;
    // [out]: "hello 1"
    js_runtime.run_event_loop(Default::default()).await?;

    // main_module updated here, now it contains 
    // source code below
    fs::write(file_path, "console.log(\"2\")")?;

    // Maybe something like .update_side_es_module?
    let mod_id = js_runtime.load_side_es_module(&main_module).await?;

    let _ = js_runtime.mod_evaluate(mod_id).await?;
    // should log [out]: "hello 2"
    // but got error
    // assertion `left == right` failed: Module already evaluated.
    // Perhaps you've re-provided a module or extension that was already included in the snapshot
[...]

PR #890, opened 2024-08-29
feat: `remove_es_module`  

gearonix:
Related issue: https://github.com/denoland/deno_core/issues/885
Example usage: https://github.com/kurtexdev/kurtex/blob/main/crates/kurtex_core/src/deno/runtime.rs#L110

Look at the tests to understand in more detail what the functionality is about. The point is to update existing V8 modules with the old `JsRuntime` instance, without creating new ones.

If you have any ideas, let me know, and I'll finish this PR.

### DIFF
--- a/core/modules/map.rs
+++ b/core/modules/map.rs
@@ -248,6 +248,7 @@ impl ModuleMap {
     self.data.borrow().get_id(name, requested_module_type)
   }
 
+  // Removes a module or its alias from the module map.
   pub(crate) fn remove_id(
     &self,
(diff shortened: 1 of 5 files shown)


devsnek (changes requested):
Can you elaborate on the use case here? If you just want hot reloading, for example, you should absolutely not do this and use the [inspector hot patching api](https://chromedevtools.github.io/devtools-protocol/tot/Debugger/#method-setScriptSource) instead.

PR #895, opened 2024-09-05
fix: explicitly set the named lifetime to static

irbull:
Due to a change in Rust nightly, the Rust linter now warns when an elided lifetime ends up being a named lifetime [1]. In this case, the elided lifetime is `'static`, so we should explicitly set the lifetime to `'static`.

[1] https://github.com/rust-lang/rust/pull/129207


marvinhagemeister (approved):
Sweet, thanks!

PR #896, opened 2024-09-05
chore: remove `core.metrics()`

iuioiua:
As `Deno.metrics()` is now removed from the runtime. Disclaimer: I'm not very familiar with this codebase. I'm not sure if there are other things that could be cleaned up.


core/ops_metrics.rs
```diff
@@ -116,76 +113,3 @@ impl OpMetricsSummary {
     self.ops_dispatched_async > self.ops_completed_async
   }
 }
-
-#[derive(Default, Debug)]
-pub struct OpMetricsSummaryTracker {
```
iuioiua, line 121: IIUC, `core.metrics()` was the only symbol that used this struct.

irbull:
I've filed a PR for the failing builds in deno_core. https://github.com/denoland/deno_core/pull/895

core/ops_metrics.rs
bartlomieju: I believe this is actually used by the `deno test` for leaking ops discovery. This should most likely stay. See `runtime/worker.rs` in Deno repo.

bartlomieju (changes requested):

bartlomieju (approved):

bartlomieju (approved):

iuioiua:
Please merge, @bartlomieju

PR #897, opened 2024-09-05
chore: remove `core.resources()`

iuioiua:
As `Deno.resources()` has been removed from the runtime.


bartlomieju:
I would keep this one for debugging purposes in Deno CLI - if it's not exposed on the `Deno` namespace that should be good enough. Sometimes working on certain APIs it's useful to be able to dump all existing resources in the table.

PR #906, opened 2024-09-14
perf: fast path for cached dyn imports

littledivy:
This patch adds a fast path for already loaded and evaluated dynamic imports by immediately resolving the promise and skipping most of the event loop machinery.

image

```js
// Warmup
await import('./other.mjs')
await import('./other.mjs')
await import('./other.mjs')
await import('./other.mjs')
await import('./other.mjs')

const start = Date.now()
for (let i = 0; i < 100000; i++) {
    await import('./other.mjs')
}
```

### DIFF
--- a/core/modules/map.rs
+++ b/core/modules/map.rs
@@ -941,16 +941,41 @@ impl ModuleMap {
   // Initiate loading of a module graph imported using `import()`.
   pub(crate) fn load_dynamic_import(
     self: Rc<Self>,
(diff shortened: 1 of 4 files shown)


bartlomieju (commented):
Can you open a PR to Deno first before we land it?

littledivy:
@bartlomieju https://github.com/denoland/deno/pull/25636

devsnek (approved):

PR #912, opened 2024-09-18
fix(ops): do not coerce to booleans

lucacasonato:


littledivy (commented):
Can you open an draft PR in denoland/deno? This might break some ops

devsnek (approved):

littledivy (approved):
LGTM. Deno CI passes

PR #914, opened 2024-09-22
Fix panic in `prepare_stack_trace_callback` when global interceptor throws

littledivy:
Fix panic when a global interceptor is misconfigured or throws an exception. 

Fixes https://github.com/denoland/deno/pull/24985#issuecomment-2365460210 and https://github.com/denoland/deno/issues/26240


bartlomieju (approved):

bartlomieju:
Maybe a short unit test?

PR #943, opened 2024-10-24
feat: `FastString` - ability to unsafely specify if string is ASCII or not

### LINKED ISSUE #942: Ability to unsafely create `FastString` that is ascii or not
There's a perf cost to checking if a string is ascii:

```rs
pub const fn from_static(s: &'static str) -> Self {
    if s.is_ascii() {
      Self {
        inner: FastStringInner::StaticAscii(s),
      }
    } else {
      Self {
        inner: FastStringInner::Static(s),
      }
    }
  }
```

...if this information is known upfront, maybe there could be an API to tell deno_core this unsafely? I'm creating strings from bytes that are `&'static [u8]` for `deno compile` and it would be better to store upfront if they're ascii for perf reasons.

dsherret:
Closes #942

### DIFF
--- a/core/fast_string.rs
+++ b/core/fast_string.rs
@@ -175,12 +175,18 @@ impl FastString {
   /// Create a [`FastString`] from a static string that is known to contain
   /// only ASCII characters.
   ///
+  /// Note: This function is deliberately not `const fn`. Use `from_static`
+  /// in const contexts.
+  ///
+  /// # Safety
+  ///
+  /// It is unsafe to specify a non-ASCII string here because this will be
+  /// referenced in an external one byte static string in v8, which requires
+  /// the data be Latin-1 or ASCII.
+  ///
   /// This should only be used in scenarios where you know a string is ASCII
   /// and you want to avoid the performance overhead of checking if a string
   /// is ASCII that `from_static` does.
-  ///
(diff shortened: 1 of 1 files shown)


core/fast_string.rs
```diff
@@ -171,6 +172,47 @@ impl FastString {
...
+  /// # Safety
+  ///
+  /// It is unsafe to specify a non-ASCII string here because this will be
+  /// referenced in an external one byte static string in v8, which requires
+  /// the data be Latin-1 or ASCII.
+  ///
+  /// This should only be used in scenarios where you know a string is ASCII
+  /// and you want to avoid the performance overhead of checking if a string
+  /// is ASCII that `from_static` does.
+  pub unsafe fn from_ascii_static_unsafe(s: &'static str) -> Self {
```
nathanwhit, line 190:
```suggestion
  pub unsafe fn from_ascii_static_unchecked(s: &'static str) -> Self {
```

bikeshedding, but feels more consistent with e.g. `str::from_utf8_unchecked`

nathanwhit (approved):
LGTM!

core/fast_string.rs
dsherret: Good idea.

PR #946, opened 2024-10-25
fix: return wrapped module instead of original module

nathanwhit:
We were adding the `__esModule` export as intended, but then returning the original module instead of the wrapped one.


dsherret (approved):
LGTM

PR #959, opened 2024-11-07
#[stack_trace] in fast calls

littledivy:


crowlKats (approved):
LGTM, tested and works

PR #965, opened 2024-11-17
Named accessor ops

littledivy:
Implement get/set accessors for object wrap ops.

```rust
#[op2]
impl DOMPoint {
  #[getter]
  fn x(&self) -> f64 {
    self.x
  }
  
  #[setter]
  fn x(&self, x: f64) {}
}
```

```js
point.x;
point.x = 10; 
```


core/runtime/bindings.rs
```diff
@@ -456,6 +515,44 @@ fn op_ctx_function<'s>(
...
+fn create_accessor_store(ctx: &OpMethodCtx) -> AccessorStore {
+  let mut store = AccessorStore::new();
+
+  for method in ctx.methods.iter() {
+    // Populate all setters first.
+    if method.decl.accessor_type == AccessorType::Setter {
+      let key = method.decl.name_fast.to_string();
+
+      // All setters must start with "__set_".
+      let key = key.strip_prefix("__set_").expect("Invalid setter name");
```
bartlomieju, line 529: Is there a particular reason for this? Why don't getters require to start with `__get_`?

bartlomieju (approved):
Very nice, LGTM!

core/runtime/bindings.rs
littledivy: This is to allow for getter and setters to be defined with the same name. Only one needs to be changed to avoid name collision

```rust
#[getter]
fn x() {}

#[setter]
fn x() {} // op2 automatically converts this to __set_x to avoid name collision
```

PR #966, opened 2024-11-18
Default to camelCase for object wrap ops

littledivy:
`fn from_point` in Rust is exposed as `.fromPoint` in JS. I think this is better than having an explict rename attribute since that's very rare to be used.

### DIFF
--- a/testing/unit/resource_test.ts
+++ b/testing/unit/resource_test.ts
@@ -67,8 +67,8 @@ test(async function testCppgcAsync() {
 test(function testDomPoint() {
   const p1 = new DOMPoint(100, 100);
   const p2 = new DOMPoint();
-  const p3 = DOMPoint.from_point({ x: 200 });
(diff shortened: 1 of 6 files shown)


testing/unit/resource_test.ts
```diff
@@ -67,8 +67,8 @@ test(async function testCppgcAsync() {
 test(function testDomPoint() {
   const p1 = new DOMPoint(100, 100);
   const p2 = new DOMPoint();
-  const p3 = DOMPoint.from_point({ x: 200 });
-  const p4 = DOMPoint.from_point({ x: 0, y: 100, z: 99.9, w: 100 });
+  const p3 = DOMPoint.fromPoint({ x: 200 });
+  const p4 = DOMPoint.fromPoint({ x: 0, y: 100, z: 99.9, w: 100 });
```
bartlomieju, line 71: TSC complains about this one

bartlomieju (approved):

PR #976, opened 2024-11-21
Add `#[required(<count>)]`

littledivy:
Equivalent to Deno's `webidl.requiredArguments` internal.

```rust
#[fast]
#[required(2)]
fn set_item(
  &self,
  state: &mut OpState,
  #[string] key: &str,
  #[string] value: &str,
) {
 // ...
}
```

### DIFF
--- a/ops/op2/dispatch_slow.rs
+++ b/ops/op2/dispatch_slow.rs
@@ -120,6 +120,13 @@ pub(crate) fn generate_dispatch_slow(
     quote!()
   };
 
+  let with_required_check = if generator_state.needs_args && config.required > 0
+  {
+    with_required_check(generator_state, config.required)
+  } else {
+    quote!()
+  };
+
   let with_self = if generator_state.needs_self {
     with_self(generator_state, &signature.ret_val)
       .map_err(V8SignatureMappingError::NoSelfMapping)?
@@ -143,6 +150,7 @@ pub(crate) fn generate_dispatch_slow(
         #with_scope
         #with_retval
         #with_args
+        #with_required_check
         #with_opctx
         #with_self
         #with_isolate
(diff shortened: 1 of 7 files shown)


ops/op2/dispatch_slow.rs
```diff
@@ -227,6 +235,34 @@ pub(crate) fn with_fn_args(
...
+  required: u8,
+) -> TokenStream {
+  generator_state.needs_scope = true;
+  let arguments_lit = if required > 1 {
+    "arguments"
+  } else {
+    "argument"
+  };
+  gs_quote!(generator_state(fn_args, scope, self_ty) =>
+    (if #fn_args.length() < #required as i32 {
```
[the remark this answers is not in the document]
littledivy: `#fn_args.length() < #required` => `1 < 2` so it won't pass

ops/op2/dispatch_slow.rs
```diff
@@ -227,6 +235,34 @@ pub(crate) fn with_fn_args(
...
+      let msg = format!(
+        "Failed to execute '{}' on '{}': {} {} required, but only {} present",
```
crowlKats, line 251: two notes on this:
- `Failed to execute '{}' on '{}':` we will need to extract this out when we add more of the webidl conversion so we dont duplicate this. no action needed now, just noting for future.
- this is only applicable for methods. could you handle normal functions as well? ie https://github.com/denoland/deno/blob/main/ext/canvas/01_image.js#L158-L166
- Same applies for constructors, as those say `Failed to construct '{}':`

littledivy: - noted.
- ah didn't know that. this pr is only for object wraps but easy to have this on functions too.
- yup will fix that, ig it also needs to be 'Failed to construct {}.{}` for static methods

crowlKats (approved):
LGTM

PR #981, opened 2024-11-23
Use wrapping cast for #[smi] arguments

littledivy:
This is to match the default equivalent WebIDL converter. I don't know of any ops in Deno that depend on the clamping behavior.

### DIFF
--- a/testing/unit/resource_test.ts
+++ b/testing/unit/resource_test.ts
@@ -82,4 +82,8 @@ test(function testDomPoint() {
     caught = e;
   }
   assert(caught);
+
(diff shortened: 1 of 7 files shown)


testing/unit/resource_test.ts
```diff
@@ -82,4 +82,13 @@ test(function testDomPoint() {
     caught = e;
   }
   assert(caught);
+
+  assertEquals(p1.wrappingSmi(4294967295), 4294967295);
```
bartlomieju, line 86: Add a comment on this "magic" integer (I assume it's usize::max)

PR #986, opened 2024-11-27
fix: FastString - v8_string should error when cannot allocate

dsherret:


core/error.rs
```diff
@@ -167,7 +167,7 @@ pub fn to_v8_error<'a>(
 pub(crate) fn call_site_evals_key<'a>(
   scope: &mut v8::HandleScope<'a>,
 ) -> v8::Local<'a, v8::Private> {
-  let name = v8_static_strings::CALL_SITE_EVALS.v8_string(scope);
+  let name = v8_static_strings::CALL_SITE_EVALS.v8_string(scope).unwrap();
```
dsherret, line 170: This unwrap has been pushed up to the caller. There's probably some questionable places we do this, but I didn't inspect them thoroughly because it's existing behaviour.

bartlomieju:
Ref back to https://github.com/denoland/deno/pull/27033/files#diff-6bad3474ae87d5566f34fbe0c2af44650f1aa8fedfecc0dd39147d91db49c8a4R1387 to remove that code once this lands

littledivy (approved):

core/error.rs
littledivy: Yeah, looks good

dsherret:
I should have marked this as breaking actually.

PR #987, opened 2024-12-03
feat: Add ModuleLoader::finish_load hook

bartlomieju:


dsherret (approved):
LGTM

core/modules/loaders.rs
```diff
@@ -83,6 +83,18 @@ pub trait ModuleLoader {
...
 
+  /// This hook can be used by implementors to do some cleanup
+  /// work after loading of modules.
+  ///
+  /// For example implementor might drop transpilation and
+  /// static analysis caches before
+  /// yielding control back to the runtime.
+  ///
+  /// It's not required to implement this method.
+  fn finish_load(&self) -> Pin<Box<dyn Future<Output = Result<(), Error>>>> {
```
bartlomieju, line 94: Add a test that ensures this works if module load succeeds and fails.

bartlomieju, line 94: Making this async makes it really painful to handle all scenarios - if we made it sync it would be much easier to ensure that both success and failure in module load properly calls this hook. Given that the CLI is most likely gonna be the only consumer of it, I'm inclined to make it sync.

PR #992, opened 2024-12-09
perf: calculate capacity of generated Wasm module code

dsherret:
100_000 iterations of `test_render_js_wasm_module`

Before: 400.4625ms
After: 239.3486ms


core/modules/map.rs
```diff
@@ -1996,100 +1997,118 @@ pub fn script_origin<'a>(
...
-    imports: &[wasm_dep_analyzer::Import],
-  ) -> IndexMap<String, Vec<String>> {
-    let mut imports_map = IndexMap::default();
+  fn aggregate_wasm_module_imports<'a>(
+    imports: &'a [wasm_dep_analyzer::Import],
+  ) -> IndexMap<&'a str, ImportInfo> {
+    let mut imports_map = IndexMap::with_capacity(imports.len());
 
     for import in imports.iter().filter(|i| {
       matches!(i.import_type, wasm_dep_analyzer::ImportType::Function(..))
```
dsherret, line 2011: Btw, this seems wrong. Shouldn't we be importing all imports?

bartlomieju: Good point, I haven't seen a Wasm module though that had something else an import. But we definitely should support it. Let's track in a new issue though.

bartlomieju (approved):

PR #995, opened 2024-12-09
ci: only run test-valgrind on main

dsherret:
It takes too long to run it. I think it's worth it to only run on main.


dsherret:
Ok, I'm not spending my time figuring this out.

PR #997, opened 2024-12-09
ci: only run test-valigrind on main

bartlomieju:


dsherret:
An issue is that this still runs an XL runner, which costs a bit of money each time I think:https://github.com/denoland/deno_core/actions/runs/12240311256/job/34142831809?pr=997

littledivy (approved):

dsherret (changes requested):
Is there a way we can get this to not run an xl runner?

bartlomieju:
@dsherret I changed it to always use a regular runner. I think it's okay if valgrind runs for an hour on `main` branch.

dsherret (approved):
LGTM

Issue #999, opened 2024-12-10
`TypeId` not reliable for object wrap templates


littledivy:
`TypeId` for a type may change across builds even with same-rustc version. `TypeId` during snapshot build can be different from the one at runtime. 

This causes `make_cppgc_object` to not use the correct FunctionTemplate when returning a wrap'ed object from an op.

PR #1000, opened 2024-12-12
Store object wrap templates in snapshot

littledivy:
OpState is not stored with the snapshot hence function templates need to be serialized with a stable-ID (I chose type_name over TypeIds)

Fixes https://github.com/denoland/deno_core/issues/999


bartlomieju (approved):
LGTM

PR #1003, opened 2024-12-16
feat: WebIDL derive macro

crowlKats:
- adds a `WebIdlConverter` trait
  - implemented for Vec, to have sequence converters
  - implemented for Option, to have nullable converters
  - implemented for HashMap, to have record converters
  - implemented for integers
  - implemented for floats, and creates `Unrestricted` structs for unrestricted float and unrestricted double
  - implemented for booleans
  - implemented for string
  - creates new `ByteString` struct for bytestring conversion
  - creates new `BigInt` struct for bigint conversion
- adds a `WebIDL` derive macro, current only allowed on structs as a dictionary converter and on enums as enum converters


core/webidl.rs
```diff
@@ -0,0 +1,154 @@
...
+        context,
+        WebIdlErrorKind::ConvertToConverterType("sequence"),
+      ));
+    };
+
+    let mut out = vec![];
+
+    let next_key = v8::String::new(scope, "next").unwrap();
+    let done_key = v8::String::new(scope, "done").unwrap();
+    let value_key = v8::String::new(scope, "value").unwrap();
```
bartlomieju, line 113: Change these to OneByteConst before landing - lookup `v8_static_string!` macro and use it here

ops/webidl/mod.rs
```diff
@@ -0,0 +1,229 @@
+use proc_macro2::Ident;
```
bartlomieju, line 1: Copyright

ops/webidl/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+          let val = if field.required {
+            quote!(val)
+          } else {
+            quote!(Some(val))
+          };
+
+          quote! {
+            let #name = {
+              if let Some(__value) = __obj.as_ref().and_then(|__obj| {
+                let __key = v8::String::new(__scope, #string_name).unwrap();
```
bartlomieju, line 90: Since this is compile time, please use OneByteConst here too

ops/webidl/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+  fn try_from(value: Field) -> Result<Self, Self::Error> {
+    let is_optional = if let Type::Path(path) = value.ty {
+      if let Some(last) = path.path.segments.last() {
+        last.ident == "Option"
+      } else {
+        false
+      }
+    } else {
+      false
+    };
```
littledivy, line 213: I don't know if this will work. We can have a field that is initialized to `undefined` but still required.

crowlKats: good point. will rethink this

crowlKats: maybe having a required attribute that only works on options might do the trick

ops/webidl/mod.rs
```diff
@@ -0,0 +1,353 @@
...
+              __value: ::deno_core::v8::Local<'a, ::deno_core::v8::Value>,
+              __prefix: std::borrow::Cow<'static, str>,
+              __context: std::borrow::Cow<'static, str>,
+              __options: &Self::Options,
+            ) -> Result<Self, ::deno_core::webidl::WebIdlError> {
+              let __obj = if __value.is_undefined() || __value.is_null() {
+                None
+              } else {
+                Some(__value.to_object(__scope).ok_or_else(|| {
+                  ::deno_core::webidl::WebIdlError::new(
```
littledivy, line 180: Shouldn't this be a `try_into` a v8::Object instead of converting to one?

ops/webidl/mod.rs
```diff
@@ -0,0 +1,353 @@
...
+                  None
+                } else {
+                  Some(__value)
+                }
+              }) {
+                let val = ::deno_core::webidl::WebIdlConverter::convert(
+                  __scope,
+                  __value,
+                  __prefix.clone(),
+                  format!("'{}' of '{}' ({__context})", #string_name, #ident_string).into(),
```
littledivy, line 150: This is expensive. This can be precomputed in the macro?

ops/webidl/mod.rs
```diff
@@ -0,0 +1,358 @@
...
+          };
+
+          let new_context = format!("'{string_name}' of '{ident_string}'");
+
+          quote! {
+            let #name = {
+              let __key = #v8_static_name
+                .v8_string(__scope)
+                .map_err(|e| ::deno_core::webidl::WebIdlError::other(__prefix.clone(), &__context, e))?
+                .into();
```
littledivy, line 136: Field name strings can be cached as `v8::Eternal` in OpState.

Pseudo-code:
```rust
struct #v8_static_name _SLOT (v8::Eternal<v8::String>);

let __key = if let Some(__eternal) = state.borrow::<#v8_static_name _SLOT>() {
  __eternal.get(__scope) // fast path
} else {
  let __key = #v8_static_name
    .v8_string(__scope)
    .map_err(|e| ::deno_core::webidl::WebIdlError::other(__prefix.clone(), &__context, e))?
    .into();

  let __eternal = v8::Eternal::empty();
  __eternal.set(__scope, __key);

  state.put(#v8_static_name _SLOT (__eternal));
  __key
};
```

littledivy (approved):
LGTM! Let's try it out in Deno and iterate based on what's needed.


cc @ry  @bartlomieju

bartlomieju (approved):
LGTM too, let's try it out.

PR #1007, opened 2024-12-17
refactor: ability to get borrowed source map bytes from loader

dsherret:
This will reduce an allocation in a refactor I'm doing for `deno compile` because we can return a static reference to the bytes.


testing/checkin/runner/ts_module_loader.rs
```diff
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
```
bartlomieju, line 1: Move a line below

bartlomieju (approved):

testing/checkin/runner/ts_module_loader.rs
dsherret: Opened https://github.com/denoland/deno_core/pull/1008

PR #1012, opened 2024-12-22
Remove intermediate OpMethodCtx

littledivy:
It is cheaper to store all OpCtx in one place and pass around OpMethodDecl and an offset that divides the OpCtx in normal ops and method ops.

### DIFF
--- a/testing/unit/resource_test.ts
+++ b/testing/unit/resource_test.ts
@@ -65,8 +65,8 @@ test(async function testCppgcAsync() {
 });
 
 test(function testDomPoint() {
-  const p1 = new DOMPoint(100, 100);
   const p2 = new DOMPoint();
+  const p1 = new DOMPoint(100, 100);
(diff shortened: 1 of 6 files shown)


testing/unit/resource_test.ts
```diff
@@ -65,8 +65,8 @@ test(async function testCppgcAsync() {
 });
 
 test(function testDomPoint() {
-  const p1 = new DOMPoint(100, 100);
   const p2 = new DOMPoint();
+  const p1 = new DOMPoint(100, 100);
```
bartlomieju, line 69: Unrelated change?

bartlomieju (approved):
LGTM, good catch

testing/unit/resource_test.ts
littledivy: Oops, will change it back in another PR

PR #1020, opened 2025-01-02
fix: apply source map to callsite objects

nathanwhit:
Ref https://github.com/denoland/deno/issues/25794.

Lazily source map the callsite if `getFileName`, `getLineNumber`, `getColumnNumber`, or `toString` is called. Cache the result on the callsite wrapper object as an array of `[fileName, lineNumber, columnNumber]`, and then use the cached value for subsequent calls.


bartlomieju (approved):
Cool, thanks!

PR #1024, opened 2025-01-03
feat(webidl): distinguish Nullable from Option, allow setting default value in webidl op2 macro attribute

crowlKats:
replaces hashmap with indexmap for records, and some minor fixes
And adds a `SameObject` utility to cppgc

### DIFF
--- a/core/webidl.rs
+++ b/core/webidl.rs
@@ -1,7 +1,7 @@
 // Copyright 2018-2025 the Deno authors. MIT license.
 
+use indexmap::IndexMap;
 use std::borrow::Cow;
-use std::collections::HashMap;
 use v8::HandleScope;
 use v8::Local;
 use v8::Value;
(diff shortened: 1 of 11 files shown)


core/webidl.rs
```diff
@@ -82,6 +82,41 @@ pub enum WebIdlErrorKind {
...
+  match value.type_of(scope).to_rust_string_lossy(scope).as_str() {
+    "undefined" => Type::Undefined,
+    "boolean" => Type::Boolean,
+    "number" => Type::Number,
+    "string" => Type::String,
+    "symbol" => Type::Symbol,
+    "bigint" => Type::BigInt,
+    // Per ES spec, typeof returns an implementation-defined value that is not any of the existing ones for
+    // uncallable non-standard exotic objects. Yet Type() which the Web IDL spec depends on returns Object for
+    // such cases. So treat the default case as an object.
```
petamoriken, line 115: nits: Per current ES spec, it no longer returns an implementation-defined value.
https://github.com/tc39/ecma262/pull/1441

crowlKats: ah interesting; i'll remove the comment then

ops/webidl/test_cases/dict.rs
```diff
@@ -13,7 +13,6 @@ pub struct Dict {
   c: Option<u32>,
   #[webidl(rename = "e")]
   d: u64,
-  f: std::collections::HashMap<String, f32>,
```
littledivy, line 16: Is IndexMap being test anywhere?

littledivy (approved):

ops/webidl/test_cases/dict.rs
crowlKats: yeah, in the `records` test in `webidl.rs`

PR #1025, opened 2025-01-03
Add text and binary module types

SpaceManiac:
Pursuant to https://github.com/denoland/deno/issues/20374 and https://github.com/denoland/deno/issues/25354.

* Add `RequestedModuleType::as_str` convenience method.
* Add `RequestedModuleType` parameter to `prepare_load` to mirror `load`.
  * This is needed to let text and binary loading skip parsing as JS syntax, which happens during `prepare_load`. Otherwise any file that looks-like JS or JSON based on file extension or `Content-Type` header can't be imported as text or binary.
* Add and implement `ModuleType::Text` and `Binary`.
  * Meant to eventually be used like:
    ```typescript
    import myText from "./deno.jsonc" with { type: "text" };
    import myBinary from "./logo.png" with { type: "binary" };
    myText satisfies string;
    myBinary satisfies ArrayBuffer;
    ```
  * [ ] **Question**: Does this belong in `deno_core`? It seems broadly useful, but maybe you want to keep core closer to the ECMAScript standard and implement these in the `deno` CLI embedding. I think this would be the first module type not to go in core if so.
  * [ ] **Question**: Is `"binary"` the right name, or do you want `"bytes"`?
  * [ ] **Question**: Is it okay that `ArrayBuffer` is mutable? If not, can we hackily make the import return something immutable, or must we block on [standardization][pia] and v8 implementation?

Fully implementing this feature will also involve changes to [deno_graph] and [deno], and a one-line change to [deno_doc].

* [ ] Might need tests.

[pia]: https://github.com/tc39/proposal-immutable-arraybuffer
[deno_graph]: https://github.com/denoland/deno_graph/compare/main...SpaceManiac:deno_graph:patch/import-text-binary
[deno]: https://github.com/denoland/deno/compare/main...SpaceManiac:deno:patch/import-text-binary
[deno_doc]: https://github.com/denoland/deno_doc/compare/main...SpaceManiac:deno_doc:patch/import-text-binary

### DIFF
--- a/core/modules/tests.rs
+++ b/core/modules/tests.rs
@@ -530,7 +530,7 @@ fn test_lazy_loaded_esm() {
 }
 
 #[test]
-fn test_json_module() {
+fn test_json_text_bytes_modules() {
   let loader = Rc::new(TestingModuleLoader::new(StaticModuleLoader::default()));
   let mut runtime = JsRuntime::new(RuntimeOptions {
     module_loader: Some(loader.clone()),
@@ -552,7 +552,7 @@ fn test_json_module() {
 
   let module_map = runtime.module_map().clone();
 
-  let (mod_b, mod_c) = {
+  let (mod_b, mod_c, mod_d, mod_e) = {
     let scope = &mut runtime.handle_scope();
     let specifier_b = ascii_str!("file:///b.js");
 
@@ -566,6 +566,13 @@ fn test_json_module() {
           import jsonData from './c.json' with {type: "json"};
           assert(jsonData.a == "b");
           assert(jsonData.c.d == 10);
+          import txt from './d.txt' with {type: "text"};
+          assert(txt == "hello there");
+          import bytes from './e.bin' with {type: "bytes"};
+          assert(bytes.length === 3);
+          assert(bytes[0] === 1);
+          assert(bytes[1] === 2);
+          assert(bytes[2] === 3); 
         "#
         ),
         false,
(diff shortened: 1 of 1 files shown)


bartlomieju:
I don't think these changes are necessary, if we decide to do it in Deno, the `prepare_module_load` hook can figure out what to do with the file. That said, there's no plan to address these two issues at the moment, but that might change later in the quarter.

SpaceManiac:
> the `prepare_module_load` hook can figure out what to do with the file.

The trouble I ran into here is that dynamic imports are handled directly by `deno_core`, call `RecursiveModuleLoad::prepare` which (without this patch) throws away the `RequestedModuleType`, and then `deno`'s `ModueLoader` impl and therefore `ModuleLoadPreparer::prepare_module_load` has no knowledge of whether a URL was imported with `type: "text"` or not. Then it goes down and tries to parse Unknown or JSON files per JavaScript syntax which of course fails on arbitrary text files, or on files which have `.json` extension (and thus media-type) but are actually JSONC.

petamoriken:
IMHO, this feature should not be included because it is still a proposal in the discussion stage in the web standards.
https://github.com/whatwg/html/issues/9444

At least, we need to make sure that there is no opposition in the standard position of the three browsers.

petamoriken:
Or, it requires the unstable flag.

core/modules/map.rs
```diff
@@ -805,6 +817,52 @@ impl ModuleMap {
...
+  }
+
+  pub(crate) fn new_binary_module(
+    &self,
+    scope: &mut v8::HandleScope,
+    name: impl IntoModuleName,
+    code: ModuleCodeBytes,
+  ) -> Result<ModuleId, ModuleError> {
+    let name = name.into_module_name();
+    let backing_store = match code {
```
bartlomieju, line 851: We can use an immutable backing store here

core/modules/map.rs
```diff
@@ -805,6 +817,56 @@ impl ModuleMap {
...
+      }
+      ModuleCodeBytes::Boxed(bytes) => {
+        v8::ArrayBuffer::new_backing_store_from_boxed_slice(bytes)
+      }
+      ModuleCodeBytes::Arc(bytes) => {
+        v8::ArrayBuffer::new_backing_store_from_vec(bytes.to_vec())
+      }
+    };
+    let source_arraybuffer =
+      v8::ArrayBuffer::with_backing_store(scope, &backing_store.make_shared());
```
bartlomieju, line 864: Probably want to use `Uint8Array` instead of `ArrayBuffer`

bartlomieju:
Need to add a few more tests before landing, and adjust the commit message

bartlomieju (approved):
Thanks for the initial work @SpaceManiac!

PR #1027, opened 2025-01-04
fix: don't call `alloc` with a length of 0

nathanwhit:
Ref https://github.com/denoland/deno/issues/27545.

 Fixes a leak in deno.serve with an empty response. We would call `to_string_ptr` with an empty string and then call alloc with a length of 0 which is UB.

Instead of using `alloc` directly, which has a large number of invariants to uphold, I've changed to just use `Vec` APIs. That leaves us with fewer invariants to uphold, handles the tricky cases for us, and we need to make a `Vec` eventually anyway (since `String` just wraps a Vec).

### DIFF
--- a/core/runtime/ops.rs
+++ b/core/runtime/ops.rs
@@ -322,6 +322,10 @@ pub fn to_cow_one_byte(
     WriteOptions::NO_NULL_TERMINATION,
   );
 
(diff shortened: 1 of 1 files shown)


littledivy (approved):
LGTM

PR #1033, opened 2025-01-06
perf: reduce generated error handling code in ops

ry:
deno release build is about 700kb smaller with this patch
```
110239496 this patch
110942920 main
```


littledivy (approved):
LGTM, can you update the tests output with `UPDATE_EXPECTED=1 CI=1 cargo test`?

PR #1037, opened 2025-01-08
fix: revert js_class_and_message_to_exception to just do TypeErrors

crowlKats:
This is the previous behaviour which was unexpected and we want to change this to actually throw the correct error, however this is breaking a few things in CLI so reverting this single change for now and will revisit at a later date


dsherret (approved):
Can you add a comment as to why this is commented out?

PR #1039, opened 2025-01-09
refactor: simplify async op error construction

crowlKats:
Removes usage of serde_v8 for constructing errors

### DIFF
--- a/core/00_infra.js
+++ b/core/00_infra.js
@@ -17,7 +17,7 @@
     Promise,
     PromiseReject,
     PromiseResolve,
-    PromisePrototypeThen,
+    PromisePrototypeCatch,
     RangeError,
     ReferenceError,
     SafeArrayIterator,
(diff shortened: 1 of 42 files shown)


core/00_infra.js
```diff
@@ -133,18 +133,22 @@
       MapPrototypeSet(promiseMap, oldPromiseId, oldPromise);
     }
 
-    const promise = new Promise((resolve) => {
-      promiseRing[idx] = resolve;
+    const promise = new Promise((resolve, reject) => {
+      promiseRing[idx] = [resolve, reject];
```
littledivy, line 137: Can you run a `op_void_async_deferred` microbenchmark? Slighty worried this will cause a regression because it's in the hot path. Maybe `{ resolve, reject }` is faster because of shape optimizations.

crowlKats: main:
```
test bench_op_async_void_deferred         ... bench:     739,054 ns/iter (+/- 77,089)
test bench_op_async_void_deferred_nofast  ... bench:     715,870 ns/iter (+/- 78,892)
test bench_op_async_void_deferred_return  ... bench:     678,101 ns/iter (+/- 38,345)
```

PR:
```
test bench_op_async_void_deferred         ... bench:     712,190 ns/iter (+/- 61,168)
test bench_op_async_void_deferred_nofast  ... bench:     742,969 ns/iter (+/- 52,853)
test bench_op_async_void_deferred_return  ... bench:     689,404 ns/iter (+/- 41,140)
```

littledivy (approved):
LGTM

core/00_infra.js
marvinhagemeister: If the array allocation is a concern we could also move two arrays of ids instead:

```ts
const promiseRing = ArrayPrototypeFill(new Array(RING_SIZE), NO_PROMISE);
const rejectRing = ArrayPrototypeFill(new Array(RING_SIZE), NO_PROMISE);

// later
const promise = new Promise((resolve, reject) => {
  promiseRing[idx] = resolve;
  promiseReject[idx] = reject;
});
```

Basically keep two rings at the same time. That should be better for memory and GC pressure.

PR #1042, opened 2025-01-10
feat: support options in error constructors

### LINKED ISSUE #1031: Public JS Error classes should accept the options argument
Currently public error classes (`BadResource`, `Interrupted`, and `NotCapable`) do not accept the 2nd argument. That is inconvenient when the downstream implementation want to use, for example, `cause` option of Error class (See https://github.com/denoland/deno/pull/27463/files#diff-847a11c353dfb7aeae6e35bc104b83a091ef797bdfc559d22ccfc804a3615000R380 for example)

kt3k:
This PR adds support of the 2nd arguments (`options`) to JS error constructors (`BadResource`, `Interrupted`, and `NotCapable`). closes #1031

By this change, downstream implementations can use `cause` option of `Error` class.

### DIFF
--- a/core/01_core.js
+++ b/core/01_core.js
@@ -301,24 +301,24 @@
   // errors in the JS code (eg. "deno_net") so they are provided in "Deno.core"
   // but later reexported on "Deno.errors"
   class BadResource extends Error {
-    constructor(msg) {
-      super(msg);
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM 👍

PR #1046, opened 2025-01-15
fix: make error properties overwritable

crowlKats:


testing/unit/ops_error_test.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright 2018-2025 the Deno authors. MIT license.
+import { assertEquals, test } from "checkin:testing";
+import { throwCustomErrorWithCode } from "checkin:error";
+
+// Test that stack traces from async ops are all sane
```
bartlomieju, line 5: This comment seems wrong

PR #1052, opened 2025-01-22
refactor: reuse code from deno_path_util

dsherret:


Cargo.lock
dsherret, line 1: I think there's a lot of changes here because the url crate got upgraded.

marvinhagemeister (approved):

Cargo.lock
littledivy: mhmm thats a lot of new stuff. what's the `icu_*` crate for?

dsherret: It’s stuff already in the cli, so probably ok. @crowlKats would know.

PR #1057, opened 2025-01-24
chore(ci): remove old release workflow

littledivy:
Not being used and replaced by `version_bump`


crowlKats (approved):
was thinking about this the other day, good to see it removed

PR #1059, opened 2025-01-24
feat(cppgc): add get and unwrap utlities to `SameObject`

petamoriken:
These utilities are useful for implementing the Geometry Interface Module Level 1 (https://github.com/denoland/deno/pull/27527).

### DIFF
--- a/core/cppgc.rs
+++ b/core/cppgc.rs
@@ -182,10 +182,12 @@ impl FunctionTemplateData {
   }
 }
 
+#[derive(Debug)]
 pub struct SameObject<T: GarbageCollected + 'static> {
   cell: std::cell::OnceCell<v8::Global<v8::Object>>,
   _phantom_data: std::marker::PhantomData<T>,
 }
+
 impl<T: GarbageCollected + 'static> SameObject<T> {
   #[allow(clippy::new_without_default)]
   pub fn new() -> Self {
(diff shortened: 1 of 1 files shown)


core/cppgc.rs
```diff
@@ -182,10 +182,12 @@ impl FunctionTemplateData {
   }
 }
 
+#[derive(Debug)]
```
crowlKats, line 185: the debug output is probably quite useless; any specific need for it?

petamoriken: The intent is that this would allow `#[derive(Debug)]` to be added to `DOMQuadInner`, but I don't have a particularly strong opinion.

crowlKats: oh i see, that makes sense!

core/cppgc.rs
```diff
@@ -212,4 +214,19 @@ impl<T: GarbageCollected + 'static> SameObject<T> {
       })
       .clone()
   }
+
+  pub fn set(
```
crowlKats, line 218: when is this useful? sameobject should always just be set once and forget, any reason `get` cannot be used in your scenario?

petamoriken: Although `get` is sufficient, I think it may be useful when we want to assign a value in advance in a constructor or static method.
https://github.com/denoland/deno/blob/dfa4af12b093ac3a504d7fe98004ad7affd1c91a/ext/geometry/lib.rs#L311-L315

crowlKats: makes sense

crowlKats (approved):
LGTM

PR #1065, opened 2025-01-29
feat(op2): fake async, fix raw fields for WebIDL converter, add more implementations for ToV8 and FromV8

crowlKats:


littledivy (approved):
LGTM

PR #1069, opened 2025-02-04
Allow String as `execute_script` name parameter

allevo:
This PR allows the String usage for `execute_script`
```rust
let mut runtime = JsRuntime::new(RuntimeOptions::default());

let id = 1;
let script_name = format!("script-{}", id);
let script_content = format!("console.log('{}')", id);
let result = runtime.execute_script(script_name, script_content).unwrap();
```


bartlomieju (approved):
Thanks

PR #1071, opened 2025-02-10
feat: optional code cache for extension sources

nathanwhit:
Wires up the code cache through the `ExtModuleLoader` and extension initialization, so that embedders can use the V8 code cache on extension sources.

### DIFF
--- a/core/runtime/jsrealm.rs
+++ b/core/runtime/jsrealm.rs
@@ -3,6 +3,8 @@
 use super::exception_state::ExceptionState;
 #[cfg(test)]
 use super::op_driver::OpDriver;
+use crate::ModuleSourceCode;
+use crate::SourceCodeCacheInfo;
 use crate::_ops::OpMethodDecl;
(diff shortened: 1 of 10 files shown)


core/runtime/jsrealm.rs
```diff
@@ -352,6 +354,91 @@ impl JsRealm {
...
+    &self,
+    isolate: &mut v8::Isolate,
+    name: ModuleSpecifier,
+    source_code: impl IntoModuleCodeString,
+    get_cache: &dyn Fn(
+      &ModuleSpecifier,
+      &ModuleSourceCode,
+    ) -> SourceCodeCacheInfo,
+    cache_ready: &dyn Fn(ModuleSpecifier, u64, &[u8]),
+  ) -> Result<v8::Global<v8::Value>, CoreError> {
```
bartlomieju, line 367: Maybe add a TODO that this should be deduplicated with other methods?

bartlomieju (approved):

PR #1094, opened 2025-03-16
Reduce dependency on `futures` crate

paolobarbolini:
Over time many `futures` APIs have been stabilized in `std`. This migrates most uses to the std variants.


ry (approved):
LGTM - Thanks!

PR #1098, opened 2025-03-19
chore: simplify formatting of rust code, move to edition 2024

ry:
* don't use dprint for formatting rust code - we get inconsistent behavior between `cargo fmt` and `tools/format.ts`
* this seems to all have been introduced to support the nightly-only `imports_granularity=item` - let's just remove that formatting option
* use edition 2024 in `.rustfmt.toml`


dsherret (dismissed):
Going to block for a bit so I can investigate why we're not using dprint here. I never run tools/format.js because it happens automatically in my editor, but this pr would break that.

dsherret (approved):
LGTM

PR #1104, opened 2025-03-25
Ref'ing read and writes on resource handles

littledivy:
Adds a `unrefed_resources` hashset similar to `unrefed_ops` and `#[op2(async, promise_id)]` for `op_read`, `op_read_all`, `op_write`. 

This allows ref'ing and unrefing using resource handles instead of inflight promises, similar to `uv_ref` and `uv_unref` in libuv. Useful for implementing `HandleWrap` in Deno.

### DIFF
--- a/core/ops.rs
+++ b/core/ops.rs
@@ -242,6 +242,7 @@ pub struct OpState {
   pub feature_checker: Arc<FeatureChecker>,
   pub external_ops_tracker: ExternalOpsTracker,
   pub op_stack_trace_callback: Option<OpStackTraceCallback>,
+  /// Reference to the unrefered ops state in `ContextState`.
   pub(crate) unrefed_ops: UnrefedOps,
   /// Resources that are not referenced by the event loop. All async
   /// resource ops on these resources will not keep the event loop alive.
@@ -275,6 +276,7 @@ impl OpState {
     std::mem::take(&mut self.resource_table);
   }
 
+  // Silly but improves readability.
   pub fn uv_unref(&mut self, resource_id: ResourceId) {
     self.unrefed_resources.insert(resource_id);
   }
--- a/core/ops.rs
+++ b/core/ops.rs
@@ -2,6 +2,7 @@
 
 use crate::FeatureChecker;
 use crate::OpDecl;
+use crate::ResourceId;
 use crate::error::JsStackFrame;
 use crate::gotham_state::GothamState;
 use crate::io::ResourceTable;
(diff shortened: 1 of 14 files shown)


core/ops.rs
```diff
@@ -239,6 +242,8 @@ pub struct OpState {
   pub feature_checker: Arc<FeatureChecker>,
   pub external_ops_tracker: ExternalOpsTracker,
   pub op_stack_trace_callback: Option<OpStackTraceCallback>,
+  pub(crate) unrefed_ops: UnrefedOps,
+  pub unrefed_resources: HashSet<ResourceId>,
```
ry, line 246: should this be exposed instead as something like `uv_unref()` ?

https://github.com/nodejs/node/blob/main/src/handle_wrap.cc#L54

core/ops.rs
```diff
@@ -263,6 +274,18 @@ impl OpState {
...
+    self.unrefed_resources.insert(resource_id);
+  }
+
+  pub fn uv_ref(&mut self, resource_id: ResourceId) {
+    self.unrefed_resources.remove(&resource_id);
+  }
+
+  pub fn has_ref(&self, resource_id: ResourceId) -> bool {
+    !self.unrefed_resources.contains(&resource_id)
+  }
```
ry, line 288: It's silly, but I think this is much better for readability later on.

ops/op2/dispatch_async.rs
```diff
@@ -60,8 +60,10 @@ pub(crate) fn generate_dispatch_async(
     quote!()
   };
 
+  let input_index = if config.promise_id { 0 } else { 1 };
   // input_index = 1 as promise ID is the first arg
-  let args = generate_dispatch_slow_call(generator_state, signature, 1)?;
+  let args =
+    generate_dispatch_slow_call(generator_state, signature, input_index)?;
```
ry, line 66: update this comment and add a bit of context here...

core/ops_builtin.rs
```diff
@@ -277,32 +277,45 @@ pub fn op_wasm_streaming_set_url(
   Ok(())
 }
 
-#[op2(async)]
+fn get_resource(
+  state: Rc<RefCell<OpState>>,
+  rid: ResourceId,
+  promise_id: i32,
+) -> Result<Rc<dyn Resource>, JsErrorBox> {
```
ry, line 284: add triple slash comment - give context

ry (commented):
There should be a test of the `#[op2(async, promise_id)]` - I find it difficult to follow how it works

littledivy:
@ry PTAL

ry (approved):
LGTM

PR #1121, opened 2025-04-26
Defer scope borrow to avoid borrow conflicts with arguments

### LINKED ISSUE #903: Op that takes handle scope and a cppgc argument doesn't compile with `#[op2(fast)]`
Repro

```
pub struct Foo;
impl deno_core::GarbageCollected for Foo {}
#[op2(fast)]
pub fn op_repro(_scope: &mut v8::HandleScope, #[cppgc] _foo: &Foo) {}
```

Fails to compile with

```
error[E0502]: cannot borrow `*fast_api_callback_options.isolate` as mutable because it is also borrowed as immutable
   --> repro/main.rs:3:1
    |
3 | #[op2(fast)]
    | ^^^^^^^^^^^-
    | |          |
    | |          immutable borrow might be used here, when `scope` is dropped and runs the `Drop` code for type `CallbackScope`
    | mutable borrow occurs here
    | immutable borrow occurs here
    |
    = note: this error originates in the attribute macro `op2` (in Nightly builds, run with -Z macro-backtrace for more info)
```

If you remove the fast attr, then `op2` errors with `This op is fast-compatible and should be marked as (fast)`.

The only way to get it to compile is with the `nofast` attr.

littledivy:
Fixes https://github.com/denoland/deno_core/issues/903

### DIFF
--- a/ops/op2/dispatch_fast.rs
+++ b/ops/op2/dispatch_fast.rs
@@ -102,24 +102,35 @@ impl FastSignature {
     for arg in &self.args {
       match arg {
         FastArg::Actual { arg, name_out, .. }
-        | FastArg::Virtual { name_out, arg } => call_args.push(
-          map_v8_fastcall_arg_to_arg(generator_state, name_out, arg).map_err(
-            |s| V8SignatureMappingError::NoArgMapping(s, arg.clone()),
-          )?,
-        ),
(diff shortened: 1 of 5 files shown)


lucacasonato (approved):
LGTM

bartlomieju (approved):
LGTM too

PR #1133, opened 2025-05-14
feat: add `additional_properties` to `JsError`

crowlKats:
currently js errors that are printed via console.log in cli display their additional properties, however the same does not apply for when displayed from rust. this allows for `fmt_errors.rs` in cli to change that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -667,9 +667,9 @@ dependencies = [
 
 [[package]]
 name = "deno_error"
-version = "0.5.6"
+version = "0.5.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


littledivy (approved):
LGTM

stefnotch:
Does this already populate the `additional_properties` of an error, or is https://github.com/denoland/deno_core/blob/971b775a7cd7a60d9e59712a6106237d76160b27/core/error.rs#L242, which is called by https://github.com/denoland/deno_core/blob/971b775a7cd7a60d9e59712a6106237d76160b27/core/error.rs#L299C6-L299C31 still missing?

Am asking, since I was interested in extracting an "expected" and "actual" value from the error object in the Deno LSP for that nice VSCode diff view.

PR #1135, opened 2025-05-19
fix: `#[undefined]` attribute

littledivy:
Apparently `assertEquals` is not strict enough, replaced with a normal `assert`


dsherret:
Can you update the title to use `fix:`? We should be doing that on all repos to communicate intent. It's easy to tell for this one, but not so much for some of the other commits that have landed on main.

dsherret (approved):

PR #1137, opened 2025-05-22
feat(op2): add `#[validate(fn)]` for custom arg validation

littledivy:
Node.js APIs written in object wrap need to throw errors like `ERR_INVALID_ARG_TYPE` and `ERR_INVALID_CONSTRUCTOR` before argument de-serialization which is not directly exposed in op2. This is currently not possible.

This PR adds custom validation hooks that run before de-serializing each argument using `#[validate(fn)]` attribute.

```rust
fn int(
  _scope: &mut v8::HandleScope,
  value: v8::Local<v8::Value>,
) -> Result<(), JsErrorBox> {
  if value.is_int32() {
    return Ok(());
  }

  Err(JsErrorBox::type_error("Expected int"))
}

#[op2]
impl Object {
  #[fast]
  fn n(
    &self,
    #[validate(int)]
    #[smi]
    n: u32,
  ) {
    // ...
  }
}
```

### DIFF
--- a/testing/unit/resource_test.ts
+++ b/testing/unit/resource_test.ts
@@ -1,5 +1,11 @@
 // Copyright 2018-2025 the Deno authors. MIT license.
-import { assert, assertArrayEquals, assertEquals, test } from "checkin:testing";
+import {
+  assert,
+  assertArrayEquals,
+  assertEquals,
+  assertThrows,
+  test,
+} from "checkin:testing";
 import {
   DOMPoint,
   DOMPointReadOnly,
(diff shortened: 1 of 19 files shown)


testing/unit/resource_test.ts
```diff
@@ -136,6 +136,15 @@ test(async function testDomPoint() {
...
   assert(wrap.undefinedResult() === undefined);
 
+  wrap.withValidateInt(10);
+
+  try {
+    wrap.withValidateInt("bad");
+  } catch (e) {
+    assert(e instanceof TypeError);
+    assertEquals(e.message, "Expected int");
+  }
```
bartlomieju, line 146: I think we need `assertThrows()` helper? If the function doesn't throw at all this test will still pass

bartlomieju (approved):
LGTM

PR #1138, opened 2025-05-28
feat: support numerical additional error properties

littledivy:
Ref https://github.com/denoland/deno_error/pull/14

### DIFF
--- a/core/runtime/setup.rs
+++ b/core/runtime/setup.rs
@@ -29,6 +29,7 @@ fn v8_init(
     " --turbo_fast_api_calls",
     " --harmony-temporal",
     " --js-float16array",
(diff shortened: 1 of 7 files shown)


core/runtime/setup.rs
```diff
@@ -29,6 +29,7 @@ fn v8_init(
     " --turbo_fast_api_calls",
     " --harmony-temporal",
     " --js-float16array",
+    " --js-explicit-resource-management",
```
littledivy, line 32: Enabled by default because `deno_ast` no longer transpiles `using`

bartlomieju (approved):

crowlKats (approved):

PR #1147, opened 2025-06-13
feat(BREAKING): move import meta override to `ModuleLoader` trait

dsherret:
Makes things a lot easier in the CLI because I can control import meta resolution at the same place as regular resolution (note: enum is a private type in the cli code).

image


marvinhagemeister (approved):
LGTM

PR #1150, opened 2025-06-14
cppgc: Support the new APIs from rusty_v8

simonask:
This implements the necessary changes to support the soundness fixes from https://github.com/denoland/rusty_v8/pull/1802.

I'm creating this as a draft for now because (a) it currently can't build until there is a release of the v8 crate with those changes, and (b) there are designs that probably need discussion.

@devsnek


simonask:
This PR should incorporate the changes in https://github.com/denoland/rusty_v8/pull/1807 (`GcCell`) when/if that is accepted, removing the uses of `UnsafeCell`.

PR #1153, opened 2025-06-24
fix(core): prevent panic when Array.prototype[0] has getter-only property

krugerLucas:
## Summary

This PR fixes a critical panic that occurs when `Array.prototype[0]` is overridden with a getter-only property and certain Deno APIs that create V8 arrays are called.

## Problem

When `Array.prototype[0]` is defined with a getter-only property descriptor, calling `v8::Array::new_with_elements()` causes a panic with the error:
```
Custom error class must have a builder registered: Uncaught TypeError: Cannot set property 0 of [object Array] which has only a getter
```

This affects multiple Deno APIs including `Deno.readTextFile()` and any operation that triggers error creation with stack traces.

## Solution

Replace all instances of `v8::Array::new_with_elements()` with the pattern:
```rust
let arr = v8::Array::new(scope, length);
for (i, item) in items.iter().enumerate() {
    arr.set_index(scope, i as u32, *item);
}
```

This approach avoids the internal V8 mechanism that triggers the prototype chain lookup and subsequent panic.

## Changes

1. **core/error.rs** - Fixed 4 locations:
   - Line ~308: `to_v8_error` for individual properties
   - Line ~318: `to_v8_error` for array of properties  
   - Line ~1383: `source_mapped_call_site_info` (already using correct pattern)
   - Line ~1653: `prepare_stack_trace_callback` for patched callsites

2. **core/runtime/tests/error.rs** - Added test:
   - `array_prototype_pollution_no_panic()`: Verifies the fix prevents panics

## Testing

The new unit test sets up Array.prototype pollution and verifies that error handling continues to work without panicking:
```rust
#[test]
fn array_prototype_pollution_no_panic() {
    // Test setup with getter-only Array.prototype[0]
    // Verify no panic occurs when creating errors
}
```

## Related Issues

Fixes denoland/deno#29824

## Impact

This fix prevents runtime crashes in production applications that may have Array.prototype pollution either intentionally (for debugging/monitoring) or due to third-party code.

### DIFF
--- a/core/error.rs
+++ b/core/error.rs
@@ -305,14 +305,21 @@ pub fn to_v8_error<'a>(
         PropertyValue::Number(value) => v8::Number::new(tc_scope, value).into(),
       };
 
-      v8::Array::new_with_elements(tc_scope, &[key, value]).into()
+      // Use direct array creation with set_index to avoid prototype pollution issues
+      let arr = v8::Array::new(tc_scope, 2);
+      arr.set_index(tc_scope, 0, key);
+      arr.set_index(tc_scope, 1, value);
+      arr.into()
     })
     .collect::<Vec<_>>();
 
   if !additional_properties.is_empty() {
-    args.push(
-      v8::Array::new_with_elements(tc_scope, &additional_properties).into(),
-    );
+    // Create array using new() and set_index to avoid prototype pollution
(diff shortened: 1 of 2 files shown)


petamoriken:
I reported it to V8.
https://issues.chromium.org/issues/427341635

PR #1156, opened 2025-07-04
fix(flags): fix "--turbo-fast-api-calls" v8 flags

linrongbin16:
The "--turbo_fast_api_calls" v8 flag should be "--turbo-fast-api-calls", see here: 

https://github.com/chromium/chromium/blob/8589ffbc2b20e6510858392492a0f893b58a0366/gin/v8_initializer.cc#L379

Please correct me if I miss anything.

### DIFF
--- a/core/runtime/setup.rs
+++ b/core/runtime/setup.rs
@@ -26,7 +26,7 @@ fn v8_init(
   let base_flags = concat!(
     " --wasm-test-streaming",
     " --no-validate-asm",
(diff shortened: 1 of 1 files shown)


bartlomieju:
> The "--turbo_fast_api_calls" v8 flag should be "--turbo-fast-api-calls", see here:

V8 accepts flag using both underscores (`_`) as well as dashes (`-`) so it's not really required to change anything here

PR #1158, opened 2025-07-11
perf: optimize event_loop_tick and async ops

CyanChanges:
## Superseded by 
- [ ] https://github.com/denoland/deno_core/pull/1163
- [ ] https://github.com/denoland/deno_core/pull/1164

## Summary

34% more performant async ops + less memory usage
(by resolve `Promise`s natively)

```js
const ops = Deno.core.ops;
// ops.op_print(Object.keys(ops).join('\n')+'\n')
function tImm(tag, p) {
  const symbol = Symbol('k.notImm')
  function check(v) {
    const imm = v !== symbol
    // ops.op_print(`@${tag} ${imm}\n`)
  }
  const p1 = Promise.any([p, Promise.resolve(symbol)])
  p1.then(check, check)
}
for (let i = 0; i < 50000; i++) {
tImm("op_void_async", ops.op_void_async())
tImm("op_void_async_deferred", ops.op_void_async_deferred())
tImm("op_error_async", ops.op_error_async())
}
// ops.op_print("Hello World\n")
```

image

### DIFF
--- a/core/ops_builtin_v8.rs
+++ b/core/ops_builtin_v8.rs
@@ -13,6 +13,7 @@ use crate::ops_builtin::WasmStreamingResource;
 use crate::resolve_url;
 use crate::runtime::JsRealm;
 use crate::runtime::JsRuntimeState;
+use crate::runtime::op_driver::OpDriver;
 use crate::source_map::SourceMapApplication;
 use crate::stats::RuntimeActivityType;
 use deno_error::JsErrorBox;
@@ -23,7 +24,6 @@ use std::fmt::Write;
 use std::rc::Rc;
 use v8::ValueDeserializerHelper;
 use v8::ValueSerializerHelper;
-use crate::runtime::op_driver::OpDriver;
 
 #[op2]
 pub fn op_add_main_module_handler(
@@ -47,21 +47,33 @@ pub fn op_set_handled_promise_rejection_handler(
 }
 
 #[op2(fast)]
-pub fn op_ref_op_promise(scope: &mut v8::HandleScope, promise: v8::Local<v8::Promise>) {
+pub fn op_ref_op_promise(
+  scope: &mut v8::HandleScope,
+  promise: v8::Local<v8::Promise>,
+) {
   let context_state = JsRealm::state_from_scope(scope);
-  let promise_id = match context_state.pending_ops.promise_id_from_promise(scope, promise) {
+  let promise_id = match context_state
+    .pending_ops
+    .promise_id_from_promise(scope, promise)
+  {
     Some(promise_id) => promise_id,
-    None => return
+    None => return,
   };
   context_state.unrefed_ops.borrow_mut().remove(&promise_id);
 }
 
 #[op2(fast)]
-pub fn op_unref_op_promise(scope: &mut v8::HandleScope, promise: v8::Local<v8::Promise>) {
+pub fn op_unref_op_promise(
+  scope: &mut v8::HandleScope,
+  promise: v8::Local<v8::Promise>,
+) {
   let context_state = JsRealm::state_from_scope(scope);
-  let promise_id = match context_state.pending_ops.promise_id_from_promise(scope, promise) {
+  let promise_id = match context_state
+    .pending_ops
+    .promise_id_from_promise(scope, promise)
+  {
     Some(promise_id) => promise_id,
-    None => return
+    None => return,
   };
   context_state.unrefed_ops.borrow_mut().insert(promise_id);
 }
@@ -83,9 +95,14 @@ pub fn op_unref_op(scope: &mut v8::HandleScope, promise_id: i32) {
 }
 
 #[op2]
-pub fn op_promise_promise_id(scope: &mut v8::HandleScope, promise: v8::Local<v8::Promise>) -> Option<i32> {
+pub fn op_promise_promise_id(
+  scope: &mut v8::HandleScope,
+  promise: v8::Local<v8::Promise>,
+) -> Option<i32> {
   let context_state = JsRealm::state_from_scope(scope);
-  context_state.pending_ops.promise_id_from_promise(scope, promise)
+  context_state
+    .pending_ops
+    .promise_id_from_promise(scope, promise)
 }
 
 #[op2(fast)]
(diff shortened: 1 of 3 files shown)


CyanChanges:
Timer in deno is way slower than Node.js and Bun. (the result is better in dcore with that dcore ported version above tho)

image(https://canary.discord.com/channels/684898665143206084/778060818046386176/1393869232562765886)

I added traces for time consumed for **queue timers**, **poll timers**, and **call timers**,
both poll and call is actually very fast,
and seems the bottle-neck is at **queue timers** 

image 

<sup>
time it took each `op_queue_timer` (debug build, tracked inside of the op)
</sup>

(https://canary.discord.com/channels/684898665143206084/688040863313428503/1393599894022783038)

If I just took the worst case, and multiplied by the timer count, and it's a value about the same to the actual result. 

It needs two inserts to the BTree, and it's probably the reason.

Take a look at the implementation of Node.js 
using a map for `timeout` value, and each value is a linked list for a list of timers
(https://github.com/nodejs/node/blob/main/lib/internal/timers.js)
image
It's a constant time insert `O(1)`

Take a look at Bun's, it's using a [Pairing Heap](https://en.wikipedia.org/wiki/Pairing_heap) <sup>[[1]](https://github.com/oven-sh/bun/blob/d6d7251352647335ef05cea4dbe377023d8a91cc/src/io/heap.zig#L4)</sup>, it's 
https://github.com/oven-sh/bun/blob/main/src/bun.js/api/Timer.zig
The time complexity is `O(1) Amortized` (according to AI tho)

image


While for BTree, it's `O(log(N))` for most cases
image

core/runtime/jsrealm.rs
```diff
@@ -57,13 +57,18 @@ impl Hasher for IdentityHasher {
...
+  pub(crate) event_loop_tick_cb: v8::Global<v8::Function>,
+}
 
 pub struct ContextState {
   pub(crate) task_spawner_factory: Arc<V8TaskSpawnerFactory>,
-  pub(crate) timers: WebTimers<(v8::Global<v8::Function>, u32)>,
-  pub(crate) js_event_loop_tick_cb: RefCell<Option<v8::Global<v8::Function>>>,
+  pub(crate) timers: WebTimers<v8::Global<v8::Function>>,
   pub(crate) js_wasm_streaming_cb: RefCell<Option<v8::Global<v8::Function>>>,
+  pub(crate) js_bindings: RefCell<Option<JsBindings>>,
```
bartlomieju, line 71: Could you actually store these callbacks/bindings as separate functions instead of additional struct?

core/runtime/jsrealm.rs
```diff
@@ -202,7 +207,6 @@ impl JsRealmInner {
     let isolate = unsafe { raw_ptr.as_mut().unwrap() };
     // These globals will prevent snapshots from completing, take them
     state.exception_state.prepare_to_destroy();
-    std::mem::take(&mut *state.js_event_loop_tick_cb.borrow_mut());
```
bartlomieju, line 205: This should be kept - if you don't drop these globals before snapshotting, it will most likely error out

ops/op2/dispatch_async.rs
```diff
@@ -17,10 +15,185 @@ use super::dispatch_slow::with_self;
...
+use super::signature::{Arg, ArgMarker, ArgSlowRetval, ParsedSignature};
 use proc_macro2::TokenStream;
-use quote::quote;
+use quote::{format_ident, quote};
+
+pub fn resolve_value_infallible(
+  generator_state: &mut GeneratorState,
+  ret_type: &Arg,
+) -> Result<TokenStream, V8MappingError> {
+  // In the future we may be able to make this false for void again
```
bartlomieju, line 27: Can you explain why changes to the `op2` macro are needed?

bartlomieju (commented):
Thanks, lots of interesting stuff in here. I'd like to ask you to split this PR into smaller chunks though - it's quite hard to review (and merge) with confidence with so many moving pieces. I think it should be split into 3 distinct PRs in following order:
- changes to handling of promises in ops
- changes to `op2` macro (once I understand why they are needed anyway)
- changes to timers implementation

Let me know if you need any help with that

core/runtime/jsrealm.rs
CyanChanges: I thought put them together will be a good option, or I have to wrap them in `RefCell<Option<...>>` individually, i don't think it's necessary to have many these wrappers

ops/op2/dispatch_async.rs
CyanChanges: **TL;DR** Since I can just `resolve` and `reject` natively, I changed the macros to `resolve` `reject` the promise instead of `scope.throw` or `retval.set`. Avoid the cost of `throw`ing errors.

Oringinally, dispatch async uses the same code as dispatch slow, if it's `Err` immediately, it throw errors directly to scope, while in async, the errors should result in `Promise.reject`.

So In the async wrap, it do `try { ... } catch (err) { return Promise.reject(err) }`.

Now I removed the async wrap, the error will be thrown directly, that might cause problem.
Since I can just `resolve` and `reject` natively, I changed the macros to `resolve` `reject` the promise instead of `scope.throw` or `retval.set`

CyanChanges:
> Thanks, lots of interesting stuff in here. I'd like to ask you to split this PR into smaller chunks though - it's quite hard to review (and merge) with confidence with so many moving pieces. I think it should be split into 3 distinct PRs in following order:
> 
> * changes to handling of promises in ops
> * changes to `op2` macro (once I understand why they are needed anyway)
> * changes to timers implementation
> 
> Let me know if you need any help with that

These two needs to be done together tho:

- changes to handling of promises in ops
- changes to op2 macro (once I understand why they are needed anyway)

since I needs needs resolve promise natively, currently the op2 macro needs the async wrap to behave correctly.

This is still WIP, I guess I would use a Pairing Heap or something and check if there's performance improvements:

- changes to timers implementation

core/runtime/jsrealm.rs
CyanChanges: Oops, I forgotten

bartlomieju:
> since I needs needs resolve promise natively, currently the op2 macro needs the async wrap to behave correctly.

Yeah, that's fine. 2 PRs then 👍

core/runtime/jsrealm.rs
bartlomieju: I don't think this collects all callbacks into the struct, leading to bifurcation. Let's keep them separate and once these changes are done we can refactor all callback handling into a single struct.

CyanChanges:
Superseded by 
- [ ] https://github.com/denoland/deno_core/pull/1163
- [ ] https://github.com/denoland/deno_core/pull/1164

PR #1167, opened 2025-07-23
feat: Add `op_structured_cloned`

bartlomieju:
Specialized version of `op_serialize` and `op_deserialize` that
can be used to implement `structuredClone` API in Deno (only
the codepath that doesn't use `options` bag).


littledivy (approved):
LGTM!

PR #1168, opened 2025-07-25
chore: update rusty v8 to 139.0.0

nathanwhit:


testing/checkin/runner/ops.rs
```diff
@@ -74,10 +74,12 @@ pub fn op_stats_delete(
 
 pub struct TestObjectWrap {}
 
-impl GarbageCollected for TestObjectWrap {
+unsafe impl GarbageCollected for TestObjectWrap {
```
littledivy, line 77: I don't like that we'll have to do this everywhere in Deno 😓

littledivy (approved):
LGTM

testing/checkin/runner/ops.rs
littledivy: oh that will break a lot of stuff. i think we might have to revert this change.

bartlomieju: I'm in favor of reverting this change

PR #1174, opened 2025-08-06
feat: support immediates

bartlomieju:
This commit adds support for immediates - they are run after "next ticks" but before "macrotasks"


littledivy (commented):
Do we need a test?

core/ops_builtin_v8.rs
```diff
@@ -231,6 +219,54 @@ pub fn op_set_has_tick_scheduled(scope: &mut v8::PinScope, v: bool) {
...
+
+#[op2(fast)]
+pub fn op_immediate_set_has_outstanding(
+  scope: &mut v8::PinScope,
+  has_outstanding: bool,
+) {
+  JsRealm::state_from_scope(scope)
+    .immediate_info
+    .borrow_mut()
+    .has_outstanding = has_outstanding;
```
dsherret: I wonder if it would be faster to use a Cell for all these instead of RefCell

dsherret (approved):
LGTM

PR #1179, opened 2025-08-13
feat: expose dispatch_exception function

petamoriken:
Currently, the function to terminate a process when an exception occurs can only be accessed from JS (ops). I would like to make it accessible from Rust as well for https://github.com/denoland/deno/pull/30128.

### DIFF
--- a/core/error.rs
+++ b/core/error.rs
@@ -314,6 +314,9 @@ pub fn to_v8_error<'a>(
   }
 }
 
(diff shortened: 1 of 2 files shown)


core/error.rs
```diff
@@ -314,6 +314,25 @@ pub fn to_v8_error<'a>(
   }
 }
 
+pub fn dispatch_exception(
```
bartlomieju, line 317: Can you add a docstring to this function?

PR #1192, opened 2025-09-11
refactor: rewrite inspector implementation

bartlomieju:
This is meant to replace `LocalInspectorSession` that abstract interaction into async Rust.

PR #1195, opened 2025-09-16
fix(inspector): avoid double RefCell borrow

bartlomieju:
Fixing some of the fall out from https://github.com/denoland/deno_core/pull/1193.

Testing in CLI it turned out that there's a double RefCell borrow panic occuring,
when running the REPL - when the exception was raised when evaluating an expression -
and said expression was evaluated as part of dispatching an inspector message
- the `op_dispatch_exception` op was called that tries to mutable borrow the
inspector instance.

This is now avoided by storing `SessionContainer` in the `LocalInspectorSession`,
instead of `JsRuntimeInspector`.

This will be further simplified in follow up PRs like https://github.com/denoland/deno_core/pull/1192.


nathanwhit (approved):
LGTM

PR #1207, opened 2025-09-24
refactor: modernize InspectorServer implementation

bartlomieju:
Pulled out from https://github.com/denoland/deno_core/pull/1198 this commit
modernizes InspectorServer implementation - it now relies more on `tokio`
than `futures` crate.


littledivy (approved):
LGTM

PR #1211, opened 2025-10-06
feat: (optionally) format stack traces with relative paths

nathanwhit:
Before:

```
Error: Error
    at <absolute_path_to_cwd>/foo.ts:1:7
```

After:

```
Error: Error
    at ./foo.ts:1:7
```

The way this works is that for errors originating from the runtime, you can opt into this behavior by putting an `deno_core::error::InitialCwd` into `OpState`, which will then get picked up and used by the formatting logic.

If you're calling the formatting functions explicitly, then you just pass the initial cwd as a parameter `format_frame(frame, Some(initial_cwd))`

### DIFF
--- a/core/error.rs
+++ b/core/error.rs
@@ -1154,9 +1154,7 @@ fn abbrev_file_name(
   let Ok(url) = Url::parse(file_name) else {
     return None;
   };
-  if url.scheme() == "file"
-    && let Some(initial_cwd) = maybe_initial_cwd
-  {
+  if let Some(initial_cwd) = maybe_initial_cwd {
     return Some(relative_specifier(initial_cwd, &url)?);
   }
   if url.scheme() != "data" {
@@ -1790,6 +1788,7 @@ pub fn format_stack_trace<'s, 'i>(
     .try_borrow::<InitialCwd>()
     .map(|i| &i.0)
     .cloned();
+  eprintln!("maybe_initial_cwd: {:?}", maybe_initial_cwd);
   let mut result = String::new();
 
   if let Ok(obj) = error.try_cast() {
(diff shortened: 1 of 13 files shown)


core/error.rs
```diff
@@ -1106,8 +1106,38 @@ pub(crate) fn has_call_site<'s, 'i>(
...
+
+pub fn relative_specifier(from: &Url, to: &Url) -> Option<String> {
+  let is_dir = to.path().ends_with('/');
+
+  if is_dir && from == to {
+    return Some("./".to_string());
+  }
+
+  // workaround for url crate not adding a trailing slash for a directory
+  // it seems to be fixed once a version greater than 2.2.2 is released
```
bartlomieju, line 1125: Is this pending a release from a `url` crate then?

core/error.rs
```diff
@@ -1736,11 +1786,20 @@ pub fn prepare_stack_trace_callback<'s, 'i>(
   prepare_stack_trace_inner::<true>(scope, error, callsites)
 }
 
+pub struct InitialCwd(pub Url);
```
bartlomieju, line 1789: Please wrap `Url` in an `Rc` - it should make clones a lot cheaper

core/ops_builtin.rs
```diff
@@ -436,12 +434,6 @@ async fn op_shutdown(
   resource.shutdown().await
 }
 
-#[op2]
-#[string]
-fn op_format_file_name(#[string] file_name: &str) -> String {
-  format_file_name(file_name)
-}
```
bartlomieju, line 443: 👍

testing/integration/error_async_stack/error_async_stack.out
bartlomieju, line 1: Great change 👍

bartlomieju (approved):

core/error.rs
nathanwhit: Ah copied it from the CLI, I'm pretty sure it's already on a newer version. I'll remove that stuff

PR #1218, opened 2025-10-31
fix: disallow js execution for op_import_sync

dsherret:


dsherret:
Closing because I don't remember what this is for and it doesn't work.

PR #1228, opened 2025-11-04
feat(BREAKING): remove support for `op2` `#[state]` attribute

bartlomieju:
Besides usage in `deno_core` this attribute was not used anywhere else.

Given how complex the implementation is, after internal discussions,
we've decided that it's better for maintainability to remove it.


flightofthefox:
> Besides usage in `deno_core` this attribute was not used anywhere else.

Is there a recommended alternative for those who _are_ using it?

bartlomieju:
> > Besides usage in `deno_core` this attribute was not used anywhere else.
> 
> Is there a recommended alternative for those who _are_ using it?

```rust
#[op2]
fn op_foo(#[state] foo: &mut Foo) {
  // ...
}
```
becomes
```rust
#[op2]
fn op_foo(state: &mut OpState) {
  let mut foo = state.borrow_mut::<Foo>();
  // ...
}
```

PR #1229, opened 2025-11-04
implement ToV8 and FromV8 for Vec and tuple

nathanwhit:
For typed array types like `Vec<u8>`, `Vec<u16>` converts it to/from typed arrays.

For arbitrary `Vec<T>` where `T: ToV8` it just converts to an array of `T::to_v8`

needs test

### DIFF
--- a/core/convert.rs
+++ b/core/convert.rs
@@ -571,7 +571,7 @@ macro_rules! typedarray_to_v8 {
         let backing_shared = backing.make_shared();
         let ab = v8::ArrayBuffer::with_backing_store(scope, &backing_shared);
         v8::$v8ty::new(scope, ab, 0, len)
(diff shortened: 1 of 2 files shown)


core/convert.rs
```diff
@@ -120,6 +120,42 @@ pub trait FromV8<'a>: Sized {
...
+#[derive(Debug, thiserror::Error, deno_error::JsError, Clone, Copy)]
+#[class(type)]
+#[error(transparent)]
+pub struct DataError(pub v8::DataError);
+
+impl From<v8::DataError> for DataError {
+  fn from(err: v8::DataError) -> Self {
+    DataError(err)
+  }
+}
```
crowlKats: duplicate with `error.rs`

PR #1237, opened 2025-11-07
feat: add ToV8 and FromV8 macros, and implement traits on various number types

crowlKats:
Both macros only supports structs currently, not enums.

The ToV8 macro supports the following otions on individual fields via a `to_v8` attribute:
- rename: renames the value in js
- serde: uses serde_v8 instead of ToV8 trait

The FromV8 macro supports the following otions on individual fields via a `from_v8` attribute:
- rename: renames the value in js
- serde: uses serde_v8 instead of ToV8 trait
- default: uses the provided default value instead of erroring out when the value is missing

Additionally, both macros support a `v8` attribute, that supports the common options between the two macros.

### DIFF
--- a/ops/op2/test_cases/sync/from_v8.rs
+++ b/ops/op2/test_cases/sync/from_v8.rs
@@ -2,12 +2,12 @@
 
 #![deny(warnings)]
 deno_ops_compile_test_runner::prelude!();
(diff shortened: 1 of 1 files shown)


littledivy (approved):
LGTM

PR #1243, opened 2025-11-14
perf: fast wasm streaming

littledivy:


core/ops_builtin.rs
```diff
@@ -278,6 +279,46 @@ pub fn op_wasm_streaming_set_url(
...
+    .get::<WasmStreamingResource>(rid)
+    .map_err(|_| JsErrorBox::type_error("stream not found"))?;
+
+  loop {
+    let resource = state
+      .borrow()
+      .resource_table
+      .get_any(stream_rid)
+      .map_err(|_| JsErrorBox::type_error("stream not found"))?;
+    let view = deno_core::BufMutView::new(65536);
```
devsnek: its a shame that read_byob takes this by value instead of by reference, since it might end up being reallocated every loop

devsnek (approved):

PR #1247, opened 2025-11-21
feat: source phase imports

nayeemrmn:
- For https://github.com/denoland/deno/issues/28013
- Downstream: https://github.com/denoland/deno/pull/31486

### DIFF
--- a/core/modules/map.rs
+++ b/core/modules/map.rs
@@ -1921,7 +1921,7 @@ impl ModuleMap {
                 let key = ModuleSourceKey::from_reference(module_reference);
                 let source = {
                   let data = self.data.borrow();
-                  let source = data.sources.get(&key).unwrap().as_ref();
+                  let source = data.sources.get(&key).expect("Source had to have been inserted successfully, or recursion would error.").as_ref();
                   v8::Local::new(scope, source).into()
                 };
                 {


core/modules/map.rs
```diff
@@ -150,7 +155,7 @@ pub(crate) struct ModuleMap {
     RefCell<FuturesUnordered<Pin<Box<CodeCacheReadyFuture>>>>,
   pending_code_cache_ready: Cell<bool>,
   module_waker: AtomicWaker,
-  data: RefCell<ModuleMapData>,
+  pub(crate) data: RefCell<ModuleMapData>,
```
bartlomieju: Can we remove this `pub(crate)` and instead add a method to handle that?

core/modules/map.rs
```diff
@@ -1033,21 +1075,56 @@ impl ModuleMap {
...
+    {
+      Some(ModuleSourceData {
+        value,
+        kind: ModuleSourceKind::Wasm,
+      }) => Some(v8::Local::new(scope, value)),
+      None => {
+        let message = v8::String::new(
+          scope,
+          &format!(
+            r#"Module source can not be imported for "{specifier_str}""#
```
bartlomieju: This is a bit hard to read and I was surprised it doesn't cause borrow checker error. Consider rewriting into `if let Some(...) {}`

PR #1250, opened 2025-11-27
ci: update `tools/` to latest Deno

bartlomieju:
This commit updates `tools/` directory with latest Deno conventions.

> Refactors the tool infrastructure to decentralize quality checks. The centralized tools/check.ts script that orchestrated formatting and linting is deleted. Its responsibilities are split into separate tools/format.ts and tools/lint.ts scripts. Import paths across tools are updated from ./deps.ts to directly use @deno/rust-automation, with a new tools/deno.json configuration mapping this import. Several tool files update their imports and invocation paths accordingly, and some scripts adjust their shebang and Deno runtime flags. The overall workflow remains functionally similar but now uses modular, dedicated scripts instead of a single orchestrator.

### DIFF
--- a/.github/workflows/ci-job.yml
+++ b/.github/workflows/ci-job.yml
@@ -57,7 +57,9 @@ jobs:
         run: cargo binstall cargo-nextest --secure
 
       - name: Install Deno
-        uses: denoland/setup-deno@v1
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
RSLGTM

PR #1261, opened 2025-12-13
chore: use `deno fmt`

iuioiua:
Seems to do the same job fine.


.dprint.json
```diff
@@ -1,30 +0,0 @@
...
-  "json": {
-    "deno": true
-  },
-  "exec": {
-    "cwd": "${configDir}",
-    "commands": [{
-      "command": "rustfmt",
-      "exts": ["rs"]
-    }]
-  },
```
bartlomieju: This was ensuring that Rust code was formatted too and now it's lost. Can you add the check to the CI script?

PR #1262, opened 2025-12-14
feat(op2): remove `async` attribute and make `FromV8`/`ToV8` the default behaviour

crowlKats:
This PR is split into multiple commits; I would recommend to review them separately.

This PR does the following changes:
- removes the `async` keyword on the macro declaration; it is fully inferred from the function definition. The `async` alternatives (`fake`, `lazy`, & `deferred`) remain unchanged.
- removes the `global` attribute for arguments and returns types.
- adds a `FromV8Scopeless` trait, which is the default behaviour for arguments now, and requires a `scoped` attribute on arguments to use the `FromV8` trait.
- return types will default to usage of the `ToV8` trait.
- `from_v8` and `to_v8` attributes are removed as these are now the default behaviour.
- Adds some new structs that implement `FomV8`/`ToV8`.

There is https://github.com/denoland/deno/pull/31607 ready for updating in CLI.

### DIFF
--- a/core/webidl.rs
+++ b/core/webidl.rs
@@ -1495,6 +1495,19 @@ mod tests {
     assert_eq!(converted, Enumeration::FooBar);
     assert_eq!(converted.as_str(), "foo-bar");
 
+    let val = v8::String::new(scope, "foo-bar").unwrap();
+    let val = v8::Array::new_with_elements(scope, &[val.into()]);
+    let converted = Enumeration::convert(
+      scope,
+      val.into(),
+      "prefix".into(),
+      (|| "context".into()).into(),
+      &Default::default(),
+    )
+    .unwrap();
+    assert_eq!(converted, Enumeration::FooBar);
+    assert_eq!(converted.as_str(), "foo-bar");
+
     let val = v8::String::new(scope, "baz").unwrap();
     let converted = Enumeration::convert(
       scope,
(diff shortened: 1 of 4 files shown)


core/runtime/ops.rs
```diff
@@ -2491,8 +2469,7 @@ mod tests {
   }
 
   #[op2]
-  #[to_v8]
-  fn op_bool_to_from_v8(#[from_v8] value: Bool) -> Bool {
+  fn op_bool_to_from_v8(#[v8_slow] value: Bool) -> Bool {
```
nathanwhit: Why does this one require `#[v8_slow]` but not `Number<f64>`?

core/convert.rs
```diff
@@ -823,14 +1043,127 @@ where
...
+impl<'s, T, E> ToV8Fast<'s> for v8::Local<'s, T>
+where
+  Local<'s, T>: TryInto<Local<'s, v8::Value>, Error = E>,
+  E: Into<V8ConvertError>,
+{
+  fn to_v8(self) -> Result<Local<'s, v8::Value>, Self::Error> {
+    self.try_into().map_err(Into::into)
+  }
+}
+
```
nathanwhit: Why is this impl ok? I thought that you can't really return values from fast calls in general, though now I can't actually find the v8 docs on it

littledivy (approved):
Looks good. Let's land in Deno and run benchy to measure perf

PR #1268, opened 2025-12-19
fix(core): queue dynamic imports for modules with pending TLA

fraidev:
Closes https://github.com/denoland/deno_core/issues/1149
Related https://github.com/denoland/deno/issues/29123

When a module with Top-Level Await (TLA) is dynamically imported multiple times before the TLA resolves, subsequent imports would fail with "Cannot access 'default' before initialization".

This happened because V8 sets module status to `Evaluated` immediately after calling `module.evaluate()`, even when TLA is pending. Deno would see `Evaluated` status and try to resolve immediately, exposing uninitialized bindings (TDZ violation).

The fix queues subsequent import resolvers in `pending_tla_waiters` when a module has pending TLA, and resolves them all when the TLA promise completes.

### DIFF
--- a/testing/integration/dyn_import_pending_tla/dyn_import_pending_tla.out
+++ b/testing/integration/dyn_import_pending_tla/dyn_import_pending_tla.out
@@ -1,0 +1,2 @@
+All imports resolved to same instance
+Default export: tla module loaded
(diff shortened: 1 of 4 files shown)


fraidev:
> Can you add a test from original issue?

I added an integration test for it. I also add it in https://github.com/denoland/deno/pull/31670 before. Should we have the test only here?

PR #1293, opened 2026-01-28
feat: always call `prepare_load` for modules loaded via `op_import_sync`

dsherret:
Also removes some allocations. I couldn't help myself while doing this.

CLI PR: https://github.com/denoland/deno/pull/31974

### DIFF
--- a/.github/workflows/ci-publish/action.yml
+++ b/.github/workflows/ci-publish/action.yml
@@ -4,4 +4,11 @@ runs:
   steps:
     - name: Cargo publish
       shell: bash
-      run: cargo publish --workspace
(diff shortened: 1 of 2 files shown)


core/modules/map.rs
```diff
@@ -1199,24 +1198,20 @@ impl ModuleMap {
 
   pub(crate) async fn load_main(
     module_map_rc: Rc<ModuleMap>,
-    specifier: impl AsRef<str>,
+    specifier: String,
```
dsherret: All these changes are just removing allocations. I'll point out the main change for this PR below.

PR #1309, opened 2026-02-23
fix: event loop rewrite and libuv compat layer

littledivy:


core/01_core.js
```diff
@@ -162,32 +159,30 @@
...
-  function eventLoopTick() {
-    let didAnyWork = false;
-
-    // First respond to all pending ops.
-    for (let i = 0; i < arguments.length - 3; i += 3) {
-      didAnyWork = true;
-      const promiseId = arguments[i];
+  // Phase 2: Resolve completed async ops. Called from Rust with flat args:
+  // (promiseId, isOk, res, promiseId, isOk, res, ...)
+  function __resolveOps() {
```
nathanwhit: We should benchmark and see how much invoking all of these extra JS callbacks costs us per event loop tick

